PR #12, opened 2017-01-04
Y axis context menu behavior in Metric Explorer

jsperhac:
In the Metric Explorer, the Y axis context menu prevented the axis min/max value(s) and log scale controls from being changed together. The menu's log scale control, if toggled, caused unsaved axis min and max values to be disregarded. This was a misleading result as there is an Ok/Cancel button pair provided on the menu. This behavior is corrected with the fix in this pull request.

## Description
In the Metric Explorer's Y axis context menu, the control that toggles log scaling is changed to a form.checkbox from a menu.checkitem. The existing handler is adapted to include the new checkbox. The handler now incorporates any set of changes made in the context menu into a single change on the change stack. User input of log plot min value <=0 is corrected prior to plotting. Log scale and scaling are applied, treated as a single change, after the OK button is clicked. The existing behavior of the context menu is otherwise retained.

## Motivation and Context

This menu behavior inconsistency issue was reported here:
https://app.asana.com/0/14787510600562/210274615925079

Additionally, it fixes the following:
 - an unreported bug in which user input of log plot min value <0 causes y-axis to disappear
  - an unreported bug in which change to log plot with min value 0 throws Highcharts error
 - incorporates plot changes made in this context menu to the change stack, as a unit
 - prevents memory leaks (from this menu) by destroying menu after use
 - format, variable definition order issues found by linter

## Tests performed

Tests were made to verify that changes to min, max, and log setting could be made either simultaneously, or separately, following click of OK button. Tested the following:
 - Toggling between log and non-log 
 - Zooming in or out on plot (min and max values) 
 - Entering minimum value <= 0, with log/linear plot
 - Changing min, max, and log/linear settings together, or separately
 - Interacting with change stack: undo/redo changes as a single unit
 - Separately adjusting y axis settings, for multiple data series on multiple or shared axes
 - Checking number of menu objects created on OK, Cancel, click away (memory leak)

Both firefox and chrome browsers were tested, on Ubuntu 16. Eslint was used to guide small format changes of the relevant code, submitted as a separate commit.

## Types of changes

## Checklist:

### DIFF
--- a/html/gui/js/modules/metric_explorer/MetricExplorer.js
+++ b/html/gui/js/modules/metric_explorer/MetricExplorer.js
@@ -1606,6 +1606,7 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
         if (instance === undefined) {
             instance = CCR.xdmod.ui.metricExplorer;
         }
+        var handler;
         var axisIndex = axis.options.index;
         var axisTitle = axis.options.title.text;
         var originalTitle = axis.options.otitle;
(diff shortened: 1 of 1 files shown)


html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1610,6 +1610,24 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
         var axisTitle = axis.options.title.text;
         var originalTitle = axis.options.otitle;
         var defaultTitle = axis.options.dtitle;
+
```
jpwhite4: How come this block of code was moved?

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1633,47 +1651,84 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
...
-                 * power of the base that is still larger than the
-                 * current maximum value.
-                 */
-                function calcMax(base, cur_max) {
-                    var result = 0;
-                    var exponent = 1;
-                    while (result < cur_max) {
-                        result = Math.pow(base, exponent);
-                        exponent += 1;
+            // Set the value of allLog
```
jpwhite4: Surely this code block uses the value of allLog rather than setting it.

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1684,14 +1739,20 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
...
                     newMin = null;
                 }
-            }
 
-            if (newMin == "") {
-                newMin = 0;
-            }
+            } else {
+
+                axisType = 'linear',
```
jpwhite4: You have added a new config property of the axis type to the chart object. This was not being done previously as the log scale attribute was added to the datasets themselves. Now you add it both to the datasets and the axis config. What is the motivation for and consequence of this change?

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1633,47 +1651,84 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
...
+                        record.set('log_scale', allLog);
                     }
-                    return result;
                 }
+            });
+
+            // Calculate best min and max for log plot
+            if (allLog) {
+
+                axisType = 'logarithmic',
```
jpwhite4: This should be a semi colon not a comma.

jsperhac: Quite right, that's just incorrect.

jpwhite4:
Also this change should go in to the 6.6 branch not 6.5

html/gui/js/modules/metric_explorer/MetricExplorer.js
jsperhac: I'm setting the axis type because the call to the Highcharts axis update() function in line 1780 throws an uncaught error (Highcharts #10) if it receives a null axis minimum value and log was previously set. I can get around this by placing the call to update() in a try/catch block, but the underlying cause is that Highcharts does not know whether the axis should be log or linear. Passing the attribute here makes it clear.

html/gui/js/modules/metric_explorer/MetricExplorer.js
jsperhac: Thinking aloud. It's been restored to its original location.

html/gui/js/modules/metric_explorer/MetricExplorer.js
jsperhac: You are right--comment has been corrected.

jsperhac:
Updated pull request: comment corrected, commas removed, and unnecessary code shuffling has been undone. It deploys cleanly on xdmod6.6. I intend to squash these commits together but left them separate here for clarity. 

The axis type option now being passed to the axis redraw is needed so that highcharts does not throw an error if the last setting for the axis was log and the min value is 0. See remark on commit 35f1440.

jsperhac:
Final remarks:
I made some changes to the summary above to describe the bugs found and fixed in preparation of this code. Note that the axis update() method is called nowhere else in MetricExplorer.js; passing axis type does not need to be done elsewhere in the ME code.

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1649,31 +1649,89 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
                     allLogScale = (allLogScale === undefined || allLogScale === true) && record.get('log_scale');
                 }
             }
+        });
 
+        var setLog = new Ext.form.Checkbox({
```
jpwhite4: This code leaks memory since you create a new checkbox every time the axis is clicked, but do not destroy is when it is finished with. Looks like the existing code had the same problem, but this change increases the size of the leak.

jsperhac: The context menu is now destroyed on click away, click OK, or click Cancel.

jpwhite4:
This code introduces a new error on line 1663:

1663:25  NEW ERROR    'handler' was used before it was defined. (no-use-before-define)

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -1719,55 +1788,21 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
...
                 }, true);
 
                 instance.saveQuery();
             }
-            menu.hide();
+
+            // re-enable the undo stack to treat this as a single change
+            instance.fireEvent('enable_commit', true);
+
+            menu.destroy();
```
jpwhite4: This code still leaks memory when the user clicks away from the menu. (This is an example of a bug that was in the existing code, but since you are changing the code , this is a good opportunity to fix the problem).

I suggest:
- do not change the calls to menu.hide() to menu.destroy()
- add a hander to the hide event for the menu and call destroy() in the event hander.

This will mean that the memory is cleaned up in all three cases.

jpwhite4: Sorry. My mistake. I didn't spot that you had already added the event listener for the hide event.

PR #19, opened 2017-01-09
fix notices in xdsamlauth

plessbd:
## Description
notices are thrown if the idp does not have the right indexes

## Motivation and Context
Dont want to throw notices if they are not needed

## Tests performed
Verified that the notices were no longer shown when display_errors = On in php.ini

## Types of changes

## Checklist:

### DIFF
--- a/classes/Authentication/SAML/XDSamlAuthentication.php
+++ b/classes/Authentication/SAML/XDSamlAuthentication.php
@@ -113,10 +113,10 @@ class XDSamlAuthentication
             $orgDisplay = "";
             $icon = "";
             foreach ($idpAuth as $idp) {
-                if ($idp['OrganizationDisplayName']) {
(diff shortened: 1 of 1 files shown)


classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -113,10 +113,10 @@ public function getLoginLink()
             $orgDisplay = "";
             $icon = "";
             foreach ($idpAuth as $idp) {
-                if ($idp['OrganizationDisplayName']) {
+                if (array_key_exists('OrganizationDisplayName', $idp) && !empty($idp['OrganizationDisplayName'])) {
```
jpwhite4: the array_key_exists is redundant as the empty() built-in does an isset() check and does not give a warning if the variable is not set. See e.g. http://php.net/manual/en/function.empty.php

PR #28, opened 2017-01-17
Update rules to exclude restrictions on control structure indentation.

jpwhite4:
Relax the phpcs rules so that it does not disallow perfectly reasonably indented code.

## Motivation and Context
Forcing a clurly brace on one line vs the next does not result in a big improvement to the code quality, but it does result in a onerous burden on the developers. This change relaxes these types of rules.

## Tests performed
Ran phpcs on the code with the new ruleset.


smgallo:
The only error that I still saw was CamelCase, because I prefix interface class names with "i" and abstract class names with "a" to they are not confused with concrete classes. For example iDataEndpoint and aOptions.

PR #34, opened 2017-01-20
Support for Oracle ETL DataEndpoint (readonly)

smgallo:
Implement an ETL DataEndpoint for reading from an Oracle database.

## Description

Implements support for importing data from an Oracle database. In addition to adding an Oracle DataEndpoint, the following changes were made as a result of necessary features for importing

- More flexible database parameters in .ini file (not all databases require all parameters). For example, Oracle supports Easy Connect and Local database connection methods and a port does not always need to be specified.
- Include support for specifying ORDER BY in ETL queries
- The PDOOCI wrapper for OCI8 is included. The existing PDO code included with PHP is no longer supported.
- Minor code cleanup and style fixes
- Improve error messages
- Removes the old Sybase code

Note that Oracle requires downloading the [Oracle Instant Client](http://www.oracle.com/technetwork/database/features/instant-client/index-097480.html) libraries.

## Motivation and Context

We will need to pull organizational hierarchy and user affiliations from UB's data warehouse (Infosource). This will also assist SDSC with migration from their existing Oracle system.

## Tests performed

Imported the organizational hierarchy from UB's Oracle instance.

## Types of changes

## Checklist:


classes/ETL/DataEndpoint/aRdbmsEndpoint.php
```diff
@@ -123,8 +129,15 @@ public function connect()
         // DB::factory() does not support connecting to an alternate schema and needs a separate entry
         // in the config file. You can, however, explicitly reference a schema in your query.
 
-        $this->handle = DB::factory($this->config);
+        try {
+            $this->handle = DB::factory($this->config);
+        } catch (Exception $e) {
+            $msg = "Error connecting to data endpoint '" . $this->name . "'. " . $e->getMessage();
+             $this->logAndThrowException($msg);
```
jsperhac: Extra whitespace on line 136?

classes/ETL/DataEndpoint/Postgres.php
```diff
@@ -40,9 +39,11 @@ public function isSameServer(iDataEndpoint $cmp)
      * ------------------------------------------------------------------------------------------
      */
 
-    public function schemaExists($schemaName)
+    public function schemaExists($schemaName = null)
     {
-        if ( empty($schemaName) ) {
+        if ( null === $schemaName ) {
+            $schemaName = $this->getSchema();
```
jsperhac: Lines 44-49 (from Postgres.php) are seen frequently in these classes that extend aRdbmsEndpoint--is it worth putting a function in aRdbmsEndpoint to handle the test for SchemaName and the logging action?

classes/ETL/DataEndpoint/Oracle.php
```diff
@@ -0,0 +1,190 @@
...
+            $schemaName = $this->getSchema();
+        } elseif ( empty($schemaName) ) {
+            $msg = "Schema name cannot be empty";
+            $this->logAndThrowException($msg);
+        }
+
+        // See http://www.postgresql.org/docs/current/static/catalogs.html
+
+        $sql = "SELECT
+username AS name
```
jsperhac: Formatting seems off for these SELECT strings (lines 54, 114, 160).

classes/ETL/DataEndpoint/README_ORACLE.md
```diff
@@ -0,0 +1,219 @@
+# Oracle Data Endpoint
```
jsperhac: Does it make more sense to put this file under configuration/ or docs/, perhaps?

jsperhac: Overall, our documentation needs to be updated to reflect Oracle support (and emphasize that it is read-only for ETL), as well.

classes/ETL/DataEndpoint/Oracle.php
smgallo: My editor doesn't indent bare strings on the line. Plus, the extra indentation makes it more difficult to read in the debug output.

classes/ETL/DataEndpoint/aRdbmsEndpoint.php
smgallo: Interesting that phpcs didn't complain about that.

classes/ETL/DataEndpoint/Postgres.php
smgallo: Yes, that makes sense.

classes/ETL/DataEndpoint/README_ORACLE.md
smgallo: I moved it to docs/etl/data-endpoint-oracle.md. I also need to update the ETL documentation overall and move it out of confluence. There are a lot of docs to move.

tools/etl/etl_overseer.php
```diff
@@ -689,8 +728,8 @@ function usage_and_exit($msg = null)
 
     NOTE: Date and time options support "+1 day", "now", "now - 1 day", etc. notation.
 
-EOMSG
-        );
+        EOMSG;
```
plessbd: [heredoc](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) requires closing tag to be at the beginning of the line.

> Warning
> 
> It is very important to note that the line with the closing identifier must contain no other characters, except a semicolon (;). That means especially that the identifier may not be indented, and there may not be any spaces or tabs before or after the semicolon. It's also important to realize that the first character before the closing identifier must be a newline as defined by the local operating system. This is \n on UNIX systems, including Mac OS X. The closing delimiter must also be followed by a newline.
> 
> If this rule is broken and the closing identifier is not "clean", it will not be considered a closing identifier, and PHP will continue looking for one. If a proper closing identifier is not found before the end of the current file, a parse error will result at the last line. 
> 

I believe this is what is now causing travis to fail with
```
PHP Parse error:  syntax error, unexpected $end in tools/etl/etl_overseer.php on line 737
```

smgallo:
OK, all style and unit test issues are fixed as well as changes requested by @jsperhac and @plessbd

plessbd:
Looks good to me once the composer.lock is fixed to not update dependencies.

smgallo:
Composer lock file reverted. The new file only has timestamps updated and the requirements for PDOOCI. I needed to clear my composer cache or else running `composer require` would update the packages to the new versions.

PR #35, opened 2017-01-20
Travis: Manually update Composer when automatic update fails

tyearke:
## Description
This pull request fixes a case where Travis CI was not updating Composer to the latest stable version, causing builds with newer Composer config files to fail.

## Motivation and Context
When Travis attempts to automatically update Composer to the latest version while using PHP 5.3.3, it fails because of SSL issues with Travis systems using PHP 5.3.3. This meant that tests were done using the pre-installed version of Composer, 1.0.0. This now presents a problem, as newer versions of Composer have introduced backwards-incompatible changes to the lock file format.

## Tests performed
I ran the updated Travis scripts with pushes on my fork. The results may be found [here](https://travis-ci.org/tyearke/xdmod/builds/193842715).

## Types of changes

## Checklist:

### DIFF
--- a/.travis.install.sh
+++ b/.travis.install.sh
@@ -7,6 +7,16 @@ if "$using_php_533"; then
     echo "Using newer version of PHP for installing dependencies"
     phpenv global 5.3
     php --version
+
+    # Update Composer since the attempt made by the Travis setup script
+    # using PHP 5.3.3 would have failed.
+    #
+    # First, we update to the latest developer version since the version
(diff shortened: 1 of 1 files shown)


jpwhite4:
Please do not forget to check whether changes to the travis code are also needed in the XDMoD sub modules. The code is very similar, which makes it likely that an error in one also affects the other.

jpwhite4:
Also don't forget that the lock files needed to updated too when you update composer. and this needs to happen in all of the sub-modules that  also need the self-update commands.

PR #60, opened 2017-02-23
Organize realms under categories

tyearke:
## Description
This pull request allows related realms to be grouped together in categories. Further, it reorganizes the web client's UI to be based around categories rather than realms.

## Motivation and Context
In order to prepare and present grant data addressing different focuses, the in-development Value Analytics module currently provides two realms: `ValueAnalyticsGrants`, which operates on data for individual grants as a whole, and `ValueAnalyticsGrantRoles`, which operates on each assignment onto each grant. However, presenting these realms as separate entities in the Open XDMoD web client led to confusion about what the different entities meant and where to find all of the metrics that pertained to the subject of grants.

To begin to address this problem, this pull request introduces categories. A realm may now declare a category that it falls under, and in the UI, that realm's dimensions and metrics will be grouped together with those of other realms under the same category. If a realm does not declare a category, it is placed into a category with the same name as the realm. From a web client user's perspective, realms have been effectively replaced by categories, but the flow of the interface  remains the same.

**Usage Tab Menu Before**

Usage Tab Before Change

**Usage Tab Menu After**

Usage Tab After Change

Despite their removal from a web client user's perspective, realms remain important to defining and requesting data in Open XDMoD. Realms continue to have a one-to-one relationship with related sets of fact tables, and they are still populated with dimensions and metrics pertaining to the data represented by those fact tables. For example, the `ValueAnalyticsGrants` realm still represents individual, whole grant data stored in the `modw_aggregates.va_grant_fact_by_*` tables. Realms are also still the basis for most functionality related to the data warehouse for two reasons: 1) It allows the category layer of the data warehouse hierarchy to be treated largely as a cosmetic layer, allowing it to be changed while only impacting the UI, and 2) It reduces the amount of change introduced to the system as a whole, allowing pre-existing data and most API calls to continue working without change.

An important consideration with this change is that metrics from different realms may have different sets of dimensions which can be used to group or filter them. A category-based UI makes this concept even more difficult to convey than the current realm-based UI, as metrics under the same category may not all work with the same dimensions. This issue does not need to be addressed immediately, as all realms currently in Open XDMoD and released official modules have a one-to-one relationship with categories, but it does need to be addressed before multi-realm categories are officially supported.

### Bug Fix

As a part of this change, a minor bug has also been fixed in the Metric Explorer where an old entry in the Filters drop-down panel would not update the list of realms a filter applied to when that list of realms changed. For example, if a chart with a resource filter was created and saved in a vanilla instance of Open XDMoD, the list of applicable realms for that filter would only include the Jobs realm. If the SUPReMM module is then installed and that chart is opened again, the filter pane would continue to say the filter only applied to the Jobs realm, even though it now applies to the SUPReMM realm as well. By looking up this information at load time instead of using saved information, this information will stay up to date with whatever realms are present and available to a user.

## Tests performed
I manually tested a wide variety of features and use cases in the Summary, Usage, Metric Explorer, and Report Generator tabs, involving both pre-existing charts and new charts, using Chrome. I also ran automated UI tests with Firefox and verified that no new errors occurred.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse.php
+++ b/classes/DataWarehouse.php
@@ -15,6 +15,7 @@
  */
 
 use CCR\DB;
+use Xdmod\Config;
 
 /*
  *	@Class DataWarehouse
@@ -448,4 +449,52 @@ class DataWarehouse
         );
         return $results;
     } // getAllocationsByChargeNumber()
+
+    /**
+     * Get a mapping of categories to realms.
+     *
+     * If a realm does not explicitly declare a category, its name is used
+     * as the category.
+     *
+     * @return array An associative array mapping categories to the realms
+     *               they contain.
+     */
+    public static function getCategories()
+    {
+        $config = Config::factory();
+        $dwConfig = $config['datawarehouse'];
+
+        $categories = array();
+        foreach ($dwConfig['realms'] as $realmName => $realm) {
+            $category = (
+                isset($realm['category'])
+                ? $realm['category']
+                : $realmName
+            );
+            $categories[$category][] = $realmName;
+        }
+        return $categories;
+    }
+
+    /**
+     * Get the category for a given realm.
+     *
+     * If a realm does not explicitly declare a category, its name is used
+     * as the category.
+     *
+     * @param  string $realmName The name of the realm to get
+     *                           the category for.
+     * @return string            The category the realm belongs to.
+     */
+    public static function getCategoryForRealm($realmName)
+    {
+        $config = Config::factory();
+        $dwConfig = $config['datawarehouse'];
+
+        if (isset($dwConfig['realms'][$realmName]['category'])) {
+            return $dwConfig['realms'][$realmName]['category'];
+        }
+
+        return $realmName;
+    }
 } // class DataWarehouse
--- a/classes/DataWarehouse.php
+++ b/classes/DataWarehouse.php
@@ -23,43 +23,44 @@ use CCR\DB;
 class DataWarehouse
 {
 
-	private static $db = null;
+    private static $db = null;
 
-	/*
+    /*
 	 * @function connect()
 	 * @access public
 	 */
-	public static function connect()
-	{
-		if(!self::$db)
-		{
-			self::$db = DB::factory('datawarehouse');
-		}
-		return self::$db;
+    public static function connect()
+    {
+        if(!self::$db)
+        {
+            self::$db = DB::factory('datawarehouse');
+        }
+        return self::$db;
 
-	}
-	/*
+    }
+
+    /*
 	 * @function destroy()
 	 * @access public
 	 */
-	public function destroy()
-	{
-		if(self::$db !== null)
-		{
-			self::$db = null;
-		}
-	}
+    public function destroy()
+    {
+        if(self::$db !== null)
+        {
+            self::$db = null;
+        }
+    }
 
-	/*
+    /*
 	 * @function __destruct()
 	 * @access public
 	 */
-	function __destruct()
-	{
-		destroy();
-	}
+    public function __destruct()
+    {
+        destroy();
+    }
 
-	/************************************************************
+    /************************************************************
 	 * @function getPersonIdByUsername
 	 * @access public
 	 *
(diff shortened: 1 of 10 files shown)


classes/DataWarehouse.php
```diff
@@ -448,4 +449,52 @@ public static function getAllocationsByChargeNumber($person_id, $start_date, $en
...
+     *
+     * If a realm does not explicitly declare a category, its name is used
+     * as the category.
+     *
+     * @return array An associative array mapping categories to the realms
+     *               they contain.
+     */
+    public static function getCategories()
+    {
+        $config = Config::factory();
```
smgallo: I know it's not part of this pull request, but the Xdmod/Config class really needs some documentation at the top describing what it does. Not to mention the generic `@see ArrayAccess` comments do nothing at all to describe what the heck it's returning. I had to read through the class just to figure out that Config::offsetGet($offset) returns the value associated with a JSON key in a config file.

classes/DataWarehouse/Visualization/HighChart2.php
```diff
@@ -542,30 +543,33 @@ protected function registerContextMenus()
...
+                $data_description->realm
+            );
+            if(isset($pCategory) && $category != $pCategory)
             {
-                $this->_multiRealm = true;
+                $this->_multiCategory = true;
                 return;
             } else {
-                $pRealm = $data_description->realm;
+                $pCategory = $category;
```
smgallo: `$pCategory` is not initialized. If its referring to the previous category change it to `$prevCategory` because I initially thought it was referring to a parent category since we're dealing with JSON data.

classes/DataWarehouse/Visualization/HighChart2.php
```diff
@@ -542,30 +543,33 @@ protected function registerContextMenus()
...
     //
     // called by configure()
     // JMS, June 2015
     //
     // @param data_series
     // ---------------------------------------------------------
-    protected function setMultiRealm(&$data_series)
+    protected function setMultiCategory(&$data_series)
     {
         foreach($data_series as $data_description_index => $data_description)
```
smgallo: Add a one-line comment to describe what we're doing here: Search all data series to find realm categories. If there are 2 distinct categories set multi-category = true.

classes/DataWarehouse/Visualization/HighChartTimeseries2.php
```diff
@@ -153,15 +155,16 @@ public function configure(
...
-            if(isset($pRealm) && $data_description->realm != $pRealm)
+            // set multiCategory true or false
+            $category = DataWarehouse::getCategoryForRealm($data_description->realm);
+            if(isset($pCategory) && $category != $pCategory)
             {
-                $multiRealm = true;
+                $multiCategory = true;
             } else {
-                $pRealm = $data_description->realm;
+                $pCategory = $category;
```
smgallo: The above comment is also relevant here.

html/controllers/user_interface/get_menus.php
```diff
@@ -34,56 +34,114 @@
         }
     } elseif (
         isset($_REQUEST['node'])
-        && substr($_REQUEST['node'], 0, 6) == 'realm_'
+        && \xd_utilities\string_begins_with($_REQUEST['node'], 'category_')
```
smgallo: Any reason not to use `0 === strpos($_REQUEST['node'], 'category_')`?

classes/DataWarehouse/Access/Usage.php
```diff
@@ -43,752 +46,764 @@ class Usage extends Common
...
-                'results' => \xd_charting\encodeJSON(array(
-                    'success' => false,
-                    'message' => 'No charts could be generated using the given parameters.',
-                    'totalCount' => 0,
-                    $chartsKey => array(),
-                )),
-                'headers' => array(),
-            );
-        }
+                        if (\xd_utilities\array_get($meResult, 'restrictedByRoles', false)) {
```
smgallo: Since you're not using the value at all why not use `if ( ! array_key_exists('restrictedByRoles', $meResult) )` instead of the overhead of another function call?

html/controllers/user_interface/get_menus.php
tyearke: My stylistic preference for conditionals is to (where reasonable) use functions that return a boolean in order to answer a yes/no question about what it is you're checking for. Since we already had a helper function for checking if a string starts with another string, I opted to use that. If you prefer the builtins-only form instead, I can change it.

classes/DataWarehouse/Access/Usage.php
tyearke: The value of `$meResult['restrictedByRoles']` is being used to determine if `$combinedResultRestricted` should be set to true in this loop. `$combinedResultRestricted` should only be set to true if the value is both present and true. Otherwise, the variable should be left as it is.

Regardless, the only change made to this code was that it was indented in order to make it part of a for loop. I would prefer not to tweak it or similar code as part of this pull request unless it is relevant to the change being made.

Issue #64, opened 2017-03-04
PHP 5.3 Issues


smgallo:
PHP 5.3 filter_var() does not properly filter the value `bool(false)`. This behavior is fixed 5.4.8 (see https://bugs.php.net/bug.php?id=49510)

In 5.3:
```php
$a = array(0, 1, true, false, "true", "false", "totalcrap");
foreach ($a as $v) { print var_dump($v); }
```
```
int(0)
int(1)
bool(true)
bool(false)
string(4) "true"
string(5) "false"
string(9) "totalcrap"
php > foreach ($a as $v) { print var_dump(filter_var($v, FILTER_VALIDATE_BOOLEAN,
FILTER_NULL_ON_FAILURE)); }
bool(false)
bool(true)
bool(true)
NULL
bool(true)
bool(false)
NULL
```

In 5.5.9:
```php
$a = array(0, 1, true, false, "true", "false", "totalcrap");
foreach ($a as $v) { print var_dump(filter_var($v, FILTER_VALIDATE_BOOLEAN,
FILTER_NULL_ON_FAILURE)); }
```
```
bool(false)
bool(true)
bool(true)
bool(false)
bool(true)
bool(false)
NULL
```

smgallo:
A solution to this is to use
```
return ( false === $value ? false : filter_var($value, $filter, $options) );
```

PR #66, opened 2017-03-07
Adding Logging to Query.php

ryanrath:
## Description
- Added a file logger to Query.php ( query.log )
- Added logic to the function 'execute' such that, when called and 
  'sql_debug_mode' is true, the query string that is about to be executed is logged.

## Motivation and Context
It allows some visibility into the queries being generated / executed by the Query class.

## Tests performed
- I deployed the change to my cloud instance ( rr-db1 ), verified there were no issues creating the log file.
- verified that when 'sql_debug_mode' was set to 'on' queries were logged as expected. 
- verified that when 'sql_debug_mode' was set to 'off' queries were not logged as expected.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Query.php
+++ b/classes/DataWarehouse/Query/Query.php
@@ -237,7 +237,7 @@ class Query
 
         $query_string = $this->getQueryString($limit);
 
-        $debug = filter_var(\xd_utilities\getConfiguration('general', 'sql_debug_mode'), FILTER_VALIDATE_BOOLEAN);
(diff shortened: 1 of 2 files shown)


classes/DataWarehouse/Query/Query.php
```diff
@@ -229,6 +236,13 @@ public function execute($limit = 10000000)
     {
 
         $query_string = $this->getQueryString($limit);
+
+        $debug = filter_var(\xd_utilities\getConfiguration('general', 'sql_debug_mode'), FILTER_VALIDATE_BOOLEAN);
```
smgallo: filter_var is broken in php 5.3 (which is our supported version). See issue #64 for a workaround. I have a function in the ETL\Utilities class but maybe it should be brought up into xd_utilities.

PR #69, opened 2017-03-10
Remove ETL specific filter_var fix and use new fix in xd_utilities

smgallo:
## Description

@ryanrath added a fix for the PHP 5.3 filter_var bug in xd_utilities. Remove the ETL specific filter_var fix and use new fix in xd_utilities. Also added a test for the specific case where filter_var() fails.

## Motivation and Context
Cleaner code

## Tests performed
Re-ran ETLv2 after changes were made.

## Types of changes

## Checklist:

### DIFF
--- a/libraries/utilities.php
+++ b/libraries/utilities.php
@@ -556,7 +556,15 @@ EOF;
 }
 
 /**
- * A temporary shim function to use while our supported PHP version is < 5.4.8
+ * A temporary shim function to use while our supported PHP version is < 5.4.8 because 5.3
+ * incorrectly returns NULL in the following case:
+ *
+ * filter_var(false, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE)
+ *
+ * See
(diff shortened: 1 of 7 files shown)


libraries/utilities.php
```diff
@@ -556,15 +556,26 @@ function checkForCenterLogo($apply_css = true)
...
  * @param int $filter the type of filter to apply
  * @param mixed $options the options to be supplied to the filter
  * @return bool|mixed false if the value is logically false, else the results of
  * \filter_var($value, $filter, $options)
  */
+
 function filter_var($value, $filter = FILTER_DEFAULT, $options = null)
 {
-    return ( false === $value ? false : \filter_var($value, $filter, $options) );
+    return ( FILTER_VALIDATE_BOOLEAN == $filter && false === $value
```
smgallo: The solution that I gave @ryanrath originally was specific to `FILTER_VALIDATE_BOOLEAN` so I added that check here.

PR #79, opened 2017-03-27
Re-added integration test harness.

jpwhite4:
## Description
This harness used to be in XDMoD and was not copied across when the
repo was originally made public.

Also include a couple of example test files that check the metric
explorer and usage explorer controllers. These files are intended to
be used as templates to help developers write their own integration
tests.

## Motivation and Context
testing == good

## Tests performed
Yes


jsperhac:
I merged this pull request, since it was approved...however, there are a couple of issues as written:
- Exceptions as thrown from XdmodTestHelper.php do not work:
`Fatal error: Call to undefined function Exception() in /data/www/dev/jsperhac/xdmod/open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/XdmodTestHelper.php on line 114
`

- The authentication call in XdmodTestHelper does not appear to be working, an error is encountered when the token that should have been returned is retrieved:
`1) IntegrationTests\REST\Warehouse\JobViewerTest::testResourceEndPoint
Illegal string offset 'results'

/data/www/dev/jsperhac/xdmod/open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/XdmodTestHelper.php:119
/data/www/dev/jsperhac/xdmod-supremm/tests/integration_tests/lib/REST/Warehouse/JobViewerTest.php:23`

jsperhac:
When I use my dev port on xdmod-dev as the target url, I see the above illegal string offset on authorization calls. When I use the xdmod-dev instance itself, all authorizations just fail:
`2) IntegrationTests\REST\Warehouse\JobViewerTest::testInvalidJobSearch
Failed asserting that 400 matches expected 403.
`

smgallo:
In XdmodTestHelper.php the there is a `new` missing when throwing the exception. It should be

```
throw new \Exception("User role $userrole not defined in .secrets file");
```

jsperhac:
Steve is right. Aside from that, the majority of the problem was fixed with proper configuration.

PR #84, opened 2017-04-04
Changes to accompany issues found in OSG modifications

smgallo:
Changes to accompany issues found in OSG modifications

See ubccr/xdmod-xsede#23

## Description

The aggregation query for job records and tasks now uses a datatype of `DECIMAL(36,4)` rather than `DOUBLE` for the `sum_wallduration_squared`, `sum_waitduration_squared`, `sum_cpu_time_squared`, `sum_local_charge_xdsu_squared`, and `sum_node_time_squared` columns. The waitduration has been defaulted to `NULL` to support cases when we do not have this information (e.g., OSG).

Several enhancements to the table comparison tool were made:
- Ability to ignore number of columns between source and destination tables. This is useful for comparing tables with added columns to baseline tables
- Ability to ignore column types between source and destination tables. This is useful for comparing data when migrating column types from `double` to `decimal(m,n)`
- Ability to round values during comparison, useful when comparing data that may differ in the number of significant digits

## Motivation and Context

When comparing new data as the result of a change to baseline data it is helpful to use the `DECIMAL(M,N)` data type rather than `DOUBLE` which may use approximate representation of floating point numbers. This makes data verification much easier.

## Tests performed

The XSEDE job pipeline was run before and after changes to the aggregation query. Running the table verificaiton tool, data was identical with the exception of cases where the baseline used an approximate representation of the data (e.g., `4.1038701667445064e16`). In these cases the squared values were off by up to .00000000000000017%

## Types of changes

- [x] Enhancement to existing functionality

## Checklist:

### DIFF
--- a/tools/dev/verify_table_data.php
+++ b/tools/dev/verify_table_data.php
@@ -446,7 +446,7 @@ function compareTableData(
 
     $roundColumns = array();
     foreach ( $scriptOptions['round-columns'] as $column ) {
-        $parts = explode('=', $column);
+        $parts = explode(',', $column);
         $roundColumns[$parts[0]] = ( 2 == count($parts) ? $parts[1] : 0 );
     }
 
@@ -496,7 +496,7 @@ LEFT OUTER JOIN $destTableName dest ON (" . join("\nAND ", $constraints) . ")"
     $numRows = $stmt->rowCount();
 
     if ( 0 != $numRows ) {
-        $logger->warning(sprintf("Missing %d rows in %s.%s", $numRows, $destTable, $destSchema));
+        $logger->warning(sprintf("Missing %d rows in %s.%s", $numRows, $destSchema, $destTable));
         while ( $row = $stmt->fetch(PDO::FETCH_ASSOC) ) {
             $logger->warning(sprintf("Missing row: %s", print_r($row, 1)));
         }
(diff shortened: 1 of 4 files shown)


tools/dev/verify_table_data.php
```diff
@@ -453,8 +483,8 @@ function ($c1, $c2) {
     $sql = "
 SELECT src.*
 FROM $srcTableName src
-LEFT OUTER JOIN $destTableName dest ON (" . join(' AND ', $constraints) . ")"
-        . ( 0 != count($where) ? "\nWHERE " . implode(' AND ', $where) : "" )
+LEFT OUTER JOIN $destTableName dest ON (" . join("\nAND ", $constraints) . ")"
+        . ( 0 != count($where) ? "\nWHERE " . implode("\nAND", $where) : "" )
```
tyearke: The lack of whitespace after `AND` could cause problems.

tools/dev/verify_table_data.php
```diff
@@ -541,7 +541,7 @@ function usage_and_exit($msg = null)
     -n, --num-missing-rows <number_of_rows>
     Display this number of missing rows. If not specified, all missing rows are displayed.
 
-    -r, --round-column <column>[=<digits>]
+        -r, --round-column <column>[,<digits>]
```
tyearke: The indentation of this line is a little funky.

PR #91, opened 2017-04-07
Modular ACLS

ryanrath:
_note: I'll update this to not suck presently._

The intent of these changes is to provide an easy method for modules / module writers to interact with ACLS ( previously Roles ). In short, to take us from a hard coded set of roles / hierarchy to a more flexible architecture. 

## Description

The basic premise of how the system is used goes something like this: A module writer decides that they will require a new ACL with which to gate some/all of the functionality that they are providing. They write up a couple of json files to tell the system some basic information about their module and the acl they want to create. At install time these files are included / executed by an ETL action which incorporates them into the correct tables. Users can now be assigned the acl / they can check for it's existence in php code etc. 

The overarching data flow is thus: 
- json files / sql files are utilized at system / module install to ensure that database tables are present and properly populated. 
- All run time utilization of this information should now be drawn from these database tables. The config files need not be referenced. 

This flow, coupled with table foreign keys provides ample information when attempting to track down where a piece of information came from ( module ). 

## Motivation and Context

We have a very static set of roles / hierarchies at the moment. 

## Tests performed

All tests performed were manual tests utilizing both the site itself and Postman where appropriate to ensure that current  data / behavior.

## Types of changes

## Checklist:

The intent of these changes is to provide an easy method for modules / module writers to interact with ACLS ( previously Roles ). In short, to take us from a hard coded set of roles / hierarchy to a more flexible architecture.


classes/DBObject.php
```diff
@@ -0,0 +1,128 @@
...
+    protected function propertyExists($name)
+    {
+        $property = isset($this->PROP_MAP[$name]) ?  $this->PROP_MAP[$name] : null;
+        return array_key_exists($property, get_object_vars($this));
+    }
+
+    /**
+     * @inheritDoc
+     **/
+    public function __call($name, $arguments)
```
plessbd: would it be better to just use [__get and __set](http://www.php.net/manual/en/language.oop5.overloading.php) here ?

jpwhite4:
These code changes to not confirm to the coding standards. Please fix this. The travis checks do not pass with the current code.

A common failure is that the new classes have not been put in a namespace.

jpwhite4:
Is there any documentation that explains how to use this new code? As a module writer what do I need to do to define the acls for my (hypothetical) new group bys and statistics? When and how does the database get updated with the information about the acls?

html/gui/js/SessionManager.js
```diff
@@ -31,7 +31,7 @@ XDMoD.SessionManager._sessionExpiredNotification = new Ext.Window({
             new Ext.Button({
                 text: 'Complete Logout',
                 handler: function () {
-                    location.href = location.href.split('#')[0];
+                  location.href = location.href.split('#')[0];
```
jpwhite4: Do not make a single change to a file that has the only effect of changing the indentation level on one line.  Especially when the indentation was correct before the change and is wrong now.

classes/NewRest/Controllers/AclsControllerProvider.php
```diff
@@ -0,0 +1,258 @@
...
+    public function setupRoutes(Application $app, ControllerCollection $controller)
+    {
+        $root = $this->prefix;
+        $class = get_class($this);
+        $conversions = '\NewRest\Utilities\Conversions';
+
+        $isAuthorized = function (Request $request, Application $app) {
+            $authorized = $this->isAuthorized($request, array('mgr'));
+            if (!$authorized) {
+                return $app->json(
```
jpwhite4: You should throw the appropriate Unauthorized exception here and have the framework handle sending the correct data across the network.

classes/NewRest/Controllers/BaseControllerProvider.php
```diff
@@ -285,7 +286,7 @@ public function authorize(Request $request, array $requirements = null, $blackli
             if ($user->isPublicUser()) {
                 throw new AccessDeniedException($message, 401);
             } else {
-                throw new AccessDeniedHttpException($message, 403);
+                throw new AccessDeniedHttpException($message, null, 403);
```
jpwhite4: This file needs to be update as it is a merge conflict. You should use the code in pull request #87 as the template for the exception handler creation.

classes/NewRest/Controllers/MetricExplorerControllerProvider.php
```diff
@@ -84,7 +85,10 @@ public function setupRoutes(Application $app, ControllerCollection $controller)
             ->delete("$root/queries/{id}", "$base::deleteQueryById")
             ->convert('id', $idConverter);
         // QUERY ROUTES ========================================================
-
+        $controller
+            ->get("$root/descriptors", "$base::getDescriptorsForUser");
```
jpwhite4: What is the purpose of this endpoint? I cannot find anywhere in the code that tries to access it. Why does the post endpoint map to the read-only get function?

classes/NewRest/Controllers/RealmControllerProvider.php
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+
+        return $app->json(array(
+            'success'=> $success,
+            'data' => $data
+        ));
+    }
+
+    private function isSuccess(array $rows)
```
jpwhite4: Do not add unused private functions.

html/controllers/report_builder/enum_available_charts.php
```diff
@@ -9,11 +9,8 @@
    
    	$returnData['status'] = 'success';
    	
-   	$charts_in_other_roles = $cp->enumChartsUnderOtherRoles();
```
jpwhite4: What happens to a user that previously had charts defined in other roles? Are these charts going to be available in the interface now? What tests have be done to check this?

classes/XDUser.php
```diff
@@ -942,1881 +1038,1968 @@ public function saveUser()
...
-      ));
-      $this->_pdo->execute("UPDATE moddb.UserRoleParameters SET is_active=is_primary WHERE user_id=:user_id", array(
-         ':user_id' => $this->_id,
-      ));
+            }
 
-   }//assignActiveRoleToPrimary
+        } else {
 
+            $this->_pdo->execute("UPDATE moddb.UserRoleParameters SET is_active=0 WHERE user_id=:user_id AND role_id != :role_i
```
jpwhite4: Where is this function called with overwrite == true? I can't see it in the code at all.

jpwhite4: This code looks to me like it returns null if username is not set. The documentation says it throws an exception

classes/NewRest/Controllers/AclsControllerProvider.php
plessbd: in other places (https://github.com/ubccr/xdmod/blob/xdmod6.6/classes/NewRest/Utilities/Authentication.php#L102) we are throwing an error, why the change?

classes/NewRest/Controllers/AclsControllerProvider.php
```diff
@@ -0,0 +1,258 @@
...
+     * @param ControllerCollection $controller
+     * @return null
+     */
+    public function setupRoutes(Application $app, ControllerCollection $controller)
+    {
+        $root = $this->prefix;
+        $class = get_class($this);
+        $conversions = '\NewRest\Utilities\Conversions';
+
+        $isAuthorized = function (Request $request, Application $app) {
```
plessbd: why do we have an isAuthroized function here when we also have one in classes/NewRest/Utilities/Authorization.php Are they both needed?

plessbd:
Please squash commits into logical chunks to help reviewers understand the code changes a bit better.

classes/NewRest/Controllers/AclsControllerProvider.php
```diff
@@ -0,0 +1,258 @@
...
+                        'success' => false,
+                        'message' => 'Not authorized for the requested operation.'
+                    ),
+                    401
+                );
+            }
+        };
+
+        $controller->get("$root/", "$class::listAcls")
+            ->before($isAuthorized);
```
plessbd: why do we use before instead of the builtin SecurityServiceProvider (https://silex.sensiolabs.org/doc/1.3/providers/security.html) I couldnt find good 1.2 docs, but it is still in 1.2

classes/DBObject.php
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param array $options the options used to configure this instance.
+     **/
+    public function __construct($options = array())
+    {
+        $properties = $this->PROP_MAP;
+
+        foreach ($properties as $property => $value) {
+            if (array_key_exists($property, $options)) {
+                $this->$value = $options[$property];
```
ryanrath: Yup. 
 
```php
class Test extends DBObject
{
    protected $PROP_MAP = array(
        'name' => 'name'
    );
}

$test = new Test(array('name' => 'bob'));

$property = new ReflectionProperty($test, 'name');

$name = $property->getName();
$visibility = null;
if ($property->isPrivate()) {
    $visibility = 'private';
} else if ($property->isProtected()) {
    $visibility = 'protected';
} else if ($property->isPublic()) {
    $visibility = 'public';
}
echo "$name: $visibility\n";
```
Results in: 
`name: public`

plessbd: Is this test actually part of the unit tests for the new ACLs?  I didn't see any tests in this pull request.

ryanrath: This was something that I threw together quickly to ascertain the visibility of the properties created and respond to Steve's comment. I could of course "rephrase" this into an actual Unit test if the visibility of these properties is something that we want to enforce.

ryanrath:
**NOTE** 
I'm going to be leaving this pull request up to track the requested changes so far but will be submitting smaller, self contained PR's that are a combination of cherry picked commits + requested changes. I'll be addressing the basic documentation problems first ( so as to provide people context when they review the code ) and then move on to chunking up this monster into more manageable bites.

classes/Centers.php
```diff
@@ -0,0 +1,477 @@
...
+         JOIN group_bys gb ON uagbp.group_by_id = gb.group_by_id
+       WHERE a.name = 'cd'
+             AND uagbp.user_id = :user_id
+             AND gb.name = 'provider'
+     ) data;
+SQL;
+        $rows = $db->query($query, array(
+            'user_id' => $user->getUserID()
+        ));
+        if ($rows != false && count($rows) > 0) {
```
ryanrath: Alrighty, just checked out the `PDODB::query` function and it looks like this should never return false. If the underlying `PDO::Statement` returns false while executing, `PDODB::query` catches it and throws an `Exception` that includes the sql state, error code and error message. So it looks like the checks for false can be removed.

ryanrath: The same thing ( catching a return value of false and throwing an Exception ) happens with `PDODB::execute` as well. I'll update everything to take that in to account.

classes/DBObject.php
plessbd: I would say the more tests the better.  I just didn't know if I was overlooking existing tests.

classes/Centers.php
ryanrath: Updated in commit: d47e64b8e742444fcd433b391e166bcefb9fea98

classes/Centers.php
```diff
@@ -287,7 +287,7 @@ private static function _listStaffForUser(iDatabase $db, XDUser $user)
             ':user_id' => $user->getUserID()
         ));
 
-        if ($rows !== false && count($rows) > 0) {
+        if (count($rows) > 0) {
             return $rows;
         }
```
plessbd: Is there any reason to bother with this check in the first place?  $db->query returns an array which if no rows should be an empty array, why not just return that?

configuration/etl/etl_sql.d/acls/acl_tabs.sql
```diff
@@ -0,0 +1,1377 @@
+-- mgr acl tabs
+INSERT INTO ${DESTINATION_SCHEMA}.acl_tabs (acl_id, tab_id, position, is_default)
```
ryanrath: Just want to be sure I understand before writing some stuff up. Comments as to what the insert is performing? Or comments on the table? Both? Something else entirely?

configuration/etl/etl_tables.d/acl_assets.json
```diff
@@ -0,0 +1,35 @@
+{
+  "table_definition": {
+    "name": "acl_assets",
+    "engine": "InnoDB",
+    "comment": "Tracks which `acls` have access to which `assets`",
+    "columns": [
+      {
+        "name": "acl_asset_id",
+        "type": "integer",
```
ryanrath: Oooo, I'll make that change in the acl_etl branch ( as that's where this all is now ). Thanks!

classes/DBObject.php
ryanrath: It would certainly be awesome to not have to do the funky name testing for 'get' and 'set' but, at least according to PHP's Overloading page it looks like __get and __set are for properties that cannot be found not functions. If they can be used for functions then I would definitely be all about that.

[_get()](http://php.net/manual/en/language.oop5.overloading.php#object.get)  `__get() is utilized for reading data from inaccessible properties.`
[_set()](http://php.net/manual/en/language.oop5.overloading.php#object.set) `__set() is run when writing data to inaccessible properties.`

classes/CCR/DB/iDatabase.php
```diff
@@ -161,4 +161,5 @@ public function rollBack();
      * @return string The quoted string.
      */
     public function quote($string);
+
```
ryanrath: Definitely, not entirely sure why I would have added this extra line. But I'll make sure to exclude non-functional changes in the future. Specifically for this change, would you prefer it changed back? Or just left as is?

classes/DBObject.php
plessbd: I guess I am not understanding what this code is doing then.

"The following block of code dynamically generates 'getters' and 'setters' based on the properties the class currently supports."

What functions are you trying to get or set?

It looks like you are getting and setting values based on the PROP_MAP

Which seems to be what get and set do from the documentation
```php
public function __get($name)
    {
        echo "Getting '$name'\n";
        if (array_key_exists($name, $this->data)) {
            return $this->data[$name];
        }

        $trace = debug_backtrace();
        trigger_error(
            'Undefined property via __get(): ' . $name .
            ' in ' . $trace[0]['file'] .
            ' on line ' . $trace[0]['line'],
            E_USER_NOTICE);
        return null;
    }

```

html/gui/js/SessionManager.js
ryanrath: Pretty sure this was just editing the file in an editor that defaulted to 2 spaces as opposed to 4. As with the other non-functional change would you prefer this changed back or left as is?

classes/DBObject.php
ryanrath: So I think maybe a use case might help, it goes something like this: 
```php 
/**
 * 
 * This @method block is purely for devs and editors / IDE's that can utilize this info to
 * generate auto-complete entries.
 * 
 * @method string getFirstName()
 * @method void   setFirstName($firstName)
 * @method string getLastName()
 * @method void   setLastName($lastName)
 **/
class Bob extends DBObject
{
    // the 'key' defines what value DBObject will look for when going through the configuration 
    // array passed during instantiation. The 'value' defines what the property name will be used
    // should the 'key' be found in the configuration array. 
    protected $PROP_MAP = array(
        'first_name' => 'firstName',
        'last_name' => 'lastName'
    );
}

$bob = new Bob('first_name' => 'Bob', 'last_name' => 'Smith');
$firstName = $bob->getFirstName(); // Value: 'Bob'
$lastName = $bob->getLastName(); // Value: 'Smith'

// Since `getFirstName` and `getLastName` are not defined, `__call` is executed, 
// the properties  are found and their values returned.

// Without the `__call` function setup when the code evaluates
$firstName = $bob->getFirstName(); 
// No function would be found and it would error out. 
```

classes/NewRest/Controllers/MetricExplorerControllerProvider.php
ryanrath: Yeah... that should have been removed before comitting. I was experiencing some extremely odd behavior with being unable to reach new routes in the AclsControllerProvider and to just test the function I tossed it here since I knew that the routes here were working. The ultimate issue was that the json configuration's for two different controllers had the same key and so set of routes were overwriting another. I updated the key and all was well. Thanks for the catch.

classes/NewRest/Controllers/AclsControllerProvider.php
ryanrath: Because we aren't using the SecurityServiceProvider / don't have it setup? We started with a custom authentication / authorization framework and changing how users are authorized for a particular route wasn't part of the original request. I'm certainly not adverse to using something like the SecurityServiceProvider, it just wasn't part of the scope for this PR.

configuration/etl/etl_tables.d/acl_assets.json
ryanrath: Changes made in acl_etl

configuration/etl/etl_sql.d/acls/acl_tabs.sql
smgallo: I was looking for comments in general. What is the purpose of the queries, what each of them is supposed to be doing. If they are to be run once then that should be stated as well.

classes/DBObject.php
smgallo: You actually don't geed to use `__cal()`,  `__get()`, or `__set()` at all in this case. Since you are already creating the camelcase variables in the public scope, you can just access them directly. E.g., `$firstName = $bob->firstName`. Then you don't have all of this clunky overhead. If they were not in the public scope and in an array for example, then `__get()` and `__set()` would be the way to go. You can see an example in the `IngestorOptions` class.

configuration/etl/etl_tables.d/acl_assets.json
smgallo: Thanks. I really should support the default types for MySQL since it does change them on you internally. I'll add that to the todo list since using `integer` is pretty common.

configuration/etl/etl_sql.d/acls/acl_tabs.sql
ryanrath: Absolutely, coming right up.

classes/Tabs.php
```diff
@@ -39,7 +39,7 @@ public static function getTabs()
      */
     public static function getTab($tabId)
     {
-        if (!isset($tabId)) {
+        if (null === $tabId) {
```
smgallo: Since this check and the `! is_numeric()` both throw an exception with the same message they should be combined into one `if` statement.

classes/User/Acls.php
```diff
@@ -50,7 +50,7 @@ public static function getAcl($aclId)
 
     public static function createAcl(Acl $acl)
     {
-        if (!isset($acl)) {
+        if (null === $acl) {
```
smgallo: This is not necessary since `$acl` must be a valid instance of `Acl` unless you specify the parameter as `Acl $acl = null`

classes/User/Acls.php
```diff
@@ -82,7 +82,7 @@ public static function updateAcl(Acl $acl)
 
     public static function deleteAcl(Acl $acl)
     {
-        if (!isset($acl)) {
+        if (null === $acl) {
```
smgallo: This is not necessary since $acl must be a valid instance of Acl unless you specify the parameter as Acl $acl = null

classes/User/Acls.php
```diff
@@ -121,15 +117,11 @@ public static function listUserAcls(XDUser $user)
...
-        if (!isset($user)) {
-            throw new Exception('A valid user must be provided.');
-        }
-
         if (null == $user->getUserID()) {
             throw new Exception('A valid user id must be provided.');
         }
 
-        if (!isset($aclId)) {
+        if (null === $aclId) {
```
smgallo: Should you check for `is_numeric()` instead of just `null`?

classes/Statistics.php
```diff
@@ -30,14 +30,10 @@ public static function listStatistics()
...
-        if (!isset($user)) {
-            throw new Exception('A valid user is required.');
-        }
-
-        if (!isset($realmName)) {
+        if (null === $realmName) {
             throw new Exception('A valid realm is required.');
         }
-        if (!isset($groupByName)) {
+        if (null === $groupByName) {
```
smgallo: Does `$groupByName` need to be a string or just not `null`. Also need comments on these functions...

configuration/etl/etl_sql.d/acls/acl_tabs.sql
ryanrath: Comments committed to: https://github.com/ubccr/xdmod/pull/97

smgallo: Looking good.

classes/CCR/DB/iDatabase.php
jpwhite4: The final pull request should be rebased and not include spurious unrelated changes to the code.

classes/User/Acls.php
ryanrath: My thought process was to account for something like the following: 
```php
function getMsg()
{
    $msg = null;
    $seed = rand();
    if ($seed % 2 === 0) {
        $msg = "World!";
    }
    return $msg;
}

function hello($msg)
{
    $intro = "Hello,";
    echo "$intro$msg\n";
}

for($i = 0; $i < 10; $i++) {
    $msg = getMsg();
    hello($msg); 
}
```
Basically attempting to account for possible run time shenanigans.

classes/NewRest/Utilities/Authorization.php
```diff
@@ -14,84 +14,38 @@ class Authorization
...
-                }
-
-            }
+        if (count($requirements) === 0) {
+            throw new \Exception('A valid set of requirements are required to complete the requested operation.');
         }
 
+        $found = $user->hasAcls($requirements);
+        $result[self::_SUCCESS] = (!$found && $blacklist) || ($found && !$blacklist);
+        $result[self::_MESSAGE] .= (!$found && !$blacklist) || ($found && $blacklist)
```
jpwhite4: This code is very difficult to understand. It looks like the intent is to use the value of the blacklist variable to toggle the status code (using an xor operation) and the message content depends on the value of the status code.  How about changing it to:
```
$result[self::_SUCCESS] = $user->hasAcls($requirements) xor $blacklist;
$result[self::_MESSAGE] = $result[self::_SUCCESS]  ? '' :  ' [ Not Authorized ]'
```

Or if you are averse to the xor then use an if statement to toggle the return code variable.

classes/User/Acls.php
smgallo: I'm not sure that I understand what you're trying to guard against. You cannot call `createAcl(Acl $acl)` with anything other than a valid instance of an `Acl` object;

```
[smgallo@xsede-dev ~]$ php -r 'class A { public $data = 1; } function test(A $obj) { print_r($obj); } $aObj = new A(); test($aObj); test(1);'
A Object
(
    [data] => 1
)

Catchable fatal error: Argument 1 passed to test() must be an instance of A, integer given, called in Command line code on line 1 and defined in Command line code on line 1

[smgallo@xsede-dev ~]$ php -r 'class A { public $data = 1; } function test(A $obj) { print_r($obj); } $aObj = new A(); test($aObj); test(null);'
A Object
(
    [data] => 1
)

Catchable fatal error: Argument 1 passed to test() must be an instance of A, null given, called in Command line code on line 1 and defined in Command line code on line 1

```

classes/Realms.php
```diff
@@ -0,0 +1,356 @@
+<?php
```
jpwhite4: Don't forget that all new code has to follow the coding standards. I recommend running phpcs on your new files and fixing all of the various errors first before you address the other comments.

classes/User/Acls.php
ryanrath: yeah... I blame the lack of caffeine this morning. I had of course missed that whole 'type' thing in my initial test. You are definitely correct. I'll remove all of the not required checks shortly.

html/gui/js/SessionManager.js
jpwhite4: Rebase it so this change never appears.

classes/XDAdmin.php
```diff
@@ -137,22 +137,21 @@ public function enumerateExceptionEmailAddresses()
...
+              rf.organization_id AS id,
+              o.abbrev           AS organization,
+              o.name             AS name
             FROM
-                modw_aggregates.jobfact_by_quarter,
-                organization o
-            WHERE o.id = organization_id
+              organization o
+              JOIN resourcefact rf
+                ON o.id = rf.organization_id
```
jpwhite4: What was the reason for changing this query? It does not appear to be releated to the necessary ACL changes. The new query looks like it will return different results to the old one. What is the impact of this change? How has it been tested?

classes/NewRest/Controllers/UserControllerProvider.php
```diff
@@ -55,16 +59,33 @@ class UserControllerProvider extends BaseControllerProvider
     public function setupRoutes(Application $app, \Silex\ControllerCollection $controller)
     {
         $root = $this->prefix;
+        $class = get_class($this);
+        $conversions = '\NewRest\Utilities\Conversions';
+
+        $isAuthorized = function (Request $request, Application $app) {
+            $user = $this->getUserFromRequest($request);
+            $authorized = Authorization::authorized($user, array('mgr'));
```
jpwhite4: The function authorized() always returns an array, which contains a status code. This should be checked.

Why don't you change the authorized() function to just directly throw the unauthorized exception, thus reducing the amount of boiler plate code that you have to write and maintain?

classes/NewRest/Controllers/AclsControllerProvider.php
```diff
@@ -0,0 +1,252 @@
...
+    public function setupRoutes(Application $app, ControllerCollection $controller)
+    {
+        $root = $this->prefix;
+        $class = get_class($this);
+        $conversions = '\NewRest\Utilities\Conversions';
+
+        $isAuthorized = function (Request $request, Application $app) {
+            $user = $this->getUserFromRequest($request);
+            $authorized = Authorization::authorized($user, array('mgr'));
+            if (!$authorized) {
```
jpwhite4: The function authorized() always returns an array, which contains a status code. This should be checked.

Why don't you change the authorized() function to just directly throw the unauthorized exception, thus reducing the amount of boiler plate code that you have to write and maintain?

classes/Realms.php
```diff
@@ -0,0 +1,356 @@
...
+        if (strlen($name) < 1) {
+            throw new Exception('Valid realm name required. (length)');
+        }
+        return self::_getRealmByName(
+            DB::factory('database'),
+            $name
+        );
+    }
+
+    public static function getRealmById($realmId)
```
jpwhite4: What is this function used for? I can't find anywhere in the code that calls it.

classes/Statistics.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+use CCR\DB;
+use CCR\DB\iDatabase;
+use Exception;
+use Statistic;
+
+/**
```
jpwhite4: Don't forget that all new classes need to be in a namespace. Please make sure that all of the new classes that you have added are in a namespace. (I note that phpcs will find these errors).

classes/NewRest/Controllers/StatisticsControllerProvider.php
```diff
@@ -0,0 +1,72 @@
...
+
+    public function getPermittedStatistics(Request $request, Application $app)
+    {
+        $user = $this->getUserFromRequest($request);
+        $realmName = $this->getStringParam($request, 'realm', true);
+        $groupByName = $this->getStringParam($request, 'group_by', true);
+        $old = $this->getBooleanParam($request, 'old', false, false);
+
+
+        $statistics = $old == false
```
jpwhite4: This code looks like you are trying to support an 'old' mechanism as well as the new one. Why are you doing this?

classes/Models/Hierarchy.php
```diff
@@ -0,0 +1,23 @@
+<?php namespace Models;
+
+use DBObject;
```
jpwhite4: You should not "use DBObject" it is a class! DBObject needs to be in a namespace and then you can use the namespace that contains it.

classes/NewRest/Controllers/StatisticsControllerProvider.php
ryanrath: Not support so much as test that the new output matches the old. But it obviously shouldn't have made it into the PR.

classes/NewRest/Controllers/AclsControllerProvider.php
ryanrath: Agreed ( and change incoming ) re: checking the status code.
The reason for the funky return format / not throwing an Exception inside authorized was to try and keep the code in BaseControllerProvider.php:276 - BaseControllerProvider.php:290 the same. But there's nothing to say that we can't move this logic into the 'authorized' function itself seeing as how we have access to the XDUser there.

classes/XDAdmin.php
ryanrath: The reason for changing the query was that I ran into issues testing the user admin interface ( namely that you cannot select a resource to associate with a user until after that resource has had a job run under ( for? ) it). At the time, having talked it over with Steve it was decided that instead of only returning organizations that had been referenced by a record in the jobfact_by_quarter table it should return the list of organizations with a corresponding resourcefact entry. But I will say that since the UI changes are not included in this PR it seems like this should be reverted until they are added.

jpwhite4: Even if you were making UI changes, I would still want this change to be in a separate pull request.

classes/NewRest/Controllers/StatisticsControllerProvider.php
jpwhite4: You should add unit and integration tests to check that the new output matches the old output.

classes/NewRest/Utilities/Authorization.php
```diff
@@ -21,31 +23,30 @@ class Authorization
...
     * @param array $requirements an array of acl names
     * @param bool $blacklist     whether or not to test for the presence of the $requirements or the absence
-    * @return array              in the form:
-    *                            array(
-    *                                0 => <success>,
-    *                                1 => <message>
-    *                            );
-    * @throws \Exception if the user provided is null
-    *                    if the requirements is not an array or it is an array but has no contents
+    * @throws \Exception                if the user provided is null
```
plessbd: User cant be null and it doesn't throw an exception for that.  Please remove.

classes/DBObject.php
```diff
@@ -0,0 +1,128 @@
...
+ * tables. It does this via providing a few conveniences for working with the
+ * results of sql queries ( string keyed array ) and populating the object
+ * properties from these results. It also provides dynamic 'getter' and 'setter'
+ * support for calls that follow the form 'getCamelCasePropertyName()' and
+ * 'setCamelCasePropertyName($propertyName)' the property name is assumed to be
+ * in the form: lcfirst(CamelCase(column_name)) => columnName.
+ *
+ * @author Ryan Rathsam <ryanrath@buffalo.edu>
+ */
+class DBObject implements \ArrayAccess, JsonSerializable {
```
ryanrath: Because after a conversation with @smgallo on the purpose / features of DBObject he had requested that it be able to be used as an array. Now with that said, per a comment of his on this PR I believe he's leaning more towards just using the properties defined in the constructor and not having methods or ArrayAccess but I'm certainly open to a discussion on what direction this class goes.

smgallo: If @smgallo remembers correctly, he meant storing the data in an array and then using `__get()` and `__set()` to access the data similar to how it is implemented in `ETL/aOptions` rather than providing an array iterator.

ryanrath: *\*facepalm\** Don't mind me, I'll just be over here, making those changes.

classes/XDAdmin.php
ryanrath: fair enough, I can certainly see the reasons for wanting it broken out.

classes/Realms.php
ryanrath: Yeah, it looks like both getRealmByName and getRealmById are not currently used. Not sure what getRealmByName was about but I believe getRealmById was to support basic CRUD operations:
- GET /realms : list all currently installed realms
- GET /realms/{realm_id} : list information about realm identified by realm_id
- POST /realms : create a new realm
- DELETE /realms/{realm_id} : delete the realm identified by realm_id
- PATCH|PUT /realms/{realm_id} : update the realm identified by realm_id

etc. etc. 
Seeing as how the route(s) that would utilize this function doesn't exist it pretty obviously doesn't need to be here. I'll go ahead and remove it.

classes/NewRest/Controllers/StatisticsControllerProvider.php
ryanrath: agreed, I'll make sure to have those added in the new smaller PR's

PR #97, opened 2017-04-13
Modular Acls: ETL Support

ryanrath:
Adding in the files required to bootstrap / support the new modular Acl system.

## Description
There are two ETL pipelines:
- acls::AclTableManagement: Can be run any number of times, will keep the db structure up to date. 
- acl_bootstrap::AclTableImport: Should only be run once. Boostraps the tables with their initial data.

Included are the table definition files ( etl/etl_tables.d/acls/*.json ) and the scripts for populating them 
( etl/etl_sql.d/acls/*.sql ). Where possible the scripts attempt to only insert records when a record that matches the values being inserted is not found. 

## Motivation and Context
To serve as the basis for the rest of the Acl system ( and to provide a baseline for the rest of the broken up PR's )

## Tests performed
manual tests on a dev cloud instance installing locally. 

## Types of changes

## Checklist:


configuration/etl/etl_sql.d/acls/modules.sql
```diff
@@ -0,0 +1,4 @@
+CALL ${DESTINATION_SCHEMA}.add_module_with_version('xdmod', 'XDMoD', 6, 5, 0, '');
```
jpwhite4: Is this code tying to define the version numbers of the installed modules? The version numbers specified here are never going to be correct since this code will not be backported to the 6.5.0 release.

ryanrath: That is definitely what the code is doing. One clarification I'd like to make is that the versions are stored historically ( in the module_versions table with a reference to the current version in modules  ). So we can either leave the call as it currently while adding a line to call `add_module_version` thereby having a record of both versions or we can just update it whatever the current version is when the code is just prior to integrating it into the system.

jpwhite4: Surely the lines that define version numbers for the (optional) modules should be stored in the source code repos for the respective modules. We also already define the version number in the build.json files, does this change mean that we now have to maintain the same information in two different locations?

ryanrath: I definitely agree that the data specific to a particular module should reside within that module. That was certainly an oversight on my part and I'll work on moving the module specific data into its respective modules. In so far as duplicating data in build.json and some where else, that's something I definitely don't want to do. I'll see if there's a way we can utilize the modules build.json information in constructing the module creation sql. I'm currently working with Jeff to figure out the best way to have these sql statements executed on module install.

configuration/etl/etl.d/acls-xdmod-management.json
```diff
@@ -14,7 +14,7 @@
       }
     }
   },
-  "acls": [
+  "acls-xdmod-management": [
     {
       "#": "AclTableManagement is meant to be run more than once in it's lifetime. Doing so will keep the associated tables up to date.",
```
jpwhite4: Please state when the acl-xdmod-management is intended to be run. I assume that you should run it every time the software is updated to a new version or a new module is installed.

ryanrath: I'll update the description to state more exactly when it is intended to be run. Which is basically whenever the structure of one of the tables it manages changes. New module installation may not trigger a run as hopefully they're just inserting data into the existing tables not modifying their structure.

classes/OpenXdmod/Setup/AclImport.php
```diff
@@ -16,7 +20,7 @@ public function handle()
 to input the values to the best of your knowledge.
```
jpwhite4: The phrase "to the best of your knowledge" is redundant.

bin/acl-xdmod-import
```diff
@@ -0,0 +1,168 @@
...
+CWD="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )";
+
+# Make sure that they have PHP installed.
+if [ -z "$PHP" ];then
+    echo "Unable to find 'php' in the current path. Please remedy this and run the script again.";
+    exit;
+fi
+
+# Parse the command line arguments.
+while [[ $# -gt 0 ]]
```
jpwhite4: How come you did not use getopt to parse the options?

classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -96,7 +96,7 @@ public function initialize(EtlOverseerOptions $etlOverseerOptions = null)
         $etlTableKey = key($this->etlDestinationTableList);
         if ( count($this->etlDestinationTableList) > 1 ) {
             $msg = $this . " does not support multiple ETL destination tables, using first table with key: '$etlTableKey'";
-            $logger->warning($msg);
+            $this->logger->warning($msg);
```
jpwhite4: This looks like a bug fix for ETL code. If so then does it really belong in this pull request?

configuration/etl/etl_sql.d/acls/xdmod/module.sql
```diff
@@ -0,0 +1,9 @@
+-- =============================================================================
+-- NAME:      modules.sql
+-- EXECUTION: once on installation
+-- PURPOSE:   Provides some initial data for the module / module_versions table.
+--            This file was generated manually based on knowledge of the
+--            currently supported modules.
+-- =============================================================================
+
+CALL ${DESTINATION_SCHEMA}.add_module_with_version('xdmod', 'XDMoD', 6, 5, 0, '');
```
jpwhite4: Again. I would like to see this file being autogenerated by the build process.

bin/acl-xdmod-import
ryanrath: Because this bit of code accomplished what I needed. Are there some common cases where this will fall down / I should look at getopt?

ryanrath:
This latest commit moves as much data as is feasible out of the sql files and into json files ( although the sql files are still present in case they are needed for reference / alternative methods ). I'm going to begin looking at how we can hook into the build process and snag the module version so that we don't have to have it defined in more than one place.

classes/OpenXdmod/Build/Config.php
```diff
@@ -228,6 +228,47 @@ private static function normalizeFileMapDestination($src, $dest)
...
+     *     NUMBERS-ALPHA_NUMERIC
+     *     NUMBERSALPHAN_NUMERIC
+     *     etc.
+     * @param string $version the version as provided by the file 'build.json'
+     **/
+    private function setVersionDetails($version)
+    {
+        $MAJOR = 1;
+        $MINOR = 2;
+        $MICRO = 3;
```
plessbd: conventionally this should be called patch I believe (http://semver.org/)

configuration/etl/etl_sql.d/acls/xdmod/module.sql
ryanrath: With the latest commit files in the configuration folder ( as specified on a module by module basis ) can now take advantage templating to get access to module version information. This includes the whole version string as well as each part of the version broken out so it can be referenced independently.

classes/OpenXdmod/Build/Config.php
ryanrath: 'micro' ( and it's associated function ) have been renamed 'patch' as per semver.

configuration/etl/etl.d/acls-xdmod-management.json
```diff
@@ -0,0 +1,44 @@
...
+          "name": "Database",
+          "config": "database",
+          "schema": "moddb"
+        }
+      }
+    }
+  },
+  "acls-xdmod-management": [
+    {
+      "#": "AclTableManagement is meant to be run more than once in it's lifetime. Doing so will keep the associated tables up to date.",
```
jpwhite4: Please change this comment to be more specific about when the AcLTableManagement should be run.

For example:
AclTableManagment should always be run after any changes are made to the definition files listed below.

configuration/etl/etl_data.d/acls/xdmod/acl-tabs.json
```diff
@@ -0,0 +1,89 @@
...
+        ["mgr","app_kernel_viewer","app_kernels",null,null],
+        ["mgr","app_kernel_explorer","app_kernels",null,null],
+        ["mgr","app_kernel_notification","app_kernels",null,null],
+        ["mgr","my_allocations",null,null,null],
+        ["usr","tg_summary",null,null,null],
+        ["usr","tg_usage",null,null,null],
+        ["usr","metric_explorer",null,null,null],
+        ["usr","report_generator",null,null,null],
+        ["usr","about_xdmod",null,null,null],
+        ["usr","app_kernels",null,null,null],
```
jpwhite4: We should not reference optional sub-modules in the main framework

configuration/etl/etl_data.d/acls/xdmod/tabs.json
```diff
@@ -0,0 +1,26 @@
...
+        ["tg_summary","Summary","100","1","XDMoD.Module.Summary","CCR.xdmod.ui.tgSummaryViewer","Displays summary information","Summary Tab",null,"xdmod" ],
+        ["tg_usage","Usage","200","0","XDMoD.Module.Usage","CCR.xdmod.ui.chartViewerTGUsage","Displays usage","Usage Tab",null, "xdmod" ],
+        ["metric_explorer","Metric Explorer","300","0","XDMoD.Module.MetricExplorer","CCR.xdmod.ui.metricExplorer","","Metric Explorer",null, "xdmod" ],
+        ["report_generator","Report Generator","1000","0","XDMoD.Module.ReportGenerator","CCR.xdmod.ui.reportGenerator","","Report Generator",null, "xdmod" ],
+        ["about_xdmod","About","10000","0","XDMoD.Module.About","CCR.xdmod.ui.aboutXD","","About",null, "xdmod" ],
+        ["job_viewer","Job Viewer","5000","0","XDMoD.Module.JobViewer","CCR.xdmod.ui.jobViewer","View detailed job-level metrics","Job Viewer",null, "xdmod" ],
+        ["app_kernels","App Kernels","400","0","XDMoD.Module.AppKernels","CCR.xdmod.ui.appKernels","Displays data reflecting the reliability and performance of grid resources","App Kernels", null,"xdmod" ],
+        ["app_kernel_viewer","App Kernel Viewer","100","0","","","","","app_kernels","xdmod" ],
```
jpwhite4: Should not see app kernel tabs in here.

configuration/etl/etl_data.d/acls/xdmod/tabs.json
```diff
@@ -0,0 +1,26 @@
...
+        ["app_kernel_explorer","App Kernel Explorer","200","0","","","","","app_kernels","xdmod" ],
+        ["app_kernel_notification","App Kernel Explorer","300","0","","","","","app_kernels","xdmod" ],
+        ["my_allocations","Allocations","350","0","XDMoD.Module.Allocations","CCR.xdmod.ui.AllocationViewer","Displays your allocation usage","Allocations Tab",null, "xdmod" ],
+        ["compliance","Compliance","2000","0","XDMoD.Module.Compliance","CCR.xdmod.ui.complianceTab","","Compliance Tab",null, "xdmod" ],
+        ["custom_query","Custom Queries","3000","0","XDMoD.Module.CustomQueries","CCR.xdmod.ui.customQuery","","Custom Queries Tab",null, "xdmod" ],
+        ["sci_impact","Sci Impact","4000","0","XDMoD.Module.SciImpact","CCR.xdmod.ui.impact","Scientific Impact by user, organization, and project","Sci Impact Tab",null, "xdmod"]
```
jpwhite4: Should not see sci impact tab here. They are xsede-specific

configuration/etl/etl_data.d/acls/xdmod/acl-tabs.json
ryanrath: Absolutely, this has been resolved in the recent commits.

configuration/etl/etl_data.d/acls/xdmod/tabs.json
ryanrath: Absolutely, this has been resolved in the recent commits.

configuration/etl/etl_data.d/acls/xdmod/tabs.json
ryanrath: Absolutely, this has been resolved in the recent commits.

configuration/etl/etl_sql.d/acls/xdmod/statistics.sql
```diff
@@ -0,0 +1,3697 @@
+INSERT INTO statistics (module_id, realm_id, name, display, alias, unit, decimals, formula, description)
```
jpwhite4: This file should not be in the pull request (nor should any of the other SQL files that are autogenerated).

classes/OpenXdmod/Build/Config.php
```diff
@@ -298,6 +310,40 @@ private static function normalizeFileMapDestination($src, $dest)
...
+        $PATCH = 3;
+        $PRE_RELEASE = 4;
+
+        $matches = array();
+        preg_match("/(\d+)?\.(\d+)?\.?(\d+)?\.?-?([0-9A-Za-z-.]+)?/", $version, $matches);
+        $length = count($matches);
+        for ($i = 1; $i < $length; $i++) {
+            switch ( $i ) {
+                case $MAJOR:
+                    $this->versionMajor = trim($matches[$i]);
```
jpwhite4: No need to trim here since the regex cannot match any characters that would be trimmed.

classes/OpenXdmod/Build/Config.php
```diff
@@ -298,6 +310,40 @@ private static function normalizeFileMapDestination($src, $dest)
...
+                    break;
+                case $MINOR:
+                    $this->versionMinor = trim($matches[$i]);
```
jpwhite4: No need to trim here since the regex cannot match any characters that would be trimmed.

classes/OpenXdmod/Build/Config.php
```diff
@@ -298,6 +310,40 @@ private static function normalizeFileMapDestination($src, $dest)
...
+                    break;
+                case $PATCH:
+                    $this->versionPatch = trim($matches[$i]);
```
jpwhite4: No need to trim here since the regex cannot match any characters that would be trimmed.

classes/OpenXdmod/Build/Config.php
```diff
@@ -298,6 +310,40 @@ private static function normalizeFileMapDestination($src, $dest)
...
+                    break;
+                case $PRE_RELEASE:
+                    $this->versionPreRelease = trim($matches[i]);
```
jpwhite4: No need to trim here since the regex cannot match any characters that would be trimmed.

configuration/etl/etl_sql.d/acls/xdmod/statistics.sql
ryanrath: *nod* removed in bfa9434 ( removed all sql that isn't explicitly being used by the import pipeline. ) pulled / deployed to my local box and ran a quick import to make sure things were still working as expected. So far so good.

classes/OpenXdmod/Build/Packager.php
```diff
@@ -433,34 +417,22 @@ private function processModuleFiles()
...
+        $source = CONFIG_DIR;
 
-            $source = implode(
-                array_merge(array($baseDirectory), $directoryParts, array($this->config->getName())),
-                DIRECTORY_SEPARATOR
-            );
+        $destination = implode(
+            array($this->getPackageDir(), "configuration"),
+            DIRECTORY_SEPARATOR
+        );
```
jtpalmer: I would prefer if the `implode` parameters were reversed.

From http://php.net/implode:

> Note:
implode() can, for historical reasons, accept its parameters in either order. For consistency with explode(), however, it may be less confusing to use the documented order of arguments.

classes/OpenXdmod/Build/Packager.php
```diff
@@ -433,34 +417,22 @@ private function processModuleFiles()
...
 
-            $destination = implode(
-                array_merge(array($this->getPackageDir(), "configuration", "etl"), $directoryParts, array($this->config->getName())),
-                DIRECTORY_SEPARATOR
-            );
```
jtpalmer: Another reversed `implode`.

classes/OpenXdmod/Build/Packager.php
```diff
@@ -497,7 +469,11 @@ private function replaceTemplateValues($source, $destination, $fileRegex, array
                     $contents = str_replace($key, $value, $contents);
                 }
 
-                $destinationFile = implode(array($destination, basename($file)), DIRECTORY_SEPARATOR);
+                $baseDest = substr($file, strlen(CONFIG_DIR) + 1);
+                $fileName = pathinfo($file, PATHINFO_FILENAME);
+                $subDirectory = substr($baseDest, 0, strpos($baseDest, $fileName) - 1);
+
+                $destinationFile = implode(array($destination, $subDirectory, $fileName), DIRECTORY_SEPARATOR);
```
jtpalmer: Another reversed `implode`.

jpwhite4:
Don't forget that you also need to create pull requests for the necessary changes to the various XDMoD modules: appkernels, xsede and supremm. Presumably these will all have to go in at the same time in order for stuff to work properly.

configuration/etl/etl_data.d/acls/xdmod/module-versions.json.template
```diff
@@ -0,0 +1,9 @@
+{
+    "table_definition": {
+        "$ref": "etl_tables.d/acls/module_versions.json#/table_definition"
+    },
+    "destination_columns": ["version_major", "version_minor", "version_micro", "version_patch"],
+    "source_values": [
+        ["__VERSION_MAJOR__"," __VERSION_MINOR__"," __VERSION_PATCH__", "__VERSION_PRE_RELEASE__"]
```
jpwhite4: Is the whitespace before __VERSION_MINOR__ and __VERSION_PATCH__ intentional here?

ryanrath:
Absolutely, I have all of the pull requests for the other modules ready to go. I'll put them up / reference them here when we get back from lunch.

docs/configuration.md
```diff
@@ -99,6 +99,43 @@ groups based on users assigned to one of the groups.
...
+
+It is intended that this action be run multiple times during the
+lifetime of an XDMoD installation. Specifically, anytime the structure
+of the Acl tables changes, running this action will ensure that the
+tables in the installations database are kept in sync with these
+changes.
+
+A script has been provided for your convenience if you wish to execute
+this action outside of xdmod-setup. It is located in the
+'bin' directory and is called 'acl-xdmod-setup'.
```
jpwhite4: This documentation looks incorrect: There is no acl-xdmod-setup script.

docs/configuration.md
```diff
@@ -99,6 +99,43 @@ groups based on users assigned to one of the groups.
...
+verbosity of the action then you will need to modify the defaults
+provided.
+
+It is intended that this action only be run *once* during the lifetime
+of an XDMoD installation. Once the data has been added to the database
+there should be no need to execute this action again.
+
+A script has been provided for your convenience should you wish to
+execute this action outside of xdmod-setup. It is located in the
+'bin' directory and is called 'acl-xdmod-import'.
```
jpwhite4: There is no acl-xdmod-import script in this pull request.

classes/OpenXdmod/Setup/AclImportXdmod.php
```diff
@@ -0,0 +1,7 @@
+<?php namespace OpenXdmod\Setup;
+
+class AclImportXdmod extends AclImport
+{
+    protected $scriptName = 'acl-xdmod-import';
```
jpwhite4: Where is the acl-xdmod-import script that is referenced here?

jpwhite4:
The file configuration/etl/etl_data.d/acls/xdmod/acl-group-bys.json is 32k of data that enumerates every combination of each statistic for each groupby for each user role. This huge file is impossible for a human to be able to read and understand. The structure is such that it is going to be easy for a person to make a mistake when they have to edit it to add a new groupby or statistic.  Adding a new statistic is going to involve editing hundred of lines of text. How do you know if you correctly added the new statistic to all possible combinations that you have to? How is a reviewer expected to read and understand a nearly 1000 line diff?

open_xdmod/modules/xdmod/tests/lib/OpenXdmod/Tests/Build/PackagerTemplatingTest.php
```diff
@@ -0,0 +1,168 @@
...
+        );
+        $cmd = implode(' ', array_map('escapeshellarg', $rmArgs));
+        $this->executeCommand($cmd);
+
+        // Change the current working directory back to where we were at the
+        // beginning of the script.
+        chdir($cwd);
+    }
+
+    private function getPrivateProperty($object, $propertyName)
```
jpwhite4: Edit: I sse that it is gettting a property value anf not a function call. Please move this function into \TestHelpers\TestHelper so that other tests can use it.

configuration/setup.json
```diff
@@ -36,6 +36,16 @@
             "handler": "UpdateCheckSetup"
         },
         {
+            "position": 80,
```
jpwhite4: Why is this a separate option? Surely the aclsetup is mandatory and should be run when the datawarehouse is setup. If we give the end user an extra hoop to jump through then they will almost certainly forget to do it and we'll have more support requests to deal with.

ryanrath: I would love to have this be automated. Having it added as options in the xdmod-setup script was actually a fall back as in my conversations with Jeff there didn't seem to be a way to have the requisite actions taken on module install currently ( which was my first choice / seemed the most logical place for this to happen ). I'd definitely be down with chatting about this if you happen to have thoughts on how this could be accomplished.

ryanrath:
Re: acl-group-bys.json I can use the sql file instead for this. It still clocks in at almost 4k lines ( pretty printed sql ) but it's certainly smaller than the json. I could also compress each sql statement to a single line which would take it down to ~ 150ish lines.

smgallo:
Re: acl-group-bys.json, while compressing it down to individual lines would help, we should look into a better way to specify the information after this release. I think it's OK to include this file for this release, but there is probably a better way to go about it and if we discuss as a group we may come up with a better solution.

configuration/setup.json
jpwhite4: Just make the code run in the Database setup section. Also make the code be run automatically by xdmod-upgrade.

ryanrath:
@jpwhite4 @jtpalmer I've just finished the re-work / cleanup / update for this PR. If you wouldn't mind taking another look when you get a chance that would rock, thanks!

bin/acl-config
```diff
@@ -0,0 +1,1709 @@
...
+use CCR\DB;
+use CCR\DB\iDatabase;
+use CCR\Log;
+use CCR\Json;
+
+use User\Roles;
+use Xdmod\Config;
+
+$opts = array(
+    'r' => 'dryrun',
```
smgallo: Please change the key here to `'t'` to be consistent with other XDMoD scripts

classes/OpenXdmod/Migration/Version660To700/DatabaseMigration.php
```diff
@@ -0,0 +1,131 @@
...
+    private $configFile;
+
+    /**
+     * @see \OpenXdmod\Migration\Migration::__construct
+     **/
+    public function __construct($currentVersion, $newVersion)
+    {
+        parent::__construct($currentVersion, $newVersion);
+
+        $this->baseDirectory = realpath(BASE_DIR.DIRECTORY_SEPARATOR.'..');
```
smgallo: Please add spaces around concatenation operator here and in the lines below.

configuration/etl/etl_tables.d/acls/acl_group_bys.json
```diff
@@ -0,0 +1,57 @@
...
+        "nullable": false
+      },
+      {
+        "name": "statistic_id",
+        "type": "int(11)",
+        "nullable": false
+      },
+      {
+        "name": "visible",
+        "type": "boolean",
```
smgallo: Use `tinyint(1)` instead of `boolean` because MySQL converts `boolean` under the hood and the information schema actually contains `tinyint(1)`. This will cause continual alter table statements to be generated until I address the conversions.

configuration/etl/etl.d/acls-import.json
```diff
@@ -0,0 +1,46 @@
...
+                }
+            }
+        },
+        "acls-import": {
+            "namespace": "ETL\\Maintenance",
+            "options_class": "MaintenanceOptions",
+            "class": "ExecuteSql"
+        }
+    },
+    "#": "",
```
smgallo: Remove unused comment

bin/acl-import
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env php
```
jpwhite4: Why do you have a php script that executes a subshell to run a program? Surely  this could be a shell script with the following content:
```
acl-etl -pacls-import -g
```
Is this wrapper really needed? You could just call the acl-etl script directly.

bin/acl-xdmod-management
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env php
```
jpwhite4: Why do you have a php script that executes a subshell to run a program? Surely this could be a shell script with the following content:
```
acl-etl -pacls-xdmod-management -g
```
Is this wrapper even needed? Surely you can just call the acl-etl script directly?

classes/OpenXdmod/Migration/Version660To700/DatabasesMigration.php
```diff
@@ -0,0 +1,101 @@
...
+     *     - CONF_DIR/roles.[d|json]
+     *     - CONF_DIR/hierarchies.[d|json]
+     *
+     * @return void
+     **/
+    public function populateTables()
+    {
+        $scripts = array(
+            'acl-config' => array(),
+            'acl-import' => array(
```
jpwhite4: The acl-import script does not accept any command line arguments. Why are you passing arguments here and what are they supposed to do?

classes/OpenXdmod/Migration/Version660To700/DatabasesMigration.php
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Attempt to migrate the acl tables into the current system database. This process will
+     * utilize the script acl-xdmod-management which in turn utilizes the ETL overseer.
+     *
+     * @return void
+     **/
+    public function migrateTables()
+    {
+        $scripts = array(
+            'acl-xdmod-management' => array(
```
jpwhite4: acl-xdmod-management does not accept commandline arguments. Why are they being passed?

classes/OpenXdmod/Migration/Version660To700/DatabasesMigration.php
```diff
@@ -0,0 +1,101 @@
...
+        foreach ($scripts as $scriptName => $params) {
+            $cmd = "$scriptName ". implode(' ', $params);
+            $this->logger->info("Executing $cmd");
+
+            $output = shell_exec($cmd);
+
+            $hadError = strpos($output, 'error') !== false;
+
+            if ($hadError == true) {
+                $this->logger->debug(<<<MSG
```
jpwhite4: You need to log errors at a higher level than debug.

classes/OpenXdmod/Setup/AclImport.php
```diff
@@ -0,0 +1,61 @@
...
+        $sectionMessage = $this->sectionMessage.$scriptName;
+        $etlSection = $this->defaultSection;
+
+        $configFile = realpath(CONFIG_DIR.'/etl/etl.json');
+
+        $this->console->displaySectionHeader($this->sectionHeader);
+        $this->console->displayMessage($sectionMessage);
+        $this->console->displayBlankLine();
+
+        $cmd = "$scriptName -g";
```
jpwhite4: These scripts do not accept commandline arguments so they will never see the -g option. Is this a problem?

open_xdmod/modules/xdmod/tests/lib/TestHelpers/TestHelper.php
```diff
@@ -25,4 +25,20 @@ public static function unlockMethod($classOrObject, $methodName)
         $method->setAccessible(true);
         return $method;
     }
+
```
jpwhite4: Why has this unlockProperty function been added in this pull request? It does not appear to be used anywhere.

ryanrath: It had been used in the now removed PackagerTemplatingTest class. We had a whole conversation about it at the time and how it belonged in the TestHelper class not in the Test Class itself. Which would also explain why it's still here. I'll go ahead and remove it.

ryanrath:
@smgallo jobs_[cloud|hpc].json re-added. hpc is minus the "XdcdbJobRecordIngestor" which has been added to a jobs_hpc.json file in the xsede repo ( under the same pipeline name ).

smgallo:
@ryanrath I need to give this some thought because the order of execution is important (the jobs ingestor action has to run before the aggregator action) and when the actions are added to the pipeline they will be added to the end. Lets discuss tomorrow.

PR #100, opened 2017-04-20
Initial ETLv2 Support for RFC-6901 JSON References

smgallo:
References in JSON files will allow us to reference and include entities defined in external files into the current file, supporting re-use.

Note that the main changes are in `classes/ETL/Configuration`. Most other changes are due to style fixes and the namespace change from `ETL` to `ETL/Configuration`.

**See also ubccr/xdmod-xsede#28**

## Description

A reference follows the format specified in [IETF Draft JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03). A JSON Reference is a JSON object containing a **single** key "$ref", which has a JSON string value that is an [RFC-3896](https://tools.ietf.org/html/rfc3986) URI identifying the location of the JSON value being referenced. If the URI does not specify a scheme, `file` is assumed.  Resolution of a JSON Reference object **SHOULD** yield the referenced JSON value.  E.g., The entire object `{ "$ref": "http://example.com/example.json#/foo/bar" }` gets logically replaced with the thing that it points to.

For example:

- `{ "$ref": "table.json" }` includes the entire document defined in `table.json`
- `{ "$ref": "table.json#/foo/0" }` includes the the key "0" under "foo" if "foo" is an object or
  the first array element if "foo" is an array.
- `{ "$ref": "#/foo" }` is a reference to the key "foo" in the current document
- `{ "$ref": "http://example.com/example.json#/foo" }` is a reference to the key "foo" in the
  document found at http://example.com/example.json
- `{ "$ref": "example.json#/foo/bar" }` is a reference to the key "bar" under the key "foo" in the
  document `example.json` in the current (relative) location.
- `{ "$ref": "file:///path/to/example.json#/foo/bar" }` is a reference to the key "bar" under the
  key "foo" in the document `/path/to/example.json`.

See `docs/etl/json-references.md` for additional details.

### Handling of Local ETL Definition Files (e.g., files located in `etl.d`)

The mechanism for parsing local configuration files has been moved from the XDMoD-specific `EtlConfiguration` class into the parent `Configuration` class. This provides a more general mechanism for handling JSON configuration files and allows classes that extend `Configuration` to provide a smaller code footprint. To make it easier to extend this `Configuration` for task-specific configuration files, the following methods are available for overwriting by child classes. This allows additional processing and/or interpretation of the data.

1. `preTransformTasks()`
    Perform any pre-transformation tasks such as setting or adding key transformers
2. `interpretData()`
   Apply any logic to interpret the configuration data such as handling defaults, registering ETL actions, etc.
3. `processLocalConfig()`
    Local configuration files should be processed by instantiating the appropriate class(typically the class used to process the main file) and passing the correct options.
4. `merge()`
    Merge data from local configuration files into the global namespace. By default, new keys are created as needed and if the key exists in the the merge target and the value can be appended (i.e., is an array) then append the data found in the local configuration to the existing data.
5. `cleanup()`
    Perform any necessary data structure cleanup.

During initialization, the `Configuration` class performs the following algorithm, allowing child classes to extend key pieces of the process to suit their needs.

```php
// Parse the JSON configuration file
$this->parse();

// Perform any post-parsing tasks such as validation or manipulating the parsed
// data that must happen before transformation.
$this->preTransformTasks();

// Run the Key Transformers on the data
$this->transform();

// Perform local interpretation on the data and apply contextual meaning
$this->interpretData();

// Process any local configuration files
foreach ($localFileList as $file ) {
    // Load and parse the configuration file using the correct class
    $localConfigObj = $this->processLocalConfig($file);

    // Merge data found in the local configuration file into the global namespace.
    $this->merge($localConfigObj);
    $localConfigObj->cleanup();
}
```

### Additional Changes

1. The warning for duplicate data endpoint names has been removed. It is perfectly acceptable to re-define a data endpoint in multiple locations (only one will actually be instantiated).
2. Removed code and data structures that were no longer needed, general code cleanup.
3. Moved configuration classes from `ETL` namespace to `ETL\Configuration`
4. Added documentation for JSON reference handling and the ETL table comparison script

## Motivation and Context

As ETL gets more complicated, a way to reference and reuse files in multiple locations was required.

## Tests performed

To test the effect of the changes to support JSON references (e.g., JSON pointers) I created an `\ETL\EtlConfiguration` object with and without modifications to support references and dumped all of the pipelines that were created in the `etc/etl/etl.json` and all files in the local `etc/etl/etl.d` directory for each of the objects. Note that the code that supports references also removes all comments (i.e., "#" keys) from the files where the original code did not.

1. Stash any changes added to support JSON references
2. Deploy the baseline code and run the test script
3. Apply the stashed changes
4. Deploy the new code and run the test script
5. Compare the result.
6. Verify that parsing for other pipelines works

Test script:

```
#!/usr/bin/env php
<?php
/**
 * Test wrapper for the EtlConfiguration class.
 *
 * @author Steve Gallo <smgallo@buffalo.edu>
 */

require __DIR__ . '/../../configuration/linker.php';
restore_exception_handler();

use CCR\Log;

$conf = array(
    'emailSubject' => gethostname() . ': XDMOD: Data Warehouse: Federated ETL Log',
    'mail' => false,
    'consoleLogLevel' => Log::NOTICE
);

$logger = Log::factory('Configuration', $conf);

$config = new \ETL\EtlConfiguration(
    "/home/smgallo/xdmod-json-refs-test/etc/etl/etl.json",
    "/home/smgallo/xdmod-json-refs-test/etc/etl",
    // Remove this line when testing original code
    $logger
);


try {
    $config->initialize();
} catch ( Exception $e ) {
    exit($e->getMessage() . PHP_EOL);
}

foreach ($config as $name => $actions ) {
    print "PIPELINE: '$name',\nACTIONS: " . print_r($actions, true) . PHP_EOL;
}

?>
```

### Testing results

Note that the only difference is the comments that the new code strips out.

```
$ git stash save "Changes to support config file references"

$ ~/src/ccr-private-xdmod/scripts/deployment/update-open-xdmod-portal.php --source-dir ~/src/xdmod --conf-dir ~/src/ccr-private-xdmod --prefix ~/xdmod-no-json-refs-test --xsede && cp ~/portal_settings_tasdb1.ini ~/xdmod-json-refs-test/etc/portal_settings.ini

$ cd ~/xdmod-json-refs-test/share/tools/etl
$ php test-transforms.php | grep -v "Duplicate Data Endpoint" > /tmp/xdmod-no-json-refs.log

$ git stash apply stash@{0}
On branch etl/config-file-references
Your branch is up-to-date with 'upstream/xdmod6.6'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   ../../classes/ETL/Configuration.php
        modified:   ../../classes/ETL/EtlConfiguration.php
        modified:   ../../configuration/portal_settings.ini
        modified:   etl_overseer.php

$ ~/src/ccr-private-xdmod/scripts/deployment/update-open-xdmod-portal.php --source-dir ~/src/xdmod --conf-dir ~/src/ccr-private-xdmod --prefix ~/xdmod-json-refs-test --xsede && cp ~/portal_settings_tasdb1.ini ~/xdmod-json-refs-test/etc/portal_settings.ini

$ cd ~/xdmod-json-refs-test/share/tools/etl
$ php test-transforms.php | grep -v "Duplicate Data Endpoint" > /tmp/xdmod-json-refs.log

$ ls -l /tmp/xdmod-no-json-refs.log /tmp/xdmod-json-refs.log 
-rw-r----- 1 smgallo smgallo 78847 Apr 17 13:15 /tmp/xdmod-json-refs.log
-rw-r----- 1 smgallo smgallo 79906 Apr 17 12:59 /tmp/xdmod-no-json-refs.log

$ diff /tmp/xdmod-no-json-refs.log /tmp/xdmod-json-refs.log 
740d739
<                     [#] => By default, ingestors use unbuffered queries
857d855
<                     [#] => Be sure to exclude the same resources as the ingestor or their records will be deleted here
1124d1121
<                     [#] => Table definition relative to paths.table_config_dir if path not specified
1231d1227
<                     [#] => Manage multiple tables not created by an ingestor or aggregator
1277d1272
<                     [#] => Perform setup for the Custom Queries tab
1352d1346
<                     [#] => Table definition relative to paths.table_config_dir if path not specified
1390d1383
<                     [#] => ResourcePercentToXSEDEIngestor will set percent=0 if skipping a resource. Override that here.
1525d1517
<                     [#] => Table definition relative to paths.table_config_dir if path not specified
1683d1674
<                     [#] => These resources are not in production but still listed as active
1733d1723
<                     [#] => Table definition relative to paths.table_config_dir if path not specified
1774d1763
<                     [#] => These resources are not in production but still listed as active

$ php etl_overseer.php -c ../../../etc/etl/etl.json -p cloud-jobs -t
$ php etl_overseer.php -c ../../../etc/etl/etl.json -p xdcdb-jobs -t
$ php etl_overseer.php -c ../../../etc/etl/etl.json -p resource-allocations -t
```

## Types of changes

## Checklist:


classes/ETL/Configuration/Configuration.php
```diff
@@ -0,0 +1,811 @@
...
+     */
+
+    protected function processLocalConfig($localConfigFile)
+    {
+        $options = array(
+            'local_config_dir' => $this->localConfigDir,
+            'is_local_config' => true
+        );
+
+        $localConfigObj = new Configuration($localConfigFile, $this->baseDir, $this->logger, $options);
```
jpwhite4: Does this construct prevent a developer from inheriting from the Configuration class? If this function is called from a child class then it will create a new base class instance.

I think you should change this to
```
$localConfigObj = new static($localConfigFile, $this->baseDir, $this->logger, $options);
```

jpwhite4:
Please list your unit test coverage report for these changes.

classes/ETL/Configuration/Configuration.php
smgallo: I waffled back and forth on this. If a developer is extending this class there are 5 methods that can be overridden to support additional functionality (see comments at the top of the file). This method exists solely to allow an extending class to create a local configuration object with the proper class **and** `$options`. In the one use case that we currently have, `EtlConfiguration`, there are additional options that must be passed so using `new static` alone wouldn't work. I'm open to other suggestions, though.

smgallo:
There are no specific unit tests for this code, I've been relying on integration testing with a known baseline.

```
~/src/xdmod/open_xdmod/modules/xdmod/tests$ ./runtests.sh 
PHPUnit 4.8.23 by Sebastian Bergmann and contributors.

Runtime:        PHP 5.5.9-1ubuntu4.21 with Xdebug 2.2.3
Configuration:  /home/smgallo/src/xdmod/open_xdmod/modules/xdmod/tests/phpunit.xml.dist

...............................................................  63 / 161 ( 39%)
..................................................Apr 20 14:14:36  [info] No 'permitted_modules' found for [usr] in [roles.json], continuing...
Apr 20 14:14:36  [info] No 'permitted_modules' found for [cd] in [roles.json], continuing...
Apr 20 14:14:36  [info] No 'permitted_modules' found for [pi] in [roles.json], continuing...
Apr 20 14:14:36  [info] No 'permitted_modules' found for [cs] in [roles.json], continuing...
Apr 20 14:14:36  [info] No 'permitted_modules' found for [mgr] in [roles.json], continuing...
............. 126 / 161 ( 78%)
...................................

Time: 2.63 seconds, Memory: 43.00Mb

OK (161 tests, 31614 assertions)
```

smgallo:
The previous several commits fix the following errors:
- Bug in RFC-6901 pointer code for including entire document and missing class constants
- Bug in SQL file processing where a dangling reference from a `foreach()` loop caused the last file to be processed several times
- Issue with PHP 5.3 not allowing overloading using child clases

The following component tests have been added
- `Configuration` class
    - General invalid JSON parse error
    - Basic parsing of a JSON file
    - Removing comments
    - RFC-6901 reference with fully qualified path
    - RFC-6901 reference with relative path
- `JsonReferenceTransformer` class
    - Invalid pointer due to unknown target key
    - Reference to a scalar value
    - Reference to an object value
    - Reference to an array element value
    - Reference to a the last element of an array
    - Reference to a key containing "/"
- `EtlConfiguration` class
    - Test parsing of an XDMoD ETLv2 configuration file and 2 local configuration files in a subdirectory defining additional pipelines

smgallo:
@jpwhite4 the issue with the php 5.3 test failure was this. This is fine under PHP 5.5 and the solution was to make the signatures the same and check the object type inside the function.

```
class A {
    function merge(A $obj) {
        ...
    }
}

class B extends A {
    function merge(B $obj) {
        ...
    }
}
```

smgallo:
Any other change requests on this or can I merge it in? @jpwhite4 @plessbd

classes/ETL/Configuration/EtlConfiguration.php
```diff
@@ -464,43 +287,236 @@ public function parse($force = false)
...
+
+        foreach ( $localConfigObj as $localSectionName => $localSectionActionOptions ) {
+
+            // It is recommended that action names are unique, but they don't need to
+            // be. If the same action name is found in a local config with the same
+            // pipeline name then use the $overwrite setting to handle it.
+
+            foreach ( $localSectionActionOptions as $localActionName => $localActionOptions ) {
+
+                // if ( array_key_exists($actionName, $this->actionOptions[$sectionName]) ) {
```
plessbd: remove commented out code

PR #108, opened 2017-04-25
fix forgot password form

plessbd:
Fix the forgot password form so that it looks correct

## Motivation and Context
Improve UX
Before:
before

After:
after

## Tests performed
Verified that the form looks correctly when viewed

## Types of changes

## Checklist:


plessbd:
Added

PR #112, opened 2017-04-26
Remove redundant tokenize code

jpwhite4:
## Description
This change removes the unnecessary duplicate tokenize code. It also
replaces the function with a much simpler one that has full unit test
coverage.

## Motivation and Context
General cleanup in preparation for deep linking support.

## Tests performed
Ran the javascript unit_tests to confirm that the location hashes are parsed correctly. Ran code coverage
tools to confirm the new code has 100% test coverage (although the tools cannot check that all possible states have been hit in the regex state machine).

### DIFF
--- a/html/unit_tests/spec/.eslintrc.json
+++ b/html/unit_tests/spec/.eslintrc.json
@@ -1,0 +1,9 @@
+{
+    "env": {
+        "mocha": true
+    },
+    "globals": {
+        "expect": false
+    }
+}
+
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -1357,94 +1357,33 @@ CCR.getParameter = function (name, source) {
             ? ""
             : decodeURIComponent(results[1].replace(/\+/g, " "));
 };
+/*
+ * Process the location hash string. The string should have the form:
+ *  PATH?PARAMS
+ *
+ * with an optional # prefix. PATH is a ':' delimited path leading to a component of the form
+ *   ROOT:TAB:SUBTAB
+ *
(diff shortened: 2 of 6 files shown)


html/unit_tests/spec/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "env": {
+        "mocha": true
+    }
```
plessbd: missing ,

plessbd:
did you do some basic before and after tests to show that the same input got the same output with old and new code?

plessbd:
Will this new tokenize code break any existing code since this one always has the keys in the result?

https://github.com/ubccr/xdmod/blob/feddfebaa82f35b14a637fd4dacf3a46fe7b019c/html/internal_dashboard/js/Dashboard/Viewport.js#L65

https://github.com/ubccr/xdmod/blob/feddfebaa82f35b14a637fd4dacf3a46fe7b019c/html/gui/js/modules/About.js#L130

https://github.com/ubccr/xdmod/blob/4b606aa25ffdd474efe9877a8eba2d63214cb77b/html/index.php#L506

etc

https://github.com/ubccr/xdmod/search?utf8=%E2%9C%93&q=tokenize&type=

jpwhite4:
To answer your questions: 1) Yes I did test , 2) No it won't break anything.

I note that the XDMoD.Dashboard.tokenize function was not modified in any way so this code change does not impact any code that calls it.

plessbd:
Screenshots or it didnt happen? :)

Looks Good To Me

PR #114, opened 2017-04-26
 Resolving Firefox MetricExplorer menu bug

ryanrath:
## Description
- Typing a '%' character in a form field contained within the
  MetricExplorer chart options menu would not have the desired effect. Instead,
  the whole menu would close and the '%' character would not be recorded in the
  field. The reason for this was because of an interaction between the FireFox
  browser and the Ext.KeyNav / Ext.MenuNav components. In FireFox, when the '%'
  character was entered into a form field contained within a menu, the
  Ext.KeyNav.relay function would fire, interpret the keyCode value of '37' to
  be the 'left' key, for which the Ext.MenuNav component has a handler. This
  handler would then merrily went about closing the menu ( because that's what
  you do when you press the left key? I'm guessing to support navigating a menu
  / toolbar via the arrow keys. ). Now, when Ext identifies the browser as
  Gecko ( which is what Firefox is identified as ), it just returns true so as
  to not interrupt any further events that need to be interpreted. If the
  browser is not Firefox then it executes the same code as in the original
  handler.

## Motivation and Context
Attempting to enter a '%' character while using Firefox into the name / title text boxes in the Metric Explorer would result in the menu that contains the text boxes to be closed and the character text box values to not be updated. 

## Tests performed
- Manually tested using the left / right keys in the text boxes inside this menu on NIX / OSX
- Manually tested using all non-alpha numeric characters I could think of available via keyboard. 
- Automated tests will be added shortly ( they have been added to the XSEDE module )

## Types of changes

## Checklist:

### DIFF
--- a/html/gui/js/modules/metric_explorer/MetricExplorer.js
+++ b/html/gui/js/modules/metric_explorer/MetricExplorer.js
@@ -5674,7 +5674,12 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
                     return true;
                 }
 
-                return h.call(this.scope || this, e, this.menu);
+                // Ensure that we only call the KeyNav handlers if this is a NavKey Press.
+                if (e.isNavKeyPress()) {
+                    return h.call(this.scope || this, e, this.menu);
+                }
+
+                return true;
             };
 
             /**


jpwhite4:
Does this problem affect all Ext.menu.Menu component instances or just the one in metric explorer. If it affects all of them then surely the change should be made as an Ext js override so all instances have the fix applied. If it only affects the one in metric explorer then what is different about the metric explorer instance?

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -5694,6 +5694,26 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
...
+            /**
+              * Firefox + Ext 3.4 === '%' being interpreted as the 'left' key.
+              * This resulted in the code after the 'if' being executed. Which,
+              * is somewhat odd behavior to say the least ( While viewing the
+              * site in firefox,  type '%' into a text box, have the whole menu
+              * close.). This override fixes that behavior
+             **/
+            this.chartOptionsMenu.keyNav.left = function (e, m) {
+                // Gecko === Firefox
+                if (Ext.isGecko === true) {
```
jpwhite4: It seems wrong to me to have to check the rendering engine here. Why don't you just change the doRelay call to only relay the event if the keypress object has isNavKeyPress() === true?

PR #117, opened 2017-04-27
Update JobViewer activate function in preparation for deep linking code.

jpwhite4:
## Description

The job viewer activate function needs to be updated to support the deep
linking feature. A review of the existing code found some dead code
paths and redundant code. This is removed in this pull request. 

Note that I had to add an eslint override to allow calls to the existing underscore prefix function. The eslint rule is poorly implemented because it prevents _calls_ to badly named existing functions, rather than preventing function declarations.

## Motivation and Context
prep work for deep linking support.

## Tests performed
Added unit tests for the replacement compareNodePath() function (this function has 100% unit tests coverage).

Manually tested that the job viewer navigation works identically in Firefox linux and Chrome linux.

### DIFF
--- a/html/gui/js/modules/job_viewer/JobViewer.js
+++ b/html/gui/js/modules/job_viewer/JobViewer.js
@@ -1,3 +1,29 @@
+/* eslint no-underscore-dangle: [
+   "error",
+   {
+      "allow" : [
+         "_getPath",
+         "_processViewRequest",
+         "_copy",
+         "_createHistoryEntry",
+         "_createHistoryToken",
+         "_createHistoryTokenFromArray",
+         "_find",
+         "_fromArray",
+         "_generateURL",
+         "_generateView",
+         "_getParams",
+         "_makeRequest",
+         "_panelActivation",
+         "_performLoad",
+         "_retrieveSearchInfo",
+         "_truncatePath",
+         "_updateHistoryFromPanel",
+         "_upsertSearch"
+      ]
+   }
+] */
+
 // TODO: Move this someplace else, just here for testing...
 if (!String.prototype.trim) {
     String.prototype.trim = function () {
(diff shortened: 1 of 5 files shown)


html/unit_tests/coverage.html
```diff
@@ -32,6 +32,17 @@
   <script type="text/javascript" src="../gui/lib/extjs/adapter/ext/ext-base.js"></script>
   <script type="text/javascript" src="../gui/lib/extjs/ext-all-debug-w-comments.js"></script>
   <script type="text/javascript" src="../gui/lib/extjs/src/debug.js"></script>
+  <script type="text/javascript" src="../gui/lib/CheckColumn.js"></script>
+  <script type="text/javascript" src="../gui/lib/extjs/examples/ux/ProgressBarPager.js"></script>
+  <script type="text/javascript" src="../gui/lib/extjs/examples/ux/treegrid/TreeGridSorter.js"></script>
```
plessbd: not that it should be part of this pull request, but I think if we are using something it should not be sitting in an examples directory.

html/gui/js/modules/job_viewer/JobViewer.js
```diff
@@ -1,3 +1,29 @@
+/* eslint no-underscore-dangle: [
```
plessbd: I assume these are specific to only the job viewer and why they are in here rather than in a "global" eslintrc

html/gui/js/modules/job_viewer/JobViewer.js
```diff
@@ -876,61 +902,39 @@ XDMoD.Module.JobViewer = Ext.extend(XDMoD.PortalModule, {
...
-            var job = getParameter(jobKey, hash);
-            var view = getParameter(viewKey, hash);
-            var timeSeries = getParameter(timeSeriesKey, hash);
+            if (!params.realm) {
+                return;
+            }
 
-            var metricExplorerData = getParameter('job', hash);
+            var path = this._getPath(token.raw);
+            var isSelected = this.compareNodePath(this.currentNode, path);
```
plessbd: going back and forth on if we should put this in a variable that is only used once or if we should just put a comment saying what it does.

html/gui/js/modules/job_viewer/JobViewer.js
```diff
@@ -876,61 +902,39 @@ XDMoD.Module.JobViewer = Ext.extend(XDMoD.PortalModule, {
...
-                this.fireEvent('process_realm_node', path, isSelected);
-            } else if (hasRealm && hasSearch && !hasJob && !hasView) {
-                this.fireEvent('process_search_node', path, isSelected);
-            } else if (hasRealm && hasSearch && hasJob && !hasView) {
-                this.fireEvent('process_job_node', path, isSelected, this._processViewRequest);
-            } else if (hasRealm && hasSearch && hasJob && hasView) {
-                this.fireEvent('process_view_node', path, isSelected);
-            } else if (hasRealm && hasSearch && hasJob && hasView && hasTimeSeries) {
-                this.fireEvent('process_view_node', path, isSelected);
+            if (params.realm && params.recordid && params.jobid) {
```
plessbd: the check for params.realm is not needed here as you cant get to this code without it existing

html/unit_tests/coverage.html
smgallo: I blame that on ExtJS 3.4 since that's where they put their code in the zipfile.

html/gui/js/modules/job_viewer/JobViewer.js
jpwhite4: this is to ensure that this code is exactly the same logic as the previous version. I'll likely get rid of this unneeded check in the next pull request.

html/gui/js/modules/job_viewer/JobViewer.js
jpwhite4: Yes. Also the default eslint rules are such that you cannot even call a function with a 'bad' name, which is a problem with trying to interface to old non-compliant code.

html/unit_tests/coverage.html
jpwhite4: Repackaging extjs is beyond the scope of this project.

PR #142, opened 2017-05-15
Reset master branch to xdmod6.6

jtpalmer:
## Motivation and Context

Update open.xdmod.org


tyearke:
You can force push the `master` branches to point to the same commits as the `v6.6.0` tags if you prefer (it's what I usually just do). You'll need to temporarily disable the protections on the `master` branches, but that's fine.

jtpalmer:
Ok, I'll do that.

PR #143, opened 2017-05-17
Use Git to manage XDMoD test artifacts

tyearke:
## Description
This pull request moves responsibility for retrieval of `xdmod-test-artifacts` from Composer to the scripts that use it.

## Motivation and Context
Copying/paraphrasing from an email exchange between @smgallo, @jpwhite4, and I:

Composer's design is built around pulling in runtime and development software packages in a consistent, well-considered manner, and the lock file's use of hashes for both individual dependencies and the contents of the lock file is meant to help guard against changes that haven't been tested. For example, if two pull requests both change the lock file and one gets merged, the other will be in conflict with the target branch due to the lock file's content hash being changed by both pull requests. As part of resolving the conflict, the outstanding pull request's lock file will need to be regenerated to include the changes from the other, which in so doing will (partially) validate the combined changes.

Having developers run commands that modify the lock file as part of the testing procedure leads to more work and more opportunities for accidentally pushing bad changes to the lock file. What this pull request does instead is have the scripts that use the test artifacts repo pull in the latest version of the repo with plain Git at run time. This means no updates to run manually and no files in the code repos to tweak and accidentally break when adding to the artifacts.

## Tests Performed
Ran a test run of Travis on my fork and ran the unit tests manually on my local system.

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/tests/runtests.sh
+++ b/open_xdmod/modules/xdmod/tests/runtests.sh
@@ -14,5 +14,7 @@ if [ ! -x "$phpunit" ]; then
     exit 127
 fi
 
+./artifacts/update-artifacts.sh
+
 $phpunit ${PHPUNITARGS} .
 exit $?
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ php:
 env:
     global:
         - NODE_VERSION=6
+        - XDMOD_TEST_ARTIFACTS_MIRROR="$HOME/xdmod-test-artifacts.git"
     matrix:
         - TEST_SUITE=syntax
         - TEST_SUITE=style
@@ -25,6 +26,7 @@ cache:
         - $HOME/.npm
         - $HOME/.composer/cache
         - /tmp/pear/cache
+        - $XDMOD_TEST_ARTIFACTS_MIRROR
 
 # Delegate the installation step to the custom Travis installation script
 install: ./.travis.install.sh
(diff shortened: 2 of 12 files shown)


open_xdmod/modules/xdmod/tests/runtests.sh
```diff
@@ -14,5 +14,12 @@ if [ ! -x "$phpunit" ]; then
     exit 127
 fi
 
+artifacts_dir="./xdmod-test-artifacts"
```
jpwhite4: Can we move this out to a separate setup script and then call it here. Also move the test artifacts directory somewhere else since it could  (and will) be used by the integration_tests and regression_tests.

tyearke:
I've pushed a new commit that moves the artifact repo setup code from `runtests.sh` and the Travis scripts into a single new script. This script and the clone it creates/updates are now located in `tests/artifacts` instead of `tests`. When evaluating where to put them, I noticed that the locations of Open XDMoD test suites don't match the locations of those suites in the App Kernels and SUPReMM modules.

Project|Unit Tests|Integration Tests
---|---|---
Open XDMoD\*|`tests`|`integration_tests`
SUPReMM|`tests/unit_tests`|`tests/integration_tests`
App Kernels||`tests/integration_tests`

\* = prepend `open_xdmod/modules/xdmod/`

My suggestion would be to (as part of a separate pull request) modify Open XDMoD's layout to match SUPReMM's layout. If that will be done, then `tests/artifacts` seems like a sensible location for the artifact-related assets to go in every repo.

I'll rebase the code to resolve conflicts if the changes I've made so far look good.

PR #146, opened 2017-05-23
Acl xduser

ryanrath:
## Types of changes

## Checklist:


classes/XDUser.php
```diff
@@ -1,466 +1,486 @@
...
+     * function will return null.
+     * cl
+     * @return null|XDUser
+     */
+    public static function getPublicUser()
+    {
+        if (null !== self::$_publicUser) {
+            return self::$_publicUser;
+        } else {
+            self::$_publicUser = self::getUserByUserName('Public User');
```
jpwhite4: Where is the code that automatically creates the 'Public User' row in the database? I can't find it anywhere.

jpwhite4: Also as a coding standard comment: 

I prefer to not repeat identical lines in the code unless it is unavoidable. I suggest the following, which I think is easier to see what is happening:
```
if (null === self::$_publicUser) {
    self::$_publicUser = self::getUserByUserName('Public User');
}
return self::$_publicUser;
```

shippable.yml
```diff
@@ -18,6 +18,9 @@ build:
         - composer install --no-progress
         - cp ~/assets/secrets open_xdmod/modules/xdmod/integration_tests/.secrets
         - ./open_xdmod/modules/xdmod/integration_tests/runtests.sh --log-junit `pwd`/shippable/testresults/results.xml
+        - cp ./configuration/portal_settings.ini ./configuration/portal_settings.ini.old
+        - cp -f /etc/xdmod/portal_settings.ini ./configuration/portal_settings.ini
+        - ./open_xdmod/modules/xdmod/component_tests/runtests.sh --log-junit `pwd`/shippable/testresults/componentresults.xml
+
```
jpwhite4: You should restore the portal_settings.ini.old to its original location after running the tests (as we discussed before).

ryanrath: Thanks! I did totally miss that. Latest commit mv's the portal_settings.ini.old back to portal_settings.ini.

PR #153, opened 2017-06-07
Add gantt chart view of job peers

jpwhite4:
## Description
Change the Peers node in the search history so that it opens a new tab with an interactive graphical display of the job's peer jobs (i.e. the jobs that ran on the same shared hardware). The individual job peers no longer show as leaf nodes because this display was not useful and did not scale when there were lots of job peers. Jobs with many peers are now handled using the paging toolbar in the graphical display.

These code changes:
- Add new REST endpoint to retrieve the peer data
- Change the peer tree node type and metadata
- Add new Gantt chart class
- Add new base class for job viewer chart tabs.

Note that the new ChartTab base class is intended to be used as the base class for all job viewer timeseries plots. This should significantly reduce the amount of cut-and-paste boilerplate code. A future pull request is planned to switch all charts over to this code.

The job peers click uses the existing mechanism that supports the click through from the metric explorer.

peers

## Motivation and Context
Improve ability to analyse shared jobs.

## Tests performed
Installed on metrics-dev manually tested. Tests included:
- confirm that the chart clicks work
- confirm that paging works
- run through multiple combinations of opening and closing job tabs confirm that the UI behaves as expected. Confirm that the tree nodes get updated as expected. 

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- My code follows the code style of this project as found in the **CONTRIBUTING** document.
- All new and existing tests passed.

### DIFF
--- a/classes/NewRest/Controllers/WarehouseControllerProvider.php
+++ b/classes/NewRest/Controllers/WarehouseControllerProvider.php
@@ -116,13 +116,6 @@ class WarehouseControllerProvider extends BaseControllerProvider
                 "type" => "utf8-text",
                 "leaf" => true
             ),
-        \DataWarehouse\Query\RawQueryTypes::PEERS =>
-            array(
-                "infoid" => \DataWarehouse\Query\RawQueryTypes::PEERS,
-                "dtype" => "infoid",
-                "text" => "Peers",
-                "leaf" => false
-            ),
         \DataWarehouse\Query\RawQueryTypes::EXECUTABLE =>
             array(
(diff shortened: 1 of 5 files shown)


classes/NewRest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1352,6 +1360,72 @@ public function _processJobSearchByAction(Request $request, Application $app, XD
         return $results;
     }
 
+    protected function getJobPeers(Application $app, XDUser $user, $realm, $jobId, $start, $limit)
```
ryanrath: Documentation for the function wouldn't suck

PR #157, opened 2017-06-12
Allow DataEndpoint singleton to auto-discover available data endpoints

smgallo:
Allow DataEndpoint singleton to auto-discover available data endpoints

## Description

Rather than require a developer to modify `ETL\DataEndpoint.php` and explicitly list each available data endpoint and it's configuration code, we now allow the `DataEndpoint` singleton to automatically discover what data endpoints are available. As per PSR-4 the contiguous sub-namespace names after the "namespace prefix" (`ETL` in our case) correspond to a subdirectory within a "base directory", in which the namespace separators represent directory separators. From this, we can assume the `ETL\DataEndpoint` sub-namespace is in the `DataEndpoint` subdirectory relative to where the `DataEndpoint` singleton class is defined. Using this knowledge we use reflection to recursively examine each file in the `DataEndpoint` subdirectory and any class implementing `iDataEndpoint` and also defining an `ENDPOINT_NAME` constant will be a valid endpoint definition. The  `ENDPOINT_NAME` constant defines the endpoint name to be used in configuration files. Conflicts in endpoint names are logged.

The `etl_overseer` script has been modified to list the available endpoints:
```
$ php etl_overseer.php -c ../../../etc/etl/etl.json -l endpoint-types                       
Name    Class
directoryscanner        ETL\DataEndpoint\DirectoryScanner
file    ETL\DataEndpoint\File
jsonfile        ETL\DataEndpoint\JsonFile
mysql   ETL\DataEndpoint\Mysql
oracle  ETL\DataEndpoint\Oracle
postgres        ETL\DataEndpoint\Postgres
rest    ETL\DataEndpoint\Rest
```
## Motivation and Context

Make it easier to define data endpoints.

## Tests performed

Ran existing tests, several of which use the File and Json endpoints.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/DataEndpoint.php
+++ b/classes/ETL/DataEndpoint.php
@@ -18,31 +18,31 @@ use Log;
 class DataEndpoint
 {
     /**
-     * Namesapce, relative to the current namespace, where data endpoint classes are
+     * Namespace, relative to the current namespace, where data endpoint classes are
      * defined. This is used to automatically search for defined endpoints.
      *
-     * @var string
+     * @const string
      */
 
-    private static $dataEndpointRelativeNs = 'DataEndpoint';
+    const DATA_ENDPOINT_RELATIVE_NS = 'DataEndpoint';
 
     /**
      * Fully namespaced interface that all data endpoints must implement
      *
-     * @var string
+     * @const string
      */
 
-    private static $dataEndpointRequiredInterface = 'ETL\\DataEndpoint\\iDataEndpoint';
+    const DATA_ENDPOINT_REQUIRED_INTERFACE = 'ETL\\DataEndpoint\\iDataEndpoint';
 
     /**
      * The name of the constant expected to be defined in all data endpoint classes. This
      * is used to identify the name that will be used to refer to the endpoint in
      * configuration files.
      *
-     * @var string
+     * @const string
      */
 
-    private static $endpointNameConstant = 'ENDPOINT_NAME';
+    const ENDPOINT_NAME_CONSTANT = 'ENDPOINT_NAME';
 
     /**
      * Associative array where the keys are data endpoint names and the values are fully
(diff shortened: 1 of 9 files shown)


classes/ETL/DataEndpoint.php
```diff
@@ -61,10 +63,157 @@ private function __construct()
...
+
+    /** -----------------------------------------------------------------------------------------
      * Factory pattern to instantiate a DataEndpoint based on the options.
      *
-     * @param $options A DataEndpointOptions object containing options parsed from the ETL config.
+     * @param DataEndpointOptions $options A DataEndpointOptions object containing options
+     *   parsed from the ETL config.
+     * @param Log $logger A PEAR Log object or null to use the null logger.
      *
      * @return A data endpoint object implementing the iDataEndpoint interface.
```
ryanrath: missing a return type

classes/ETL/DataEndpoint/DirectoryScanner.php
```diff
@@ -376,209 +382,210 @@ public function connect()
...
-                $this->logger->debug(
-                    sprintf("Set max recursion depth: %d", $this->maxRecursionDepth)
-                );
-                $flattenedIterator->setMaxDepth($this->maxRecursionDepth);
-            }
+        try {
+            $dotDirFilterIterator = new \CallbackFilterIterator(
+                $iterator,
+                function ($current, $key, $iterator) {
+                    if ( $iterator->isDot() ) {
```
ryanrath: This can be simplified to `return ! $iterator->isDot();`

classes/ETL/DataEndpoint/DirectoryScanner.php
```diff
@@ -376,209 +382,210 @@ public function connect()
...
+
+        try {
+            $flattenedIterator = new \RecursiveIteratorIterator($iterator);
+        }  catch ( Exception $e ) {
+            $this->logAndThrowException(
+                sprintf("Error creating RecursiveIteratorIterator: %s", $e->getMessage())
+            );
+        }
+
+        if ( null !== $this->maxRecursionDepth ) {
```
ryanrath: May want to move this line to 464 into the try above as all of these operations depend on $flattenedIterator existing. It would also then conform to the pattern seen elsewhere in the code of: try block that creates new iterator, maybe does a few things and ultimately sets iterator = new iterator

classes/ETL/DataEndpoint.php
```diff
@@ -12,47 +12,49 @@
...
 class DataEndpoint
 {
+    /**
+     * Namesapce, relative to the current namespace, where data endpoint classes are
+     * defined. This is used to automatically search for defined endpoints.
+     *
+     * @var string
+     */
+
+    private static $dataEndpointRelativeNs = 'DataEndpoint';
```
smgallo: Actually, no they don't change programmatically so it does make sense for them to be constants. I'll make that change.

classes/ETL/DataEndpoint.php
```diff
@@ -61,10 +63,157 @@ private function __construct()
...
+        }
+
+        // As per PSR-4 (http://www.php-fig.org/psr/psr-4/) the contiguous sub-namespace
+        // names after the "namespace prefix" correspond to a subdirectory within a "base
+        // directory", in which the namespace separators represent directory separators.
+        // This means that we can assume subdirectories under the directory where this
+        // file resides represent sub-namespaces.
+
+        // The endpoint directory is relative to the directory where this file is found
+        $endpointDir =  dirname(__FILE__) . '/' . strtr(self::$dataEndpointRelativeNs, '\\', '/');
```
smgallo: DIRECTORY_SEPARATOR is a more portable way to specify it (not that it matters on the platforms that we support). I'll make that change as well.

classes/ETL/DataEndpoint/DirectoryScanner.php
smgallo: Good points.

PR #158, opened 2017-06-13
Replaced Zend with PHPMailer

jpearce185:
## Description
Removed Zend framework from xdmod and replaced it with PHPMailer.  All functions and methods relating to Zend were reconfigured to the corresponding PHPMailer methods/functions.

## Motivation and Context
To replace Zend with a more robust framework, i.e. PHPMailer.

## Tests performed
To start I continuously signed up to xdmod using a test email I created (test2) and I had a user created (test1) for me which had Manager, User, Principal Investigator, and Developer privileges.  Using this test1 account, I started out by completing the "Send Message" and "Request Feature" forms under the "Contact Us" tab (updating code as needed), and repeated this process until both the user and xdmod received the proper emails (I knew xdmod would receive these emails because I set my test1 email to the debug email, so all emails that xdmod would normally receive, my test1 received).  Next, to test creating a user, I went into the dashboard and created a test user with only User privileges using the test2 email account.  I then sent a password reset email to this account to test the password reset feature (from an admin's account) which was successful.  I then logged in as this new user and generated a report via the report manager.  I sent an email with an attachment to the user successfully.  To test again the "Send Message" and "Request Feature" forms (from a different account than previously) I completed them and it was successful.  To then test the user password reset functionality, I logged out completely, proceeded to the "Sign In" link and next to "Trouble logging in?" clicked on "Click Here" which in turn sent a password reset email to the user's email, which is what should happen.  To then test the SAML Authentication (logging in via federated credentials) I went back into the "Sign In" link and clicked on the federation login banner.
I then had another colleague sign up with another email to ensure it worked and it did.

## Checklist:


classes/MailWrapper.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class MailWrapper
+{
+
+    public static function init()
+    {
+	$mail = new PHPMailer();
+	$mail->isSendMail();
+	$mail->Sender = 'fxdmod-bounces@ccr.buffalo.edu';
```
jpwhite4: This address does not appear to be correct you have an extra f prepended.

composer.lock
```diff
@@ -2688,6 +2750,28 @@
             "description": "Symfony Yaml Component",
             "homepage": "https://symfony.com",
             "time": "2015-07-26T08:59:42+00:00"
+        },
```
jpwhite4: This should not have been added to the lock file. Please regenerate the lock file.

html/internal_dashboard/controllers/mailer.php
```diff
@@ -88,23 +88,23 @@
...
         $contact = \xd_utilities\getConfiguration('general', 'contact_page_recipient');
-        $mail->addTo($contact, 'Undisclosed Recipients');
+        $mail->addAddress($contact, 'Undisclosed Recipients');
         $mail->setFrom($contact, $title);
 
         $bcc_emails = explode(',', $target_addresses);
 
         foreach ($bcc_emails as $b) {
-            $mail->addBcc($b);
+            $mail->AddBCC($b);
```
jpwhite4: s/AddBCC/addBCC/

classes/MailWrapper.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class MailWrapper
+{
+
+    public static function init()
+    {
+	$mail = new PHPMailer();
```
jpwhite4: Please change the indentation to use spaces rather than tabs.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -136,23 +136,23 @@ public function getLoginLink()
...
         ? xd_utilities\getConfiguration('general', 'debug_recipient')
         : xd_utilities\getConfiguration('general', 'contact_page_recipient');
-        $mail->addTo($recipient);
+        $mail->addAddress($recipient);
         if ($error) {
-            $mail->setSubject("[xdmod] Error Creating federated user");
+            $mail->Subject = "[xdmod] Error Creating federated user";
         } else {
-            $mail->setSubject("[xdmod] New ". ($linked ? "linked": "unlinked") ." federated user created");
+            $mail->Subject = "[xdmod] New ". ($linked ? "linked": "unlinked") ." federated user created";
```
smgallo: Please add spaces before/after concatenation operators.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -164,12 +164,12 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
         print_r($samlAttributes, true);
 
-        $mail->setBodyText($body);
+        $mail->Body = $body;
         try {
             $mail->send();
             return true;
         } catch (Exception $e) {
-            return $e->getMessage();
+            return $e->getMessage() . "\n" . $mail->ErrorInfo;
```
smgallo: You should prefix the error reporting by the module that it occurred in. For example: "Error sending mail to '%s': %s", $email, $errorMessage.

smgallo: Rather than return `true` in one case and a string in another case, you should re-throw the exception and catch it in the calling routine. This makes error handling easier.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -164,12 +164,12 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
         "Additional SAML Attributes ----------------------------------\n\n" .
```
smgallo: Only add SAML attributes if `$samlAttributes` is not empty.

classes/MailWrapper.php
smgallo: Fix indentation.

classes/MailWrapper.php
smgallo: This should be a configurable email address. And it appears to have an extra "f".

crontabs/xdmod
```diff
@@ -0,0 +1,35 @@
+MAILTO=ccr-xdmod-cron@buffalo.edu
```
jpwhite4: This file should not be added. Please remove it and rebase your commits.

crontabs/xdtas
```diff
@@ -0,0 +1,28 @@
+# Edit this file to introduce tasks to be run by cron.
```
jpwhite4: This file should not be added.

classes/MailWrapper.php
```diff
@@ -1,15 +0,0 @@
-<?php
```
smgallo: I thought we were going to keep `MailWrapper.php` but change `init()` to `initPhpMailer($sender)`?

crontabs/xdtas
smgallo: Why is this file here?

crontabs/xdmod
smgallo: Why is this file here?

html/controllers/mailer/contact.php
```diff
@@ -72,7 +72,9 @@
   ? xd_utilities\getConfiguration('general', 'debug_recipient')
   : xd_utilities\getConfiguration('general', 'contact_page_recipient');
 
-$mail = MailWrapper::init();
+$mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

html/controllers/mailer/contact.php
```diff
@@ -116,7 +118,9 @@
 
 // =====================================================
 
-$mail = MailWrapper::init();
+$mail = new PHPMailer();
+$mail->isSendMail();
+$mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
```
smgallo: Rather than call `\xd_utilities\getConfiguration('mailer', 'sender_email')` multiple times, consider calling it once and setting a variable.

html/controllers/mailer/sign_up.php
```diff
@@ -87,8 +87,9 @@
 );
 
 // Create email.
-$mail = MailWrapper::init();
-
+$mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

html/controllers/user_admin/create_user.php
```diff
@@ -109,8 +109,9 @@
 
     // -------------------
 
-    $mail = MailWrapper::init();
-
+    $mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

html/controllers/user_admin/pass_reset.php
```diff
@@ -34,7 +34,9 @@
 
 // -------------------
 
-$mail = MailWrapper::init();
+$mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

html/controllers/user_auth/pass_reset.php
```diff
@@ -39,7 +39,9 @@
 
 // -------------------
 
-$mail = MailWrapper::init();
+$mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

html/internal_dashboard/controllers/mailer.php
```diff
@@ -88,8 +88,9 @@
 
         $response['success'] = true;
 
-        $mail = MailWrapper::init();
-
+        $mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

classes/XDReportManager.php
```diff
@@ -2045,8 +2045,9 @@ public function mailReport(
         $frequency = '',
         $additional_config = array()
     ) {
-        $mail = MailWrapper::init();
-
+        $mail = new PHPMailer();
```
smgallo: See my comment above about `MailWrapper.php`

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -159,17 +165,23 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
         "Person Details ----------------------------------\n\n" .
         "\nName:                     " . $user->getFormalName(true) .
         "\nUserame:                  " . $user->getUsername() .
-        "\nE-Mail:                   " . $userEmail .
-        "\n\n" .
-        "Additional SAML Attributes ----------------------------------\n\n" .
-        print_r($samlAttributes, true);
+        "\nE-Mail:                   " . $userEmail ;
+
+        if(!empty($samlAttributes)){
```
smgallo: Is `$samlAttributes` an array? If so, using `if( 0 == count($samlAttributes) )` would be a better choice.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -159,17 +165,23 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+                print_r($samlAttributes, true);
+        }
 
         $mail->Body = $body;
         try {
-            $mail->send();
-            return true;
+            if(!$mail->send()){
+                error_log("Mail failed to send to: " . $userEmail . "\n" . $mail->ErrorInfo);
+            }
```
smgallo: This logs the error, but still returns `true`. Is that your intention?

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -159,17 +165,23 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+            $body = $body . "\n\n" .
+                "Additional SAML Attributes ----------------------------------\n\n" .
+                print_r($samlAttributes, true);
+        }
 
         $mail->Body = $body;
         try {
-            $mail->send();
-            return true;
+            if(!$mail->send()){
```
smgallo: Spaces are your friend - they make code far more readable, especially when looking at it in a non-colorized context. This would read better as `if( ! $mail->send() ) {`

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -159,17 +165,23 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+                error_log("Mail failed to send to: " . $userEmail . "\n" . $mail->ErrorInfo);
+            }
         } catch (Exception $e) {
-            return $e->getMessage() . "\n" . $mail->ErrorInfo;
+            error_log("Mail failed to send." . $mail->ErrorInfo);
```
smgallo: Why not log the same message being thrown by the exception? It has far more information in it.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -96,11 +96,15 @@ public function getXdmodAccount()
...
                 $newUser->saveUser();
+            } catch(Exception $e) {
+                error_log('User creation failed: ' . $e->getMessage());
+                throw new Exception('User creation failed: ' . $e->getMessage());
+            }
+            try {
                 self::notifyAdminOfNewUser($newUser, $samlAttrs, ($personId != -2));
                 return $newUser;
             } catch (Exception $e) {
-                self::notifyAdminOfNewUser($newUser, $samlAttrs, ($personId != -2), true);
```
smgallo: If you're not going to do anything with the exception that is thrown, you should return `false` instead and check the return value.

classes/Authentication/SAML/XDSamlAuthentication.php
jpwhite4: Looking at the source code for phpmailer it can be configured whether or not to throw exceptions and it defaults to false. This means that the way we are currenlty using this the mailer class will not throw. We should either:

1. Enable exceptions in phpmailer and change the code to catch them (and don't bother checking the return code.
1. Leave the phpmailer config at default and don't bother trying to catch exceptions that can never be thrown.

Also can we try to find a  way of testing these error cases? Perhaps deliberately switch off postfix on a test system.

smgallo: If it supports exceptions I say we configure it that way, since that is consistent with what we do normally.

smgallo:
The most recent changes look good, but you will need to add `MailWrapper` into a namespace in order for it to pass the automated tests/style checks.

classes/MailWrapper.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use CCR;
```
smgallo: Please review the [PHP namespace](http://php.net/manual/en/language.namespaces.php) documentation  on how to fix this.

html/controllers/mailer/contact.php
```diff
@@ -86,14 +86,14 @@
...
 
-$mail->setSubject($subject);
-$mail->addTo($recipient);
+$mail->Subject = $subject;
+$mail->addAddress($recipient);
 
 //$mail->setFrom($mailer_sender, 'XDMoD');
 
 //Original sender's e-mail must be in the 'From' field for the XDMoD Request Tracker to function
 $mail->setFrom($_POST['email']);
```
jpwhite4: This is a  change to the behaviour of the code. the old version of the code would set the from address and the envelope address to the contents of the $_POST['email'] and the new code sets the envelope to the default sender and the from address to $_POST['email']. 

Is this change intended? If so then please document the motivation (i.e. what bug was fixed). If not then please revert to the original behaviour.

html/controllers/mailer/sign_up.php
```diff
@@ -87,22 +87,23 @@
...
 $recipient
     = (xd_utilities\getConfiguration('general', 'debug_mode') == 'on')
     ? xd_utilities\getConfiguration('general', 'debug_recipient')
     : xd_utilities\getConfiguration('general', 'contact_page_recipient');
-$mail->addTo($recipient);
+$mail->addAddress($recipient);
 
 // Original sender's e-mail must be in the "From" field for the XDMoD
 // Request Tracker to function
 $mail->setFrom($_POST['email']);
```
jpwhite4: Same question as above:

This is a change to the behaviour of the code. the old version of the code would set the from address and the envelope address to the contents of the $_POST['email'] and the new code sets the envelope to the default sender and the from address to $_POST['email'].

Is this change intended? If so then please document the motivation (i.e. what bug was fixed). If not then please revert to the original behaviour.

html/controllers/user_admin/create_user.php
```diff
@@ -105,22 +105,21 @@
 
     $page_title = xd_utilities\getConfiguration('general', 'title');
     $site_address = xd_utilities\getConfigurationUrlBase('general', 'site_address');
-    $mailer_sender = xd_utilities\getConfiguration('mailer', 'sender_email');
+    $sender = xd_utilities\getConfiguration('mailer', 'sender_email');
```
jpwhite4: In other parts of this pull request you call strtolower() on the sender email string. It is not done here. Why do you call tolower at all? Why is it different here?

classes/MailWrapper.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+class MailWrapper
+{
+
+    public static function initPHPMailer($sender)
+    {
+        $mail = new PHPMailer(true);
+        $mail->isSendMail();
+        $mail->Sender = $sender;
```
plessbd: Do we catch if this is not a valid email address?

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -140,9 +141,9 @@ public function getLoginLink()
     }
     private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error = false)
     {
-        $mail = new PHPMailer();
-        $mail->isSendMail();
-        $mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
+        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
+
+        $mail = MailWrapper::initPHPMailer($sender);
```
plessbd: is $sender used anywhere else, if not why bother putting it into a variable

classes/XDReportManager.php
```diff
@@ -2045,12 +2045,11 @@ public function mailReport(
...
-        $mail = new PHPMailer();
-        $mail->isSendMail();
-        $mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
-        $mailer_sender = xd_utilities\getConfiguration('mailer', 'sender_email');
+        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
 
-        $mail->setFrom($mailer_sender, 'XDMoD');
+        $mail = MailWrapper::initPHPMailer($sender);
+
+        $mail->setFrom($sender, 'XDMoD');
```
plessbd: this uses setFrom instead of sender like the other file, should we make these the same

It looks like https://github.com/PHPMailer/PHPMailer/blob/master/class.phpmailer.php#L1011 setFrom internally also sets the sender

html/controllers/mailer/contact.php
```diff
@@ -65,16 +65,14 @@
...
 
 $recipient
   = (xd_utilities\getConfiguration('general', 'debug_mode') == 'on')
   ? xd_utilities\getConfiguration('general', 'debug_recipient')
   : xd_utilities\getConfiguration('general', 'contact_page_recipient');
 
-$mail = new PHPMailer();
-$mail->isSendMail();
-$mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
+$mail = MailWrapper::initPHPMailer($sender);
```
plessbd: same as others, do we need to store $sender

html/controllers/user_auth/pass_reset.php
```diff
@@ -39,10 +38,10 @@
 
 // -------------------
 
-$mail = new PHPMailer();
-$mail->isSendMail();
-$mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
-$mail->setFrom($mailer_sender, 'XDMoD');
+$sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
```
plessbd: see previous comments

html/internal_dashboard/controllers/mailer.php
```diff
@@ -88,9 +88,9 @@
 
         $response['success'] = true;
 
-        $mail = new PHPMailer();
-        $mail->isSendMail();
-        $mail->Sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
+        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
+
+        $mail = MailWrapper::initPHPMailer($sender);
```
plessbd: see previous comments

plessbd:
Do we maybe want to have a helper function of something like
```php
MailWrapper::sendSystemEmail($subject, $body)
```
It seems most of our emails are all sent as the "XDMoD" User and the only difference is the subject and body. Might clean up the code a bit more.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -136,40 +141,41 @@ public function getLoginLink()
...
 
         $recipient
         = (xd_utilities\getConfiguration('general', 'debug_mode') == 'on')
         ? xd_utilities\getConfiguration('general', 'debug_recipient')
         : xd_utilities\getConfiguration('general', 'contact_page_recipient');
-        $mail->addTo($recipient);
+        $mail->addAddress($recipient);
         if ($error) {
-            $mail->setSubject("[xdmod] Error Creating federated user");
+            $mail->Subject = "[xdmod] Error Creating federated user";
```
plessbd: In some places we use the `xd_utilities\getConfiguration('general', 'title');` should we be using that here instead of [xdmod]

classes/XDReportManager.php
```diff
@@ -2045,19 +2045,19 @@ public function mailReport(
...
         $additional_config = array()
     ) {
-        $mail = ZendMailWrapper::init();
+        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
 
-        $mailer_sender = xd_utilities\getConfiguration('mailer', 'sender_email');
+        $mail = MailWrapper::initPHPMailer($sender);
 
-        $mail->setFrom($mailer_sender, 'XDMoD');
+        $mail->setFrom($sender, 'XDMoD');
```
plessbd: In some places we use the `xd_utilities\getConfiguration('general', 'title');` should we be using that here instead of [xdmod]

html/controllers/mailer/contact.php
```diff
@@ -102,24 +102,24 @@
...
 }
 
 // =====================================================
 
-$mail = ZendMailWrapper::init();
-$mail->setFrom($mailer_sender, 'XDMoD');
-$mail->setSubject("Thank you for your $message_type.");
-$mail->addTo($_POST['email']);
+$mail = MailWrapper::initPHPMailer($sender);
+$mail->setFrom($sender, 'XDMoD');
```
plessbd: In some places we use the `xd_utilities\getConfiguration('general', 'title');` should we be using that here instead of [xdmod]

plessbd:
in the places where we use `xd_utilities\getConfiguration('general', 'title');` we should be consistent on if we put it in `[]` or not

classes/Authentication/SAML/XDSamlAuthentication.php
jpearce185: Putting $sender in a variable and passing it into the MailWrapper class allows $sender to be set to a configurable email address, rather than be hardcoded

plessbd: Understood, but why assign it to a variable outside of the wrapper instead of just sending it in like this:

```php
$mail = MailWrapper::initPHPMailer(strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email')));
```

html/controllers/mailer/contact.php
jpearce185: yes because storing that value allows for a configurable email address

classes/XDReportManager.php
jpearce185: setFrom only sets sender if sender has no value assigned to it.  An email is sent from the setFrom address on behalf of the sender address. No, they shouldn't be the same.

classes/Authentication/SAML/XDSamlAuthentication.php
jpearce185: No, because that is unrelated to the creation of federated users.

plessbd: I agree, it is unrelated to that, it is only related to how emails are sent out.  However if you mean it is unrelated to the change from ZendMail to PHPMailer, I can agree to that.

html/controllers/mailer/contact.php
jpearce185: No, because we want the message to be from the xdmod team (which it was originally), rather than from "open xdmod" which is somewhat vague as it refers to everything

html/controllers/mailer/contact.php
jpearce185: The from address and envelope address are still the same ($_POST['email']), however, what changed was how the Sender value was set.  Sender, in the old code, was hard coded.  Now, the Sender value has been stored in a variable ($sender) that gets the value stored in the config file, thus allowing Sender to now be a configurable address.  When Sender and setFrom have different values, Sender appends itself to setFrom's value (i.e. some.email@yahoo.com via buffalo.edu, for example).  If Sender has no value, setFrom sets Sender to its own value (the same).

html/controllers/mailer/sign_up.php
jpearce185: See above.

classes/MailWrapper.php
```diff
@@ -0,0 +1,16 @@
...
+namespace CCR;
+
+class MailWrapper
+{
+
+    public static function initPHPMailer($sender)
+    {
+        $mail = new PHPMailer(true);
+        $mail->isSendMail();
+        $mail->Sender = $sender;
```
plessbd: since we always want the sender to be `\xd_utilities\getConfiguration('mailer', 'sender_email')`
but we want the from to be configurable from the caller maybe something like this would be more appropriate
```php
public static function initPHPMailer($fromEmail = null, $fromName = '');
{
  $mail = new PHPMailer(true);
  $mail->isSendMail();
  $address =  \xd_utilities\getConfiguration('mailer', 'sender_email');
  $name = \xd_utilities\getConfiguration('general', 'title');
  $mail->Sender = $address;
  
  if($fromEmail !== null) {
      $address = $fromEmail;
      $name = $fromName;
  }
  try {
    $mail->setFrom($address, $name);
  }
  catch(phpmailerException $e){
    /*
      Handle invalid sender address here
    */
  }
  
  return $mail;
}  
```

this should reduce code duplication and set up the email to send properly.

html/controllers/mailer/contact.php
plessbd: Storing it doesnt allow for the passing it in as a variable to the function does that.

html/controllers/mailer/contact.php
jpearce185: In some cases, that would be better to use because it specifies which XDMoD the email is from.

classes/MailWrapper.php
smgallo: Edits to @plessbd example for clarity

```php
public static function initPHPMailer($fromEmail = null, $fromName = null);
{
  $mail = new PHPMailer(true);
  $mail->isSendMail();
  $mail->Sender = \xd_utilities\getConfiguration('mailer', 'sender_email');
  $address =  ( null !== $fromEmail ? $fromEmail : $mail->Sender );
  $name = ( null !== $fromName ? $fromName : \xd_utilities\getConfiguration('general', 'title') );
  
  try {
    $mail->setFrom($address, $name);
  }
  catch(phpmailerException $e){
    /*
      Handle invalid sender address here
    */
  }
  
  return $mail;
}  
```

plessbd: Comments on @smgallo edits:
Doing this could end up with it coming from `\xd_utilities\getConfiguration('general', 'title') user@someDomain.tld`

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -141,9 +142,8 @@ public function getLoginLink()
     }
     private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error = false)
     {
-        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
 
-        $mail = MailWrapper::initPHPMailer($sender);
+        $mail = MailWrapper::initPHPMailer($userEmail, \xd_utilities\getConfiguration('general', 'title'));
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

Also why the change from $sender to $userEmail  where is $userEmail defined?

classes/CCR/MailWrapper.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public static function initPHPMailer($fromEmail = null, $fromName = '')
+    {
+        $mail = new \PHPMailer(true);
+        $mail->isSendMail();
+        $address = \xd_utilities\getConfiguration('mailer', 'sender_email');
+        $name = \xd_utilities\getConfiguration('general', 'title');
+        $mail->Sender = $address;
+
+        if($fromEmail !== null) {
```
plessbd: Look at steves example and wrok to find the best of both worlds

classes/XDReportManager.php
```diff
@@ -2045,11 +2046,8 @@ public function mailReport(
         $frequency = '',
         $additional_config = array()
     ) {
-        $sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
 
-        $mail = MailWrapper::initPHPMailer($sender);
-
-        $mail->setFrom($sender, 'XDMoD');
+        $mail = MailWrapper::initPHPMailer(xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

html/controllers/mailer/contact.php
```diff
@@ -116,8 +115,7 @@
 
 // =====================================================
 
-$mail = MailWrapper::initPHPMailer($sender);
-$mail->setFrom($sender, 'XDMoD');
+$mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

html/controllers/user_admin/create_user.php
```diff
@@ -105,12 +107,10 @@
 
     $page_title = xd_utilities\getConfiguration('general', 'title');
     $site_address = xd_utilities\getConfigurationUrlBase('general', 'site_address');
-    $sender = xd_utilities\getConfiguration('mailer', 'sender_email');
 
     // -------------------
 
-    $mail = MailWrapper::initPHPMailer($sender);
-    $mail->setFrom($sender, 'XDMoD');
+    $mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

html/controllers/user_admin/pass_reset.php
```diff
@@ -33,10 +35,7 @@
 
 // -------------------
 
-$sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
-
-$mail = MailWrapper::initPHPMailer($sender);
-$mail->setFrom($sender, 'XDMoD');
+$mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

html/controllers/user_auth/pass_reset.php
```diff
@@ -38,10 +40,7 @@
 
 // -------------------
 
-$sender = strtolower(\xd_utilities\getConfiguration('mailer', 'sender_email'));
-
-$mail = MailWrapper::initPHPMailer($sender);
-$mail->setFrom($sender, 'XDMoD');
+$mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
```
plessbd: new mailwrapper defaults to `\xd_utilities\getConfiguration('mailer', 'sender_email')` and `xd_utilities\getConfiguration('general', 'title')` no need to pass them in

classes/XDReportManager.php
```diff
@@ -2079,19 +2075,23 @@ public function mailReport(
                 break;
         }
 
-        $mail->Subject =
-            "Your$frequency " . $templateConfig['subject'] . " $subject_suffix";
-        $mail->Body = $templateConfig['message'];
+        try {
+            $mail = MailWrapper::initPHPMailer(null, 'XDMoD');
```
plessbd: no need to send this stuff in, it will default to what it needs to

html/controllers/mailer/contact.php
```diff
@@ -86,24 +84,26 @@
...
 
 $message = "Below is a $message_type from '{$_POST['name']}' ({$_POST['email']}):\n\n";
 $message .= $_POST['message'];
 $message .="\n------------------------\n\nSession Tracking Data:\n\n  ";
 $message .="$user_info\n\n  Token:        {$_POST['token']}\n  Timestamp:    $timestamp";
 
-$mail->Body = $message;
-
 try {
+    $mail = MailWrapper::initPHPMailer($_POST['email']);
```
plessbd: Does the post have a name we can attach to the email?

html/controllers/mailer/contact.php
```diff
@@ -129,11 +123,15 @@
     . "Center for Computational Research\n"
     . "University at Buffalo, SUNY\n";
 
-$mail->Body = $message;
-
 // -------------------
 
 try {
+    $mail = MailWrapper::initPHPMailer(null, 'XDMoD');
```
plessbd: no need to pass anything in the defaults will take care of this.

html/controllers/mailer/sign_up.php
```diff
@@ -124,13 +116,23 @@
...
 
 EOMSG;
-$mail->Body = $message;
-
-// Send email.
 
 $response = array();
 
 try {
+    $mail = MailWrapper::initPHPMailer($_POST['email']);
```
plessbd: does the post have a name we can pass in?  Same with the Replyto

html/controllers/user_admin/create_user.php
```diff
@@ -110,7 +110,7 @@
 
     // -------------------
 
-    $mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
+    $mail = MailWrapper::initPHPMailer(null, 'XDMoD');
```
plessbd: No need to pass anything in defaults should be fine here.

html/controllers/user_admin/pass_reset.php
```diff
@@ -35,19 +35,19 @@
 
 // -------------------
 
-$mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
-$mail->Subject = "$page_title: Password Reset";
-$mail->addAddress($recipient);
+try {
+    $mail = MailWrapper::initPHPMailer(null, 'XDMoD');
```
plessbd: no need to pass anything in, defaults should be fine

html/controllers/user_admin/pass_reset.php
```diff
@@ -35,19 +35,19 @@
...
+    $mail->Subject = "$page_title: Password Reset";
+    $mail->addAddress($recipient);
 
-// -------------------
+    // -------------------
 
-$message = MailTemplates::passwordReset($user_to_email);
+    $message = MailTemplates::passwordReset($user_to_email);
```
plessbd: For a future update, we should probably make this MailTemplates part of the mailwrapper class

html/controllers/user_auth/pass_reset.php
```diff
@@ -40,19 +40,19 @@
 
 // -------------------
 
-$mail = MailWrapper::initPHPMailer(\xd_utilities\getConfiguration('mailer', 'sender_email'), 'XDMoD');
-$mail->Subject = "$page_title: Password Reset";
-$mail->addAddress($recipient);
+try {
+    $mail = MailWrapper::initPHPMailer(null, 'XDMoD');
```
plessbd: no need to pass anything in the defaults should be fine.

html/controllers/user_admin/pass_reset.php
jpearce185: On another branch, yes

html/controllers/mailer/contact.php
```diff
@@ -86,43 +84,37 @@
...
 
-$mail->setBodyText($message);
-
 try {
+    $mail = MailWrapper::initPHPMailer($_POST['email'], $_POST['name']);
+
+    $mail->Subject = $subject;
+    $mail->addAddress($recipient);
+
+    //$mail->setFrom($mailer_sender, 'XDMoD');
```
jpwhite4: Please do not add commented out code. See e.g. https://softwareengineering.stackexchange.com/questions/190096/can-commented-out-code-be-valuable-documentation

classes/Authentication/SAML/XDSamlAuthentication.php
plessbd: it is being used https://github.com/ubccr/xdmod/blob/xdmod7.0/html/gui/general/login.php#L14 but looks like it should be able to be pretty easily converted to a false in there.

PR #175, opened 2017-06-28
fixed bug by increasing character count

jpearce185:
I increased the maximum number of characters allowed in the 'Name' field of the contact forms.

## Description
I changed the hard-coded number for maximum character count (100) allowed in the 'Name' field of the contact forms to 130.

## Motivation and Context
If a user had a first and last that is the maximum number of characters long (50 each), and then tried to fill out a contact form, it would say underneath the 'Name' field that the maximum number of characters (100) has been exceeded because the form automatically adds the first name, a space, then the last name (totaling 101).  This would force the user to have to delete characters in their name.  I increased the maximum number of characters allowed in the field as a simple fix and now any user that would have a maximum length first and last name can easily fill out a contact form and submit it.

## Tests performed
I opened up a contact form to fill out and in the 'Name' field, it states the maximum number of characters is 130.  I also created a test account with a first and last name that contained the maximum number of characters each (50 each: 100 total).  I then logged in as that user and opened up a contact form to which the 'Name' and 'Email' fields automatically get filled in, and I was able to submit the form with the 'Name' field consisting of 101 total characters (including the space).

## Types of changes

## Checklist:

### DIFF
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -303,7 +303,7 @@ XDMoD.constants.maxLastNameLength = 50;
 /**
  * The maximum length of a full name.
  */
-XDMoD.constants.maxNameLength = XDMoD.constants.maxFirstNameLength + XDMoD.constants.maxLastNameLength + 1;
+XDMoD.constants.maxNameLength = XDMoD.constants.maxFirstNameLength + XDMoD.constants.maxLastNameLength;
 
 /**
  * The minimum length of an email address.
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -291,11 +291,6 @@ XDMoD.GlobalToolbar.Help = function (tabPanel) {
 // by both the browser client and the server.
 
 /**
- * The maximum length of a full name.
- */
-XDMoD.constants.maxNameLength = 130;
-
-/**
  * The maximum length of a first name.
  */
 XDMoD.constants.maxFirstNameLength = 50;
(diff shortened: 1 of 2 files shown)


jpwhite4:
What happens if a user has more than 130 characters in their name? What is the impact of this limit change on the rest of the code? Will the (now) longer string cause a problem in some other part of the XDMoD software?

jpearce185:
A user cannot have more than 130 characters in their name because when you sign up and become a user, your name (first and last) total up to 100 characters (50 each), and will not allow either your first or last name to be longer than 50 characters each.  Therefore, if a user were to fill out the form they could only ever have a max of 101 characters (because the space counts) and no greater before this limit change.

The point at which more than 101 characters would be used is if:
A) A user adds extra characters to their name, or
B) A guest (someone without an account) fills out a contact form and enters more than 101 characters, in which case if they wanted an account would have to have a 50 character each first and last name

No other code is affected by this change as this limit is only used for the 'Name' field of the contact forms.

smgallo:
The name in the contact form is populated by calling `https://storage-demo.ccr.xdmod.org/rest/v1/users/current`, which returns the first and last name. Both of these have a maximum length of 50 in the database and they are concatenated together when setting the name in the contact form:
```
txtName.setValue(
    data.results.first_name + ' ' + data.results.last_name
);
```
Rather than changing the global name length to 130 and risking adverse effects elsewhere, wouldn't it be a better solution to change the text field validation from `maxLength: maxNameLength` to `maxLength: maxNameLength + 1` since the code is adding an extra space?

jpearce185:
I pushed that "lower character limit' commit just before I saw your comment @smgallo; my apologies.  However, yes, that is a better solution, and I will make the change.

PR #178, opened 2017-06-29
Centralized mail-related code

jpearce185:
## Description
Any code pertaining to creating/sending emails has been moved into a central location.  This allows each file to have only one mail-related function called in it which will handle all mail-oriented functions.

## Motivation and Context
This condenses the amount of code that goes into individual files, and instead puts it all in one file.  This will provide ease when editing code pertaining to mail because it will only need to be done in one place, not several.

## Tests performed
I started by requesting a password reset from the 'Trouble logging in' section within the 'Log in' window using multiple test emails.  From there I proceeded back into the log in window and signed up using a test email.  From here I logged in using a preexisting test account and created a report via the 'Report Generator' tab.  I then sent the report first as a PDF, and secondly as a Word document.  Next, I opened the 'Contact Us' tab, filled out, and submitted each of the three forms.  I then went into the dashboard and created a new user using a test email, and then sent that new user a password reset email as an admin.  I then logged in as this new test user and repeated the Report Generator and Contact form tests.  I then completely logged out and attempted logging in using federated user credentials to which the response I received back was "No peer endpoint available to which to send SAML response," which is the expected response.

## Types of changes

- [x] Code cleanup

## Checklist:


classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -174,8 +157,37 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+        ? xd_utilities\getConfiguration('general', 'debug_recipient')
+        : xd_utilities\getConfiguration('general', 'contact_page_recipient');
+
+        if ($error) {
+            $subject = "[xdmod] Error Creating federated user";
+        } else {
+            $subject = "[xdmod] New " . ($linked ? "linked": "unlinked") . " federated user created";
+        }
+
+        $properties = array(
```
jpwhite4: This pull request currently adds more required boilerplate code to the source rather than reducing it!

Please change this so that you only specify the properties that asre different from the default values.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -174,8 +157,34 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+            'body'=>$body,
+            'subject'=>$subject,
+            'toAddress'=>array(
+                array('address'=>$recipient)
+            ),
+            'fromAddress'=>$userEmail,
+            'fromName'=>$userName,
+            'ifReplyAddress'=>$shouldAddReply
```
plessbd: why dont we just treat this like the FromAddress, just put in the address if needed?

classes/CCR/MailWrapper.php
```diff
@@ -2,23 +2,53 @@
...
 {
 
-    public static function initPHPMailer($fromEmail = null, $fromName = null)
+    public static function initPHPMailer($properties)
     {
         $mail = new \PHPMailer(true);
         $mail->isSendMail();
         $address =  \xd_utilities\getConfiguration('mailer', 'sender_email');
         $mail->Sender = $address;
+        $mail->Body = $properties['body'];
```
plessbd: this should check if the body and subject properties are there, and error if there is no subject or body to the message

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +58,99 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
...
+     * Get the product name for this instance of XDMoD.
+     *
+     * This refers to the name of this software. For the name of this instance
+     * of this software, see getSiteTitle.
+     *
+     * @return string The product name for this instance of XDMoD.
+     */
+    public static function getProductName()
+    {
+        $name = 'Open XDMoD';
```
plessbd: we should change this to use the general title like the others

classes/CCR/MailWrapper.php
```diff
@@ -2,23 +2,53 @@
...
+        }
 
-        if($fromEmail !== null) {
-            $address = $fromEmail;
-            $name = $fromName;
+        if(!empty($properties['fromAddress'])) {
+            $address = $properties['fromAddress'];
+            $name = $properties['fromName'];
         } else {
             $name = \xd_utilities\getConfiguration('general', 'title');
```
plessbd: This should use the function to getSiteTitle or we should get rid of the function, no need for both

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +58,99 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
...
+     *
+     * This refers to the title of this instance of XDMoD. For the name of this
+     * software product, see getProductName.
+     *
+     * @return string The custom title of this site if configured.
+     *                Otherwise, "Open XDMoD".
+     */
+    public static function getSiteTitle()
+    {
+        $title = 'Open XDMoD';
```
plessbd: if we use this I am not sure we should default the title or if we should just use what is in the configs

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +58,99 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
...
+        MailWrapper::sendmail($properties);
+    }
+
+    public function decideTemplate($templateType, $templates)
+    {
+        if($templateType === 'password_reset') {
+            return $templates->getContents();
+
+        /**
+-       * Gets used by reports built and sent via the Report Generator, as
```
plessbd: the leading `-`'s here and below are wrong

classes/XDReportManager.php
```diff
@@ -2055,44 +2055,79 @@ public function mailReport(
...
-            $mail = MailWrapper::initPHPMailer();
-
-            $mail->addAddress($destination_email_address);
-
-            $mail->Subject =
-                "Your$frequency " . $templateConfig['subject'] . " $subject_suffix";
-            $mail->Body = $templateConfig['message'];
+            $subjectType = '';
+            if($templateType === 'custom_report') {
+                $subjectType = 'XDMoD Report';
```
plessbd: these should come from the site title and not be hard coded

classes/XDReportManager.php
```diff
@@ -2055,44 +2055,79 @@ public function mailReport(
...
+                $templateType = 'compliance_report';
                 break;
 
             default:
                 $include_attachment = true;
-                $templateConfig = MailTemplates::customReport(
-                    $report_owner,
-                    $frequency
+
+                $frequency = trim($frequency);
```
plessbd: these should just be passed in the the mail function,  it should not need to do anything different depending on what the template is, it should just take the properties given.

html/controllers/mailer/contact.php
```diff
@@ -121,20 +123,24 @@
     . "Center for Computational Research\n"
```
plessbd: this seems like it should come from the config as well, if someone is running their own xdmod they should be able to set the contact signature

html/controllers/mailer/contact.php
```diff
@@ -121,20 +123,24 @@
...
 
 try {
-    $mail = MailWrapper::initPHPMailer();
-    $mail->Subject = "Thank you for your $message_type.";
-    $mail->addAddress($_POST['email']);
-
-    $mail->Body = $message;
-
-    $mail->send();
+    $properties = array(
```
plessbd: no need to create the properties variable, just pass the array into the function

html/controllers/mailer/sign_up.php
```diff
@@ -121,21 +121,22 @@
...
-    // Request Tracker to function
-    $mail->addReplyTo($_POST['email'], $name);
-
-    $mail->Subject = "[$title] A visitor has signed up";
-
-    $mail->Body = $message;
+$subject = "[$title] A visitor has signed up";
+
+// Original sender's e-mail must be in the "fromAddress" field for the XDMoD Request Tracker to function
+$properties = array(
```
plessbd: no need for the properties variable just pass the array into the function

html/controllers/user_admin/pass_reset.php
```diff
@@ -36,26 +36,39 @@
...
+    $props = array(
+        'first_name'           => $user_to_email->getFirstName(),
+        'username'             => $username,
+        'reset_url'            => $resetUrl,
+        'maintainer_signature' => MailWrapper::getMaintainerSignature()
+    );
 
-    // -------------------
+    $properties = array(
+        'body'=>'',
```
plessbd: are we saying the body should be empty here?  If we are why are we sending an empty email, otherwise, we should just not pass this in as it should be defaulted or overwritten by the template so no need to pass in.

html/controllers/user_admin/pass_reset.php
```diff
@@ -36,26 +36,36 @@
 // -------------------
 
 try {
-    $mail = MailWrapper::initPHPMailer();
-    $mail->Subject = "$page_title: Password Reset";
-    $mail->addAddress($recipient);
+    $subject = "$page_title: Password Reset";
+
+    $username = $user_to_email->getUsername();
```
plessbd: no need to set these into variables if they are just going to be passed into the function call.

html/controllers/user_auth/pass_reset.php
```diff
@@ -41,25 +39,34 @@
...
-    // -------------------
-
-    $message = MailTemplates::passwordReset($user_to_email);
-
-    $mail->Body = $message;
-
-    // -----------------------------
-
-    $status = $mail->send();
+    $subject = "$page_title: Password Reset";
```
plessbd: no need to put these into variable just pass them directly into the fuction.

html/internal_dashboard/controllers/mailer.php
```diff
@@ -92,21 +91,21 @@
...
+                                  'body'=>$message,
+                                  'subject'=>$Subject,
+                                  'toAddress'=>array(
+                                      array('address'=>$contact,
+                                            'name'=>'Undisclosed Recipients'
+                                      )
+                                  ),
+                                  'fromAddress'=>$contact,
+                                  'fromName'=>$title,
+                                  'bcc'=>true
```
plessbd: BCC should be an email address not a boolean

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -143,7 +143,6 @@ public function getLoginLink()
     private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error = false)
     {
         $userEmail = $user->getEmailAddress();
```
plessbd: why not move this one too?

html/controllers/mailer/contact.php
```diff
@@ -96,13 +91,11 @@
     MailWrapper::sendmail(array(
         'body'=>$message,
         'subject'=>$subject,
-        'toAddress'=>array(
-            array('address'=>$recipient)
-        ),
         'fromAddress'=>$_POST['email'],
         'fromName'=>$_POST['name'],
         'ifReplyAddress'=>\xd_utilities\getConfiguration('mailer', 'sender_email')
```
plessbd: having if at the begining seems to indicate that this is a boolean, is it a boolean or is it actaully an email address?  if it is an email address we should not do it as a boolean

html/controllers/mailer/sign_up.php
```diff
@@ -119,20 +114,13 @@
...
         'body'=>$message,
-        'subject'=>$subject,
-        'toAddress'=>array(
-            array('address'=>$recipient)
-        ),
+        'subject'=>"[$title] A visitor has signed up",
         'fromAddress'=>$_POST['email'],
-        'fromName'=>$name,
+        'fromName'=>$_POST['last_name'] . ', ' . $_POST['first_name'],
         'ifReplyAddress'=>\xd_utilities\getConfiguration('mailer', 'sender_email')
```
plessbd: same as previous if this is an email it should not be prefixed with If

html/controllers/user_admin/pass_reset.php
```diff
@@ -49,13 +46,11 @@
...
-            'username'             => $username,
+            'username'             => $user_to_email->getUsername(),
             'reset_url'            => $resetUrl,
             'maintainer_signature' => MailWrapper::getMaintainerSignature(),
-            'subject'=>$subject,
-            'toAddress'=>array(
-                array('address'=>$recipient)
-            )
+            'subject'=>"$page_title: Password Reset",
+            'toAddress'=>$recipient
```
plessbd: this recipient seems to be a lot like the default I wonder if there is a better way to do this...

html/controllers/user_admin/pass_reset.php
```diff
@@ -49,13 +46,11 @@
         'password_reset',
         array(
             'first_name'           => $user_to_email->getFirstName(),
-            'username'             => $username,
+            'username'             => $user_to_email->getUsername(),
             'reset_url'            => $resetUrl,
```
plessbd: no need for resetUrl to be a variable just pass it in

html/controllers/user_auth/pass_reset.php
```diff
@@ -29,20 +29,17 @@
 
 // -----------------------------
 
-$page_title = xd_utilities\getConfiguration('general', 'title');
+$page_title = \xd_utilities\getConfiguration('general', 'title');
 
 $recipient
-    = (xd_utilities\getConfiguration('general', 'debug_mode') == 'on')
-    ? xd_utilities\getConfiguration('general', 'debug_recipient')
+    = (\xd_utilities\getConfiguration('general', 'debug_mode') == 'on')
```
plessbd: same with this recipient, seems like we shouldnt have to do all of this we should be able to just pass in the "real" recipeient and then test for debug mode in the mailwrapper.

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +49,128 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
...
+
+    /**
+     * Get the name of this instance of XDMoD.
+     *
+     * This refers to the title of this instance of XDMoD. For the name of this
+     * software product, see getProductName.
+     *
+     * @return string The custom title of this site if configured.
+     */
+    public static function getSiteTitle()
```
jpwhite4: This function has nothing to do with encapsulating email functionality, so really does not belong in a MailWrapper class. 

I recommend removing this function since it does not actually do anything that is not already being done by the call to getConfiguration()

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +49,128 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
...
+
+    public function sendTemplate($templateType, $properties)
+    {
+        $template = new EmailTemplate($templateType);
+        $template->apply($properties);
+        $properties['body'] = MailWrapper::decideTemplate($templateType, $template);
+        MailWrapper::sendmail($properties);
+    }
+
+    public function decideTemplate($templateType, $templates)
```
jpwhite4: This function performs no useful task and should be removed. It also has a bug if the templateType is not recognised then it falls through without a return statement.

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +49,128 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
 
         return $mail;
     }
+
+    public function sendMail($properties) {
+        $mail = MailWrapper::initPHPMailer($properties);
+        if($mail->send()) {
```
jpwhite4: Please change this code to
```
return $mail->send();
```
to make it more readable

classes/CCR/MailWrapper.php
```diff
@@ -28,4 +49,128 @@ public static function initPHPMailer($fromEmail = null, $fromName = null)
 
         return $mail;
     }
+
+    public function sendMail($properties) {
```
jpwhite4: This function should be static

classes/CCR/XDReportManager.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+namespace CCR;
```
jpwhite4: Please explain your motivation for moving this class into a namespace. Moving the class into a namespace does not appear to me to be directly related to changing the MailWrapper api. I prefer to have separate pull requests for separate unrelated functional changes to the code.

classes/Xdmod/Template.php
```diff
@@ -102,11 +102,13 @@ public function resetContents()
     public function apply(array $items)
     {
         foreach ($items as $param => $value) {
-            $this->contents = preg_replace(
-                "/\[:$param:\]/",
-                $value,
-                $this->contents
-            );
+            if(is_string($param) && is_string($value)) {
```
jpwhite4: What bug is this code change trying to fix? How has this been tested? 

What happened if the data was not a string before this change to the code? What happens if the data is not a string with this code?

html/controllers/mailer/contact.php
```diff
@@ -117,20 +111,17 @@
     = "Hello, {$_POST['name']}\n\n"
     . "This e-mail is to inform you that the XDMoD Portal Team has received your $message_type, and will\n"
     . "be in touch with you as soon as possible.\n\n"
-    . "The TAS Project Team\n"
-    . "Center for Computational Research\n"
-    . "University at Buffalo, SUNY\n";
+    . \xd_utilities\getConfiguration('general', 'maintainer_email_signature');
```
jpwhite4: Again I would like to see code changes that address completely different issues split into different pull requests. This code change modifies the contents of the email that is sent. This is functionally very different from the task of cleaning up the mailwrapper code.

html/controllers/mailer/sign_up.php
```diff
@@ -119,23 +112,16 @@
...
-    // Request Tracker to function
-    $mail->addReplyTo($_POST['email'], $name);
-
-    $mail->Subject = "[$title] A visitor has signed up";
-
-    $mail->Body = $message;
-
-   // Send email.
-    $mail->send();
+    MailWrapper::sendMail(array(
```
jpwhite4: do you need to check the return value of the sendMail() function here? Is it guaranteed to thorw an exception if there is a problem? If so then why does it return a success code?

classes/CCR/XDReportManager.php
jpearce185: Travis continuous-integration had required that I put this class into a namespace.  Since the MailWrapper class also needed to be added to a namespace (CCR) in a previous pull request, it seemed logical to add this class to the same namespace.  If Travis hadn't requested this, then the class wouldn't be in a namespace.

classes/Xdmod/Template.php
jpearce185: This fixes a bug that, when given the array of objects to be used to create mail, would cause a fatal failure when the method came across an item in the array that was anything other than a string.  Since this method is finding values and replacing them with strings to construct the body of the email, it only needs certain elements in the array provided.  This method will only use elements in the array which correspond to what is being replaced, ignoring other values and strings that are unrelated.

Since this constructs the body of some emails, the data provided that this method uses must be strings, anything else will cause an error, but this "if" block prevents the error when looping through the given array.

This has been tested by sending password reset emails (both by the user and admin), and generating and sending a report.  These are the only 3 scenarios that will use this method.

Overall, we only need to use one array, rather than multiple arrays.

html/controllers/mailer/sign_up.php
jpearce185: The return value of sendMail() shouldn't need to be checked here because if there is an issue within the call, an error will be thrown.

As for the success code, since we echo json_encode($response), $response needs to have a value whether the call succeeds or fails, and because we want to see the error when it fails by setting the $response message to the error message, the success code equaling false is not needed.  However, we need to have a value for response if the call succeeds, so true is more like a placeholder.

jpearce185:
# Tests

This is a list of the tests performed and the steps I took to perform these tests.  All of the steps start at the main page, and all tests have passed.  Tests 1 through 3 are performed before logging in.  The rest of the tests are performed after logging in.

## 1. Sign up
- Click the 'Sign in' tab
- Click 'Sign up' and fill out form

## 2. Password Reset (by user)
- Click the 'Sign in' tab
- Click 'Trouble logging in' and enter email

## 3. SAML Authentication (UBIT login)
- Click the 'Sign in' tab
- Click University at Buffalo banner logo and login using UBIT credentials
- Expected response is 'No peer endpoint available'

## 4. Contact (all forms)
- Click 'Contact Us'
- Click 'Send Message' and fill out form
- Repeat for the other 2 forms (Request Feature and Submit Support Request)

## 5. Report Generator
- Go to 'Report Generator' tab
- Select a report or create a new one to select
- Click 'Send Now' and click 'As PDF'
- Repeat, but click 'As Word Document'

## 6. Creating A User
- Navigate to the dashboard
- Switch to the 'User Management' tab
- Click 'Create and Manage Users' and complete the new user form

## 7. Password Reset (by admin)
- Navigate to the dashboard
- Switch to the 'User Management' tab
- Click on the 'Existing Users' tab
- Click 'Create and Manage Users'
- Switch to the 'Current Users' tab and select a user
- Click 'Actions' and then 'Send Password Reset'

html/controllers/report_builder/remove_report_by_id.php
```diff
@@ -1,5 +1,5 @@
 <?php
- 
+
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/remove_chart_from_pool.php
```diff
@@ -1,5 +1,5 @@
 <?php
-   
+
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/enum_available_charts.php
```diff
@@ -30,4 +30,4 @@
 	    
 	}
 
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/enum_reports.php
```diff
@@ -29,4 +29,4 @@
 	    
 	}
 	
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/enum_templates.php
```diff
@@ -25,4 +25,4 @@
 	    
 	}
 	
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/fetch_report_data.php
```diff
@@ -56,4 +56,4 @@
 	    
 	}
 
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/get_new_report_name.php
```diff
@@ -25,4 +25,4 @@
 	    
 	}
 
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

html/controllers/report_builder/get_preview_data.php
```diff
@@ -34,4 +34,4 @@
 	    
 	}
 
-?>
\ No newline at end of file
+?>
```
jpwhite4: Please don't include changes that only add/remote blank lines: This file should not appear in this pull request.

classes/Xdmod/Template.php
jpwhite4: Ok so under what circumstances does the MailWrapper::sendMail() function return false?

smgallo:
** Please hold merging this PR until we have a 7.1 branch.**

PR #183, opened 2017-07-17
Additional Cleanup for Building in PHP 7.0

chakrabortyr:
## Description
Added explicit requires, removed unused 'external_libraries' reference that was breaking the build. Should be pretty minor changes all around, mostly involved cleanup. Only noteworthy change I can think of was removing some concatenation with variables outside of the context in which they'd be known (Line 93, configuration/linker.php)

## Motivation and Context
Cleanup to get it to build properly with PHP 7.0

## Tests performed
Rebuilt XDMOD + ran tests in the open_xdmod module

## Types of changes

## Checklist:


chakrabortyr:
lemme put this on hold, run_tests.sh seems to work fine.

smgallo:
Looks like a style issue. Run PHPCS on any files that you modified (E.g., `~/src/xdmod/vendor/bin/phpcs <file>`)

configuration/linker.php
```diff
@@ -15,6 +15,7 @@
 // Register a custom autoloader for XDMoD components.
 $include_path  = ini_get('include_path');
 $include_path .= ":" . $baseDir . '/classes';
+$include_path .= ":" . $baseDir . '/classes/CCR';
```
jpwhite4: I don't think that we want to make this change to the linker. This effectively changes the interpretation of the namespaces, which I don't think is a good idea.

configuration/linker.php
```diff
@@ -136,73 +138,73 @@ class HttpCodeMessages
  */
 function handle_uncaught_exception($exception)
 {
-   $logfile = LOG_DIR . "/" . xd_utilities\getConfiguration('general', 'exceptions_logfile');
+    $logfile = LOG_DIR . "/" . xd_utilities\getConfiguration('general', 'exceptions_logfile');
 
-   $logConf = array('mode' => 0644);
-   $logger = Log::factory('file', $logfile, 'exception', $logConf);
+    $logConf = array('mode' => 0644);
+    $logger = Log::factory($logfile, $logConf);
```
jpwhite4: Why did you change this function call? Originally the code used the PEAR::Log factory function to generate a file logger. Now you have changes it to use a CCR::Log factory function that does something completely different.

classes/CCR/Log.php
```diff
@@ -2,6 +2,7 @@
 
 namespace CCR;
 
+require_once 'System.php';
```
jpwhite4: What bug this this change intended to fix?

configuration/linker.php
jpwhite4: If you have files that are incorrectly referencing the wrong namespace then please fix those files.

shippable.yml
```diff
@@ -18,3 +18,5 @@ build:
         - composer install --no-progress
         - cp ~/assets/secrets open_xdmod/modules/xdmod/integration_tests/.secrets
         - ./open_xdmod/modules/xdmod/integration_tests/runtests.sh --log-junit `pwd`/shippable/testresults/results.xml
+    on_failure:
+        - cat /var/log/xdmod/*
```
plessbd: there should be a newline at the end of this file.

chakrabortyr:
As of PHP 7.1.x, looks like they replaced InvalidArgumentException with ArgumentCountError (REF: http://php.net/manual/en/class.argumentcounterror.php). This causes IdentityTest::testNoDefaultParameterToConstructor() to look like it's failing, even though it's just using a different exception than prior versions. 

Since fixing this would break the test's compatibility for versions we presently support, I'd say accept it for now. Thoughts?

smgallo:
From the docs: "ArgumentCountError is thrown when too few arguments are passed to a user-defined function or method." and was added in PHP 7.1 while InvalidArgumentException is thrown "if an argument is not of the expected type". It looks like 7.1 now explicitly handles each case. Can PHPUnit run different tests or expect different results based on the version of PHP that is running?

chakrabortyr:
Sorry, not redundant in the sense that there's a test that does something similar, it just seems to be testing a feature of the language itself rather than any code specific to xdmod. It can be assumed that when you initialize a constructor without a parameter (where the constructor specifies for one), that it should fail. So I see it as redundant in that it doesn't seem to test anything specific to our code.

smgallo:
I don't think removing the entire set of tests is a good way to deal with the exception changes in PHP7. There are other valid tests in that file for the Identity class that should not be removed. Even `testNoDefaultParameterToConstructor()` tests to ensure that a parameter is required. Perhaps expecting `BadMethodCallException` (Exception thrown if a callback refers to an undefined method or if some arguments are missing) is a better solution.

PR #184, opened 2017-07-18
Add ui tests

plessbd:
This brings the tests over to  the public repository as committed by @agervase and then updates them for updated eslint rules with some other cleanup.

The following are the previous pull request notes written by @agervase

## Edited Tests Run with WebDriverIO v4
The automated tests have been updated to allow for testing through v4 of WebDriverIO. The new tests run synchronously on the XDMoD reference site, and the use of page objects has made the test spec files significantly less verbose. Versions of the files that run through the SauceLabs automated testing service have also been created.

Within the `tests/automated/test/specs/xdmod/` folder,
* The `o*.js` tests are the locally running synchronous WDIO v4 tests. They make use of the page objects in the `pageObjects/` directory and, at the time of last test, all passed.
* The `s*.js`tests in the `syncTests/` directory run synchronously without the use of page objects.
* The `sauce_*.js` tests in the `sauceTests/` directory run synchronously through SauceLabs using the page objects in the same directory. At the time of last test, all but the few skipped tests passed on the provided Chrome (`chr`) and Firefox (`ff`) configurations.

## Motivation and Context
Where the tests previously ran sequentially within the same browser window, each test now opens, runs within, and closes its own browser window. This not only allows for the tests to be run concurrently, but also prevents a failed test from one spec file affecting the tests from a different spec file. Also, the tests all passing and all being run on the reference set now provides for a baseline for further testing in the future.

The addition of passing SauceLabs tests allows the user to run the tests on different browsers and platforms and concurrently without the increase in loading times that running the tests concurrently and locally causes. The SauceLabs service also allows users to go back and watch videos of each test and view screen shots taken as each command is executed. The user also has the option to manually test in a browser and platform of their choosing, which will allow the tester to try to recreate the events of the tests.

## Changes
* Tests modified to pass on current version of reference set
* Tests now run synchronously and with WDIO v4
* Tests use page objects to make the main spec files less verbose
* Option to run tests on SauceLabs added, testing through other browsers on other platforms now possible


plessbd:
I agree, also moving more into the page objects and using the inheritance like this [POC](https://github.com/plessbd/xdmod/commit/98311c7176486dce7676484ebbebe798f9b8a869)

PR #191, opened 2017-07-21
Removing "impossible" code

ryanrath:
## Description
- per http://php.net/manual/en/function.unset.php
  "It is not possible to unset $this inside an object method since PHP 5."
- that and the Travis tests were failing on php7.1 ( but not 5.3, 5.4, 7.0 )
  #justphpthings

## Motivation and Context
Getting the Travis tests to pass for php7.1

## Tests performed
pushed these changes to a branch w/ Travis enabled, watched all the tests pass this time as opposed to all of the tests except for the ones run under php7.1.

## Types of changes

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -1079,8 +1079,6 @@ class XDUser {
          ':id' => $this->_id,
       ));
 
-      unset($this);
-
    }//removeUser
 
    // ---------------------------


jpwhite4:
What was this unset() code trying to achieve? If you remove it do you need to add an unset call elsewhere?

ryanrath:
'removeUser' is currently called from the following locations:

**XDMoD**:
  - html/controllers/user_admin/delete_user.php:30

**XSEDE:**
  - tools/xdmod-admin/modules.d/user.php:340

In both locations the XDUser object goes out of scope at the end of the request / function. This leads to the memory used by the XDUser object being reclaimed and as such there is no practical change in removing this piece of code.

PR #193, opened 2017-07-25
Update metric explorer tests so they work.

jpwhite4:
## Description
- Add new test that creates and saves a chart
- Update one of the test suites so that it actually checks that the
  charts are correct and runs consistently without having to add pause()
  calls
- Also add a couple of id's to metric explorer menu objects to make
  testing easier.

Note that the existing other tests have been left as is. These need
to be improved.

## Motivation and Context
testing

## Tests performed
yes

### DIFF
--- a/html/gui/js/modules/metric_explorer/MetricExplorer.js
+++ b/html/gui/js/modules/metric_explorer/MetricExplorer.js
@@ -4678,7 +4678,7 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
                 text: pmi,
                 xtype: 'menuitem',
                 menu: new Ext.menu.Menu({
-                    id: 'me_new_chart_submenu_' + pmi,
+                    id: 'me_new_chart_submenu_' + pmi.replace(/\s/g, '_'),
                     ignoreParentClicks: true,
                     items: this.getDisplayTypeItems(undefined,
                         'menuitem',
--- a/html/gui/js/modules/metric_explorer/MetricExplorer.js
+++ b/html/gui/js/modules/metric_explorer/MetricExplorer.js
@@ -4678,6 +4678,7 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
                 text: pmi,
                 xtype: 'menuitem',
                 menu: new Ext.menu.Menu({
+                    id: 'me_new_chart_submenu_' + pmi,
                     ignoreParentClicks: true,
                     items: this.getDisplayTypeItems(undefined,
                         'menuitem',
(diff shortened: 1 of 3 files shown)


jpwhite4:
Note that this build is failing due to a bug in the travis script that is fixed in #194

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -4678,6 +4678,7 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
                 text: pmi,
                 xtype: 'menuitem',
                 menu: new Ext.menu.Menu({
+                    id: 'me_new_chart_submenu_' + pmi,
```
plessbd: is `pmi` here always going to be a valid id (pretty much no spaces)?

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
```diff
@@ -70,6 +132,39 @@ class MetricExplorer {
             .closest('table')
             .attr('id');
     }
+    createNewChart(chartName, datasetType, plotType) {
+        browser.waitForChart();
```
plessbd: Should this be a `waitForChartToChange`

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
```diff
@@ -70,6 +132,39 @@ class MetricExplorer {
...
+        browser.click(this.selectors.toolbar.buttonByName('Refresh'));
+        browser.waitUntilNotExist('.ext-el-mask');
+    }
+    addDataViaCatalog(realm, statistic, groupby) {
+        browser.waitForVisible(this.selectors.catalog.container, 10000);
+        browser.waitUntilAnimEndAndClick(this.selectors.catalog.rootNodeByName(realm));
+        browser.waitUntilAnimEndAndClick(this.selectors.catalog.nodeByPath(realm, statistic));
+        browser.waitForVisible(this.selectors.catalog.addToChartMenu.container);
+        browser.waitAndClick(this.selectors.catalog.addToChartMenu.itemByName(groupby));
+        browser.waitForChart();
```
plessbd: should this be `waitForChartToChange`

html/gui/js/modules/metric_explorer/MetricExplorer.js
jpwhite4: In the current code yes it will be since the different possible values are defined on line 4671 and they do not have spaces in them.

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
jpwhite4: No because it is possible to create a new chart even when there is no existing one. I double checked and waitForChart() actually waits for the chart loading mask to disappear. This is, obviously not a robust way of doing things since there is a race condition in the check running before the mask has been created, but I don't plan to fix this in this pull request.

PR #197, opened 2017-07-27
Prevent ingestion of jobs with 0 start or end time

jtpalmer:
## Description

Changes ingestor SQL query to exclude jobs with a start or end time that is 0.

## Motivation and Context

Jobs with zero start time should not be included in the data warehouse.

## Tests performed

Shredded and ingested data containing jobs with zero start time.  Observed that before the change there were rows in `modw.jobfact` with `start_time_ts` zero before the change, but none after.

## Types of changes

## Checklist:

### DIFF
--- a/classes/OpenXdmod/Ingestor/Shredded/Jobs.php
+++ b/classes/OpenXdmod/Ingestor/Shredded/Jobs.php
@@ -39,12 +39,14 @@ class Jobs extends PDODBMultiIngestor
                 timelimit,
                 node_list
             FROM shredded_job
+            WHERE start_time > 0
+              AND end_time > 0
         ';
 
         $sql = 'SELECT MAX(id) AS max_id FROM staging_job';
         list($row) = $dest_db->query($sql);
         if ($row['max_id'] != null) {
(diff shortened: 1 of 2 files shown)


jpwhite4:
Could you also add a line or two to the documentation (or FAQ perhaps?) to explain this new (desired) behaviour? I'm anticipating someone complaining that the job count is off!

jtpalmer:
Documentation has been added to the FAQ.

jtpalmer:
@smgallo Open XDMoD will [calculate a missing start time if both the end time and wall time are present](https://github.com/ubccr/xdmod/blob/xdmod7.0/classes/OpenXdmod/Shredder.php#L860).  This PR [documents that in the FAQ](https://github.com/ubccr/xdmod/pull/197/files#diff-956da29f025cba37fca9fe6b80ab5d48R115).  I can move that note somewhere else if that would be more appropriate.

smgallo:
Lets leave the logic where it is for now. As we move from the old ETL to ETLv2 we can see if it makes sense to move it elsewhere.

PR #201, opened 2017-08-02
Ensure correct field order in INSERT and implement order_id sequence functionality

smgallo:
Ensure correct field order in INSERT and implement order_id sequence functionality for historical purposes.

## Description

The field list for the INSERT must be in the same order as the fields returned by the query or we will get field mismatches. Use the source record fields (generated from the source query in initialize()) as the correct order. Since the destination field map may have been user-specified we cannot guarantee the order.

Some historical ingestors use an order_id and treat it similar to an auto-increment field, setting its value starting at 0 and incrementing for each record. It is not clear if this is used anywhere, but the functionality is maintained here for compatibility. Note that this does not work when adding fields into a table (e.g.,only when the table is truncated) and will not work properly in cases where the order is relative to a key. For example, if a key is resource_id and the order should be maintained for each unique resource_id we cannot use this method. To not overwrite existing data, only set the order_id if the source field exists and is NULL.

## Motivation and Context

Fixes a bug where the INSERT INTO...SELECT field order for same-database inserts is not the same as the order of the fields in the SELECT.

## Tests performed

Ran tests and performed sample ingestion.

## Types of changes

## Checklist:


classes/ETL/Ingestor/pdoIngestor.php
```diff
@@ -485,7 +496,7 @@ function ($s) {
                 $destColumnList
             );
             $updateColumns = implode(',', $updateColumnList);
-            $sql = "INSERT INTO $qualifiedDestTableName ($destColumns) " . $this->sourceQueryString
+            $sql = "INSERT INTO $qualifiedDestTableName\n($destColumns)\n" . $this->sourceQueryString
```
plessbd: the mix of interpolation and not here is messing with my reading of this...

PR #205, opened 2017-08-03
Modified PHP linker to use configurable autoloaders and include paths

tyearke:
## Description
This pull request allows modules to specify PHP autoloaders and include paths using JSON config files in `etc/linker.d`.

## Motivation and Context
XDMoD requires additional PHP libraries beyond Open XDMoD's requirements. This pull request allows those libraries to be used without having to directly reference them in Open XDMoD.

## Tests Performed
Deployed modified code and verified that the instance worked as expected. Modified deployments with debug code to verify that the final include path stayed identical to the path before this change.

## Types of changes

## Checklist:

### DIFF
--- a/configuration/linker.php
+++ b/configuration/linker.php
@@ -7,24 +7,35 @@ require_once($dir . '/constants.php');
 
 // ---------------------------
 
-// Load linker configuration.
+// If present, load linker configuration from linker.json and linker.d.
 require_once("$baseDir/classes/CCR/Json.php");
 require_once("$baseDir/classes/Xdmod/Config.php");
 $config = Xdmod\Config::factory();
-$linkerConfig = $config['linker'];
+try {
+    $linkerConfig = $config['linker'];
+} catch (Exception $e) {
+    $configDir = $config->getConfigDirPath();
+    echo "Could not find valid \"linker.json\" or \"linker.d\" files in \"$configDir\".\n";
+    echo "Please set up valid linker configuration files and try again.\n";
+    exit(1);
+}
 
 // Load configured autoloaders.
-foreach ($linkerConfig['autoloaders'] as $autoloaderPath) {
-    require_once("$baseDir/$autoloaderPath");
+if (isset($linkerConfig['autoloaders'])) {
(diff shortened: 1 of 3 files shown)


configuration/linker.php
```diff
@@ -7,31 +7,26 @@
 
 // ---------------------------
 
-// Load the autoloader for dependencies managed by Composer.
-require_once($baseDir . '/vendor/autoload.php');
+// Load linker configuration.
+require_once("$baseDir/classes/CCR/Json.php");
+require_once("$baseDir/classes/Xdmod/Config.php");
+$config = Xdmod\Config::factory();
```
smgallo: We need to add some comments in here because it's a pain looking at the `Xdmod\Config` class code to figure out that it's loading the file on demand. Something like "Autoload linker configuration from linker.json" should be sufficient.

configuration/linker.php
```diff
@@ -7,31 +7,26 @@
...
+$linkerConfig = $config['linker'];
```
smgallo: If the file doesn't exist catch the exception and report a useful message.

configuration/linker.php
```diff
@@ -7,31 +7,26 @@
...
-$include_path .= ":" . $baseDir . '/classes/DB/Aggregators';
-$include_path .= ":" . $baseDir . '/classes/DB/DBModel';
-$include_path .= ":" . $baseDir . '/classes/DB/TimePeriodGenerators';
-$include_path .= ":" . $baseDir . '/classes/ExtJS';
-$include_path .= ":" . $baseDir . '/classes/REST';
-$include_path .= ":" . $baseDir . '/classes/User';
-$include_path .= ":" . $baseDir . '/classes/ReportTemplates';
-$include_path .= ":" . $baseDir . '/classes/AppKernel';
-$include_path .= ":" . $baseDir . '/libraries/HighRoller_1.0.5';
+foreach ($linkerConfig['include_dirs'] as $includeDirPath) {
```
smgallo: Should we throw a warning if `include_dirs` doesn't exist?

PR #206, opened 2017-08-04
MariaDB vs MySQL adaptations

ryanrath:
## Description
- Due to the inability of MySQL to handle UTF-8 -> latin1 coercion ( which
  mariadb apparently has no problem with ). We cast all values that are provided
  as strings to 'BINARY' during comparison with column values that are stored as
  latin1 ( Which are also cast to BINARY ). This was done for the following files:
    - acl-config
    - create_public_user.sql

## Motivation and Context
When executing these scripts against a MySQL box the queries involved errored out against a MySQL db but not against a MariaDB instance. 

## Tests performed
Manual Testing: 
- select a query that is currently failing due to illegal mix of collations error.
- execute: SELECT COLLATION('dfsdf');
    - this will tell you what collation is being used for string literals for your db.
- execute: ``` 
     SELECT table_schema, table_name, column_name, character_set_name, collation_name 
     FROM information_schema.columns WHERE table_schema IN ('<your_schema_here>')  AND
          table_name IN ('<your table_name_here>')
     ORDER BY  table_schema, table_name,ordinal_position;```
       - This will tell what encoding / collation is being used for the columns in table_schema.table_name
    - If the collation for the string text is different than that of the columns being compared to then you will need to do some casting wherever the selected query does anything like:
      - `table.column = 'textual data' ` and changing it to 
      - `BINARY table.column = BINARY 'textual data'`

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -223,7 +223,7 @@ function processHierarchies($db, $hierarchies)
        WHERE BINARY m.name = BINARY :module_name
     ) inc
     LEFT JOIN hierarchies cur
-        ON BINARY cur.module_id = BINARY inc.module_id AND
+        ON cur.module_id        = inc.module_id AND
            BINARY cur.name      = BINARY inc.name
     WHERE cur.hierarchy_id IS NULL;
 SQL;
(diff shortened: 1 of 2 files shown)


smgallo:
I agree, any change should be to `utf8mb4`. We have encountered similar issues ingesting data from Postgres which supports 4-byte utf8 characters by default and ended up converting to `latin1` as a temporary solution.

PR #209, opened 2017-08-08
Update Config.php

plessbd:
add check for `Undefined index` and keep mainValue as null allowing exception to be thrown.
This resolves only the issue of notices being thrown when the array key does not exist, this will still throw an error as expected.

## Motivation and Context

Turning notices on on dev caused an undefined index.
This allows the proper response to be sent back.

## Tests performed

added check verified that notice went away

## Types of changes

### DIFF
--- a/classes/Xdmod/Config.php
+++ b/classes/Xdmod/Config.php
@@ -296,7 +296,12 @@ class Config implements ArrayAccess
                 // If the key starts with a "+", merge the values.
 
                 $mainKey   = substr($key, 1);
-                $mainValue = $data[$mainKey];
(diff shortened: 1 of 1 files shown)


classes/Xdmod/Config.php
```diff
@@ -296,7 +296,12 @@ protected function mergeData(array $data, array $partialData)
                 // If the key starts with a "+", merge the values.
 
                 $mainKey   = substr($key, 1);
-                $mainValue = $data[$mainKey];
+                if (array_key_exists($mainKey, $data)){
```
smgallo: Line 333 will also need this check as it may potentially fail.

PR #210, opened 2017-08-08
Remove jQuery as a dependency

plessbd:
## Description

This removes the jquery library as a dependency from the public facing XDMoD
It currently leaves it in the internal dashboard as it is required by the supreem module

## Motivation and Context

Most of what jQuery provides is already provided by extJs.

the things were were using jQuery for

### extend:
This has been added as XDMoD.utils.extend

### base64:

Could be removed if we did not support IE9
https://caniuse.com/#feat=atob-btoa
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob#Browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa#Browser_compatibility

However, since I could not verify that there were still not people using IE9 it was replaced with a small library that did not depend on jquery.

https://github.com/davidchambers/Base64.js

### trim:
Converted as we do not support ie8
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim#Browser_compatibility

## Tests performed

Loaded different pieces of the UI and breakpointed to verify that what was produced is still the same with the new code

Installed Win7 ie9 vm from Microsoft.  Verified that window.atob and window.btoa existed and produced the expected output.

## Types of changes

- [x] Code Cleanup

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "moment/moment-min-file": "^2.13.0",
         "moment/moment-timezone-min-file": "^0.5.4",
         "jquery/jquery-min-file": "^1.12.4",
-        "carlo/jquery-base64-file": "^1.0",
+        "davidchambers/Base64.js-file": "^1.0",
         "tildeio/rsvpjs-min-file": "^3.0.18",
         "jaspersoft/jasperreports-library-jar": "^3.7.6",
         "apache/xalan-j-2jars": "^2.7.1",
@@ -162,15 +162,15 @@
         {
             "type": "package",
             "package": {
-                "name": "carlo/jquery-base64-file",
+                "name": "davidchambers/Base64.js-file",
                 "type": "vanilla-plugin",
                 "version": "1.0",
-                "license": "MIT",
-                "homepage": "https://github.com/carlo/jquery-base64",
+                "license": "Apache 2.0 and WTFPL",
+                "homepage": "https://github.com/davidchambers/Base64.js",
                 "dist": {
-                    "url": "https://github.com/carlo/jquery-base64/raw/master/jquery.base64.js",
+                    "url": "https://raw.githubusercontent.com/davidchambers/Base64.js/master/base64.min.js",
                     "type": "file",
-                    "shasum": "7f0ba311c1676b1b730b29b06e1a9e9e55760acc"
+                    "shasum": "c4f3f169f0ea49756e9ffcd308abfc4f3baeb165"
                 },
                 "require": {
                     "composer/installers": "~1.0"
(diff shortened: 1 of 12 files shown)


plessbd:
## further motivation and context
jQuery 1.x is unsupported and to quote the maintainers:

We don't give any guarantees for jQuery 1.x/2.x - they're officially unsupported - but if a serious vulnerability was detected (& sent to security@jquery.com) we **might** reconsider a patch.

https://github.com/jquery/jquery.com/issues/162#issuecomment-298656430

composer.json
```diff
@@ -162,15 +162,15 @@
...
+                "name": "davidchambers/Base64.js-file",
                 "type": "vanilla-plugin",
                 "version": "1.0",
-                "license": "MIT",
-                "homepage": "https://github.com/carlo/jquery-base64",
+                "license": "Apache 2.0 and WTFPL",
+                "homepage": "https://github.com/davidchambers/Base64.js",
                 "dist": {
-                    "url": "https://github.com/carlo/jquery-base64/raw/master/jquery.base64.js",
+                    "url": "https://raw.githubusercontent.com/davidchambers/Base64.js/master/base64.min.js",
```
chakrabortyr: Not sure I like the concept of adding this. Can you expand your utils namespace?

html/gui/js/CCR.js
```diff
@@ -404,6 +404,48 @@ XDMoD.validator.email = function (email) {
 // A namespace containing commonly-used utility functions.
 
 /**
+ * The jQuery $.extend() API merges the content of subsequent objects into the
+ * first one, overriding it’s original values. The method provided below returns
+ * a new object instead, preserving all of the original objects and their
+ * properties. Supported back to IE6.
+ * https://gomakethings.com/vanilla-javascript-version-of-jquery-extend/
+ */
+XDMoD.utils.extend = function () {
```
chakrabortyr: This file is already huge, can we refactor this to be somewhere else?

composer.json
plessbd: It does base64 encoding and decoding.  If you want to view the unminified source you can view it at https://raw.githubusercontent.com/davidchambers/Base64.js/master/base64.js

It is just a replacement for the jquery.base64.js that was there.

html/gui/js/CCR.js
plessbd: It could be refactored to have the different XDMoD.* in their own files, however I would consider that a different change.

plessbd:
technically this does remove it completely from core xdmod.
A pull request to xdmod-supremm would be needed to remove the requirement for it in there.
If it is decided that this will be accepted I will update xdmod-supremm with the updated code changes and update this to remove jquery resource.

jpwhite4:
I'm still not condoning this change, but I do note that, **if** you remove jQuery you should also remove it from the eslintrc definitions (and possibly the gitignore).

plessbd:
to many people want to keep an unneeded library...

PR #211, opened 2017-08-09
PBS Shredder: Handle case where job hostname is not part of ID string

tyearke:
## Description
This pull request adds support for PBS/Torque logs where the host of a job is not included in the job's ID string. This log data was already being ignored in favor of the resource specified when invoking the shredder, so removing it from this function and the arrays it returns does not have any side effects.

## Motivation and Context
This allows Torque logs that would otherwise cause errors to be successfully ingested.

## Tests Performed
I successfully ingested real-world Torque logs with this change that were causing errors without this change. I also added a new unit test for this case.

## Types of changes

## Checklist:

### DIFF
--- a/classes/OpenXdmod/Shredder/Pbs.php
+++ b/classes/OpenXdmod/Shredder/Pbs.php
@@ -278,8 +278,9 @@ class Pbs extends Shredder
     {
         $this->logger->debug("Parsing id_string '$id'");
 
-        // id_string is formatted as "sequence_number.hostname".
-        list($sequence, $host) = explode('.', $id, 2);
+        // id_string may be formatted as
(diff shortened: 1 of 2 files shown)


classes/OpenXdmod/Shredder/Pbs.php
```diff
@@ -278,8 +278,9 @@ public function getJobIdAndHost($id)
     {
```
jpwhite4: You need to update the documentation comment for this function since it now no-longer determines the hostname for a job.

I would also consider changing the name of the function too since it does not actually get the jobid and host anymore.

smgallo:
Please hold merging for 7.1

PR #216, opened 2017-08-14
Acl Authorization Update

ryanrath:
## Description
- These changes should serve to update the authorization mechanisms currently in
  place to use the Acl functionality as opposed to Roles.
- A set of tests that exercise this code will be included in the xdmod-supremm module as they 
  utilize the Job Viewer rest endpoint.

## Motivation and Context
Further moving towards a single authorization framework. 

## Tests performed
A set of tests have been written that utilize the Single Job Viewer to exercise the new code.

## Types of changes

- [X] Refactor existing code

## Checklist:


classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,50 @@
...
+        $isManager = $user->isManager();
+        $isCenterDirector = $user->hasAcl(ROLE_ID_CENTER_DIRECTOR);
+        $found = $user->hasAcls($requirements);
+
+        // This user is authorized iff:
+        //   - the absence of the requirements was required ($blacklist=true) AND the requirements were not found
+        //   - the presence of the requirements was required ($blacklist=false) AND the requirements were found.
+        $authorized = (!$found && $blacklist) || ($found && !$blacklist);
+
+        if (in_array(STATUS_MANAGER_ROLE, $requirements) && !$isManager) {
```
jpwhite4: Whay do you have explicit code paths for the "manager" and "center director" cases? I thought the point of the ACLS stuff was that we did not have to explicitly hard code special cases into the php. Why are these cases not handled by the hasAcls() function?

classes/NewRest/Controllers/BaseControllerProvider.php
```diff
@@ -268,28 +268,16 @@ public function authorize(Request $request, array $requirements = array(), $blac
     {
         // If role requirements were not given, then the only check to perform
         // is that the user is not a public user.
-        if (empty($requirements)) {
+        if ($requirements === null) {
```
jpwhite4: Is this correct? The function prototype specifies that $requirements is an array therefore it can never be a null. How have you tested this code path?

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,50 @@
...
-     * @throws \Exception
-     */
+     * This function attempts to determine whether or not the provided $user
+     * has the provided $requirements. If $blacklist is supplied then success
+     * will be whether or not the user *does not* have the provided requirements.
+     *
+     * @param XDUser $user the user to authorize
+     * @param array $requirements an array of acl names
+     * @param bool $blacklist whether or not to test for the presence of the $requirements or the absence
+     * @return bool
```
jpwhite4: Why does this function return a bool?

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,50 @@
...
+     * @throws \Exception                if the requirements is not an array or it is an array but has no contents
```
jpwhite4: This document comment looks wrong. I can't see any code that throws an exception if the array has no contents.

jpwhite4: and, of course the function prototype requires it to be an array so you can't call it with something that is not an array/

libraries/security.php
```diff
@@ -200,7 +200,7 @@ function enforceUserRequirements($requirements, $session_variable = 'xdUser')
         if (in_array(SAB_MEMBER, $requirements)) {
 
             // This user must be a member of the Science Advisory Board
-            if (!in_array('sab', $user->getRoles())) {
+            if (!in_array('sab', $user->getAcls(true))) {
```
jpwhite4: Why is this code not changed to use the hasAcl() function? Are we planning to maintain two different APIs to determine whether a user has an ACL? If so what is the justification for this design choice?

classes/NewRest/Utilities/Authorization.php
ryanrath: To keep the expected Exception / Message consistent from the previous version to this one. Re: having hasAcls handle these cases. It didn't seem to be the correct place to migrate this logic as the function of isAuthorized depends on more than just the result of hasAcls ( i.e. the $blacklist parameter ). On the other hand, if we removed the customized exceptions / messages & the blacklist parameter ( which, from what I can tell is never actually used ) then this function could be greatly simplified.

classes/NewRest/Utilities/Authorization.php
ryanrath: no logical reason to be honest. I'll be making a commit shortly that removes it.

classes/NewRest/Controllers/BaseControllerProvider.php
ryanrath: It is not, I had a previous set of fixes in my head where uses of empty were replaced with null. As you pointed out it doesn't apply in this case. I'll have this change reverted in a commit presently.

classes/NewRest/Utilities/Authorization.php
ryanrath: Yeah... I honestly don't have a clue on this one. The only thing I can surmise is that it applied to a previous version of the function and was not updated. Commit incoming.

libraries/security.php
ryanrath: It certainly could be implemented with the hasAcl function, but utilizing the getAcls(boolean) function seemed to require less of a code change. Upon looking this over / thinking about it more it certainly does make sense to use the functions for what they were intended. I'll toss this into the upcoming set of commits.

classes/NewRest/Utilities/Authorization.php
jpwhite4: Ok. but this new code has different behaviour from the old code. In the old code it would throw an Unauth execption for a public user and include the 'not a mangager' or 'not a center director' in the message this new code does not.

jpwhite4: Also the old code had a code path for SAB member and this new code does not.

jpwhite4:
In the pull request you state that 'A set of tests have been written that utilize the Single Job Viewer to exercise the new code.' These should be included in this pull request.

classes/NewRest/Utilities/Authorization.php
ryanrath: I'm a little confused, lines 39 and 41 throw an AccessDeniedHttpException w/ added "[ Not a Manager ]" / "[ Not a Center Director ]" while line 44 takes care of the UnauthorizedHttpException for public users. 

I will give you the missing 'SAB' member code path, this shouldn't have been removed without removing the other 'SAB' handling sections. It was removed because ( if I remember correctly ) of a conversation I had with @smgallo . The upshot being that SAB was no longer used.

ryanrath:
I'll need to modify that to instead reference the xdmod-supremm module ( and put up a new pr ) as they will never pass if included in xdmod ( which is as it should be since it's testing via functionality provided by the xdmod-supremm module ).

classes/NewRest/Utilities/Authorization.php
jpwhite4: By code inspection it looks to me like to have made a functional change to the behaviour of the code.

The old code would 
```
              throw new UnauthorizedHttpException('xdmod', $message); 
```
for public users. Where the `$message` contained 'Not a manager' or 'Not a center director' as appropriate.

The new code throws an AccessDenied exception for public users with the message 'Not a manager' or 'Not a center director' as appropriate.

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,48 @@
...
-                }
-
+        $isManager = $user->isManager();
+        $isCenterDirector = $user->hasAcl(ROLE_ID_CENTER_DIRECTOR);
+        $found = $user->hasAcls($requirements);
+
+        // This user is authorized iff:
+        //   - the absence of the requirements was required ($blacklist=true) AND the requirements were not found
+        //   - the presence of the requirements was required ($blacklist=false) AND the requirements were found.
+        $authorized = (!$found && $blacklist) || ($found && !$blacklist);
```
jpwhite4: perhaps:
```
$authorized = $found ^ $blacklist;
```
to improve readability.

plessbd: I would agree
and just for "proof" cause I had to do a truth table to figure out what it was doing

```
XOR
Found   Blacklist Result
true    true      0
false   false     0
true    false     1
false   true      1

WITHOUT XOR
Found   Blacklist Result
true    true      bool(false)
false   false     bool(false)
true    false     bool(true)
false   true      bool(true)
```

```php
<?php
echo "XOR\n";
echo "Found\tBlacklist\tResult\n";
echo "true\ttrue\t\t" . (true ^ true) ."\n";
echo "false\tfalse\t\t" . (false ^ false) ."\n";
echo "true\tfalse\t\t" . (true ^ false) ."\n";
echo "false\ttrue\t\t" . (false ^ true) ."\n";

echo "\n\nWITHOUT XOR\n";
echo "Found\tBlacklist\tResult\n";

$found = $blacklist = true;

$result = (!$found && $blacklist) || ($found && !$blacklist);
echo "true\ttrue\t\t";
var_dump($result);

$found = $blacklist = false;

$result = (!$found && $blacklist) || ($found && !$blacklist);
echo "false\tfalse\t\t";
var_dump($result);

$found = true;
$blacklist = false;
$result = (!$found && $blacklist) || ($found && !$blacklist);
echo "true\tfalse\t\t";
var_dump($result);

$found = false;
$blacklist = true;
$result = (!$found && $blacklist) || ($found && !$blacklist);
echo "false\ttrue\t\t";
var_dump($result);
```

classes/NewRest/Utilities/Authorization.php
```diff
@@ -9,41 +9,40 @@
...
-            throw new AccessDeniedHttpException(self::_DEFAULT_MESSAGE . "\n[ Not a Manager ]");
-        } elseif (in_array(STATUS_CENTER_DIRECTOR_ROLE, $requirements) && $isCenterDirector === true) {
-            throw new AccessDeniedHttpException(self::_DEFAULT_MESSAGE . "\n [ Not a Center Director ]");
-        } elseif($authorized === false) {
-            if ($user->isPublicUser()) {
-                throw new UnauthorizedHttpException('xdmod', self::_DEFAULT_MESSAGE . " [ Not Authorized ]"); // 401 from framework
-            }
-            throw new AccessDeniedHttpException(self::_DEFAULT_MESSAGE . " [ Not Authorized ]"); // 403 from framework
+        $messageSuffix =  " [ Not Authorized ]";
+        if (in_array(self::_SAB_MEMBER, $requirements) && !$isSabMember) {
```
plessbd: is there a need to have $isSabMember, or could we just make this

```php
if (in_array(self::_SAB_MEMBER, $requirements) && !$user->hasAcl('sab'))
```

classes/NewRest/Utilities/Authorization.php
```diff
@@ -9,41 +9,40 @@
...
+            $messageSuffix = "\n[ Not a SAB Member ]";
+        } elseif (in_array(ROLE_ID_MANAGER, $requirements) && !$isManager) {
```
plessbd: is there a need to have $isManager, or could we just make this

```php
if (in_array(self::_SAB_MEMBER, $requirements) && !$user->isManager())
```

classes/NewRest/Utilities/Authorization.php
```diff
@@ -9,41 +9,40 @@
...
+            $messageSuffix = "\n[ Not a Manager ]";
+        } elseif (in_array(ROLE_ID_CENTER_DIRECTOR, $requirements) && !$isCenterDirector) {
```
plessbd: is there a need to have $isSabMember, or could we just make this

```php
if (in_array(self::_SAB_MEMBER, $requirements) && !$user->hasAcl('ROLE_ID_CENTER_DIRECTOR'))
```

open_xdmod/modules/xdmod/tests/lib/NewRest/Utilities/AuthorizationTest.php
```diff
@@ -0,0 +1,257 @@
...
+                            : 0;
+                    }
+                    return $total === count($requested);
+                }
+                return false;
+            }
+        );
+        return $stub;
+    }
+}
```
plessbd: missing newline at end of file

open_xdmod/modules/xdmod/tests/lib/NewRest/Utilities/AuthorizationTest.php
```diff
@@ -0,0 +1,260 @@
+<?php
+
+namespace NewRest\Utilities;
+
+require_once __DIR__ . './../../../bootstrap.php';
```
jpwhite4: You should not be requireing bootstap.php in test code. The phpunit framework handles this for you.

open_xdmod/modules/xdmod/tests/lib/NewRest/Utilities/AuthorizationTest.php
```diff
@@ -0,0 +1,257 @@
...
+        $newClass = get_class($newException);
+        $oldMessage = $oldException !== null ? $oldException->getMessage() : null;
+        $newMessage = $newException !== null ? $newException->getMessage() : null;
+
+        $this->assertEquals($oldClass, $newClass, "Classes Differ, expected: $oldClass, received: $newClass, for [$user]");
+        $this->assertEquals($oldMessage, $newMessage, "Messages Differ, expected: $oldMessage, received; $newMessage, for [$user]");
+    }
+
+    /**
+     * This function represents the state of isAuthorized prior to the acl
```
jpwhite4: I'm not sure that this is the best way to test backwards compatibility since you are actually reimplementing old code within a test harness. 

The way I would test backwards compatibilty is that I would write a test that checks the expected behaviour then I woudl run that test on the old code and confirm that it passes. I would then run the _same_ test on the new code and comfirm that it passes. The test case itself would then be checked in in the pull request.

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,45 @@
...
 
 class Authorization
 {
-
-    const _DEFAULT_SESSION_KEY = 'xdUser';
-    const _SUCCESS = 0;
-    const _MESSAGE = 1;
-
     const _DEFAULT_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure.';
+    const _SAB_MEMBER = 'sab';
```
jpwhite4: Please do not define a new constant called _SAB_MEMBER when there is already a SAB_MEMBER constant that has different semantics. Also is this really the appropriate place to define a constant like this? Where else is this being used in the code?

open_xdmod/modules/xdmod/tests/lib/NewRest/Utilities/AuthorizationTest.php
```diff
@@ -0,0 +1,257 @@
...
+     * re-write. It is included here for testing backwards compatibility.
+     *
+     * @param mixed $user
+     * @param array $requirements
+     * @return array in the form array( boolean success, string message )
+     * @throws \Exception
+     */
+    public static function isAuthorized($user, array $requirements = array())
```
jpwhite4: I really really do not like taking old code, making changes to it then putting in a test harness. If you really want to do this then at least make the code identical. You must include the blacklist variable in the function and include this in the test parameters.

ryanrath: I'm cool with changing it. I just wanted to fix the integration tests before changing anything else.

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,45 @@
...
-                    if ($found === 0) {
-                        $result[self::_SUCCESS] = true;
-                        $result[self::_MESSAGE] = '';
-                    } else {
-                        $result[self::_MESSAGE] .= " [ Not Authorized ]";
-                    }
-                }
-
-            }
+        $found = $user->hasAcls($requirements);
```
plessbd: something about this just seems off.
It seems like the hasAcls should be throwing an error if the user does not have the ACL's required...

classes/NewRest/Utilities/Authorization.php
```diff
@@ -2,96 +2,45 @@
...
+        $authorized = ($found && !$blacklist) || (!$found && $blacklist);
```
plessbd: this should be using XOR or ^ instead of this crazyness.

classes/NewRest/Controllers/BaseControllerProvider.php
```diff
@@ -34,6 +34,8 @@
 
     const KEY_PREFIX = 'prefix';
 
+    const _DEFAULT_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure. [ Not Authorized ]';
```
plessbd: Thought we were trying to not do the whole dnagling _ thing

libraries/security.php
```diff
@@ -200,7 +200,7 @@ function enforceUserRequirements($requirements, $session_variable = 'xdUser')
         if (in_array(SAB_MEMBER, $requirements)) {
 
             // This user must be a member of the Science Advisory Board
-            if (!in_array('sab', $user->getRoles())) {
+            if (!$user->hasAcl('sab')) {
```
plessbd: is SAB used anywhere?

classes/NewRest/Controllers/BaseControllerProvider.php
ryanrath: dnagling?

libraries/security.php
ryanrath: good question, but not one that's meant to be addressed in this pull request.

classes/NewRest/Controllers/BaseControllerProvider.php
plessbd: dangling (the preceding _ )

ryanrath: Maybe? I don't think that I've been a part of any conversations where the underscore prefix was specifically called out as offensive. I do remember one in which Joe mentioned not having an underscore version of a const that already existed else where sans prefix ( SAB_MEMBER ( number ) vs. _SAB_MEMBER ( string ). But other than that I'm at a loss.

libraries/security.php
plessbd: So is a future PR planned to remove the SAB functionality?  Since this is already changing functionality I would see this as a place where this unused functionality could be excised, if so desired.

ryanrath: Functionally, these two lines are the same. I can either add this to the list of things to be removed in the: 'remove active / primary role' PR or I can just have another PR that solely deals with removing 'SAB' references.

classes/NewRest/Controllers/BaseControllerProvider.php
plessbd: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#42-properties

ryanrath: I'm fine with this, I would appreciate it if Travis can be made to throw an error when these cases are encountered. That way they can be fixed before a PR is submitted.

classes/NewRest/Controllers/BaseControllerProvider.php
```diff
@@ -34,7 +34,7 @@
 
     const KEY_PREFIX = 'prefix';
 
-    const _DEFAULT_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure. [ Not Authorized ]';
+    const DEFAULT_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure. [ Not Authorized ]';
```
plessbd: would prefer this be, `UNAUTORIZED_MESSAGE` or something less generic than `DEFAULT_MESSAGE`.

ryanrath: So besides the fixing of the spelling... how do you feel about an UNAUTHORIZED_MESSAGE being used for an AccessDeniedHttpException?

classes/NewRest/Controllers/BaseControllerProvider.php
plessbd: You can blame me (specifically) and the files like classes/CCR/DB/PDODB.php (generally)(https://github.com/ubccr/xdmod/commit/eda2ab131ddf00826dc89cc801ef15ad8430af82)

classes/NewRest/Controllers/BaseControllerProvider.php
plessbd: I guess it should be something else then, like `EXCEPTION_MESSAGE` as you mentioned

classes/NewRest/Controllers/BaseControllerProvider.php
```diff
@@ -34,7 +34,7 @@
 
     const KEY_PREFIX = 'prefix';
 
-    const DEFAULT_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure. [ Not Authorized ]';
+    const EXCEPTION_MESSAGE = 'An error was encountered while attempting to process the requested authorization procedure. [ Not Authorized ]';
```
plessbd: So now the question becomes, should we be putting `[ Not Authorized ]` in a message that is throwing an Access Denied.

I would say we get rid of that part of message.  Let the Error Code speak for itself.

ryanrath: The `[ Not Authorized ]` was put there to differentiate it from the other, more specific, exception messages. Seeing as how we don't have those anymore I'd be fine with removing this.

PR #220, opened 2017-08-16
Support references into complex source records such as JSON objects

smgallo:
Support references into complex source records such as JSON objects, improve verification of ETLv2 configuration files, add support for ETL variables in source fields.

### Complex Source Records

Structured file endpoints may support complex data records where the value of a field is not a
scalar but an array or an object (e.g., a JSON object). These require a method for handling complex
records and addressing individual fields within those records during the ingest process. Source
field names specified in an action's destination field map may be simple scalars (e.g., simply the
name of the source field) or references to data with a complex record with a format that is specific
to the particular endpoint (e.g., JSON pointers for referencing data in a JSON object). The `DirectoryScanner` endpoint is configurable and can use **any** StructuredFile endpoint as a
handler to process files that it discovers.  Therefore, it must be able to proxy support for complex source records to the handler.

While the implementation will be specific to the endpoint, one use case is a JSON object as a source
record. In this case, we support using a JSON pointer to reference data fields inside of an object.
For example, given the source record

```
{
    "id": "887233-bya",
    "instance_type": {
        "name": "m1.medium",
        "num_cores": 8
    }
}
```

we can specify scalar fields in the object root and use JSON pointers to reference data within nested
objects

```
"destination_field_map": {
    "instance_types": {
        "id": "instance_id",
        "name": "/instance_type/name",
        "num_cores": "/instace_type/cpu"
    }
}
```
### Improve verification of ETLv2 configuration files

The field list for the INSERT must be in the same order as the fields returned by the query or we will get field mismatches. Use the source record fields (generated from the source query in initialize()) as the correct order. Since the destination field map may have been user-specified we cannot guarantee the order.

### ETL variables in source fields

We now support ETL variables as source data. For example, to set the resource_id to the value of the ${RESOURCE_ID} variable:

```
"destination_field_map": {
    "instance_types": {
        "id": "instance_id",
        "resource_id": "${RESOURCE_ID}",
        "name": "/instance_type/name",
        "num_cores": "/instace_type/cpu"
    }
}
```
## Motivation and Context

Need a clean way for data endpoints to allow support for accessing data in complex data records.

## Tests performed

Ran unit tests and ingestion for XDMoD-VA, Resource Allocations, common job tables, and XSEDE jobs data.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/DbModel/Column.php
+++ b/classes/ETL/DbModel/Column.php
@@ -162,7 +162,7 @@ class Column extends NamedEntity implements iEntity
             //
             // **WARNING**
             //
-            // Having multime TIMESTAMP columns in the same table may result in unexpected behavior
+            // Having multiple TIMESTAMP columns in the same table may result in unexpected behavior
             // for the 2nd or following columns.
             //
             // See:
--- a/classes/ETL/DbModel/Column.php
+++ b/classes/ETL/DbModel/Column.php
@@ -154,11 +154,27 @@ class Column extends NamedEntity implements iEntity
                 $srcExtra = str_ireplace($search, "CURRENT_TIMESTAMP", $srcExtra);
             }  // if ( null !== $srcExtra)
 
-            // If no DEFAULT and no EXTRA is provided, MySQL will use:
+            // If no DEFAULT and no EXTRA is provided MySQL will use:
             // DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
             //
             // If no DEFAULT is provided but an EXTRA is provided the default will be 0 unless
             // the collumn is nullable, then it will be NULL.
+            //
+            // **WARNING**
+            //
+            // Having multime TIMESTAMP columns in the same table may result in unexpected behavior
+            // for the 2nd or following columns.
+            //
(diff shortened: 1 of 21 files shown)


classes/ETL/DbModel/Column.php
```diff
@@ -154,11 +154,27 @@ public function compare(iEntity $cmp)
...
-            // If no DEFAULT and no EXTRA is provided, MySQL will use:
+            // If no DEFAULT and no EXTRA is provided MySQL will use:
             // DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
             //
             // If no DEFAULT is provided but an EXTRA is provided the default will be 0 unless
             // the collumn is nullable, then it will be NULL.
+            //
+            // **WARNING**
+            //
+            // Having multime TIMESTAMP columns in the same table may result in unexpected behavior
```
plessbd: multime?

smgallo: It's shorthand for multiple time.

plessbd: multiple time TIMESTAMP sounds good to me :)

PR #221, opened 2017-08-18
Fix ETL overseer exit status

smgallo:
## Description
When calling `exit()` with a string you cannot also return an exit status of 1. Log and write errors to `STDERR` and call `exit(1)` to indicate a failure.

## Tests performed
Tested with and without JSON syntax errors during action verification.

## Types of changes

## Checklist:

### DIFF
--- a/tools/etl/etl_overseer.php
+++ b/tools/etl/etl_overseer.php
@@ -398,10 +398,9 @@ try {
     $etlConfig->setLogger($logger);
     $etlConfig->initialize();
 } catch ( Exception $e ) {
-    $msg = sprintf("%s%s%s", $e->getMessage(), PHP_EOL, $e->getTraceAsString());
-    $logger->err($msg);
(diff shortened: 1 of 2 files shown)


tools/etl/etl_overseer.php
```diff
@@ -560,7 +574,10 @@ function ($key) {
 try {
     $result = $utilityEndpoint->getHandle()->query("SELECT id, code from {$utilitySchema}.resourcefact");
 } catch (Exception $e) {
-    exit($e->getMessage() . "\n". $e->getTraceAsString() . "\n");
+    $msg = sprintf("%s%s%s", $e->getMessage(), PHP_EOL, $e->getTraceAsString());
```
plessbd: it looks like we do this same block of code 5 times, would it be better to have a 
function that does this in one place instead of 5?

smgallo: Yes, I suppose so.

PR #237, opened 2017-09-13
update to secret nginx documentation

plessbd:
updated location directives to work with default centos7 nginx install

## Description

Mainly so the knowledge doesnt get lost if someone else wants to try the unsupported (for now) nginx configuration

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -64,13 +64,6 @@ server {
     fastcgi_param  SCRIPT_FILENAME $document_root$fastcgi_script_name;
     include fastcgi_params;
   }
-  location ~ ^/extrest {
-    try_files (.*) /extrest/index.php break;
-    fastcgi_pass 127.0.0.1:9000;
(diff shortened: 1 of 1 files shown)


docs/nginx.md
```diff
@@ -58,14 +58,18 @@ server {
...
   }
   location ~ ^/extrest {
-    root       /usr/xdmod/share/html;
-    try_files  (.*) /extrest/index.php;
-    include    conf.d/php-fpm;
+    try_files (.*) /extrest/index.php break;
+    fastcgi_pass 127.0.0.1:9000;
+    fastcgi_index index.php;
+    fastcgi_param  SCRIPT_FILENAME $document_root$fastcgi_script_name;
+    include fastcgi_params;
```
jtpalmer: It looks like all of the `extrest` endpoints have been removed (although there is still some configuration in `portal_settings.ini`).  I don't think it was ever used in Open XDMoD.

PR #245, opened 2017-09-19
Don't assume ACLs will contain both realm and tab permissions

smgallo:
## Description

The current `bin/acl-config` script assumes that each ACL will specify both query descriptors for a realm and permitted_modules (e.g., tabs). This is not the case for ACLs granting access to only a tab such as the Custom Query Tab.

Don't assume ACLs will contain both realm and tab permissions.

## Motivation and Context

Bugfix.

## Tests performed

Ran `bin/acl-config` script and observed new entries in the database. Previously the script failed with an "index not found" error.

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -291,26 +291,30 @@ function processResult(iDatabase $db, $module, $moduleData, $modules)
 
     global $log;
 
-    $acls = $moduleData['acls'];
-    $realms = $moduleData['realms'];
+    // Data for a module may contain a list of ACLs, a list of Realms and the access that is
+    // granted, or both.
+
+    $acls = ( isset($moduleData['acls']) ? $moduleData['acls'] : null );
+    $realms = ( isset($moduleData['realms']) ? $moduleData['realms'] : null );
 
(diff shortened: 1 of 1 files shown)


bin/acl-config
```diff
@@ -291,26 +291,30 @@ function processResult(iDatabase $db, $module, $moduleData, $modules)
...
 
-    if ($acls == null) {
-        $log->err("Unable to process $module, missing role information. Is there a $module.json file in CONF_DIR/modules.d?");
-    } else if ($acls !== null && $realms !== null) {
-        processAcls($db, $module, $realms, $acls);
+    if ($acls !== null) {
+        processAcls($db, $module, ( null !== $realms ? $realms : array() ), $acls);
     } else {
-        $log->warning("No acl or realm information for module $module skipping...");
+        $log->err("Unable to process $module, missing role information. Is there a $module.json file in CONF_DIR/modules.d?");
```
jpwhite4: Pehaps change the error message to say mssing 'acl' information rather than 'role'?

PR #249, opened 2017-09-20
Made Timeframes Editable for Single Charts

chakrabortyr:
## Description
Prior to this change, single charts were not editable via the "Edit Timeframe for Selected Charts" button on the top of the Report Generator tab. This change remedies that. To not have 3 pieces of functionality that do the exact same thing, you can no longer edit timeframes by clicking the date range in the chart table.

BEFORE:
screenshot from 2017-09-20 11-29-44

AFTER:
screenshot from 2017-09-20 11-30-05

## Motivation and Context
The prior behavior was counterintuitive.

## Tests performed
Naturally

## Types of changes

## Checklist:


html/gui/js/report_builder/ChartDateEditor.js
```diff
@@ -244,22 +244,16 @@ XDMoD.Reporting.ChartDateEditor = Ext.extend(Ext.Window,  {
...
         }
 
-        if (chart_config.type.toLowerCase() == 'user defined') {
-
-           rdoGrpTimeframeMode.items.get(0).setValue(true);
-
-           this.start_date_field.setValue(chart_config.start);
-           this.end_date_field.setValue(chart_config.end);
-
+        if (chart_config.type && chart_config.type.toLowerCase() === 'user defined') {
```
plessbd: why did this code block change?  does not seem related to your change...

html/gui/js/report_builder/ReportCreatorGrid.js
```diff
@@ -491,7 +491,7 @@ XDMoD.ReportCreatorGrid = Ext.extend(Ext.Panel,  {
 
             var getData = sm.getSelections();
 
-            if (getData.length >= 2) {
```
plessbd: is there a reason this was forced to be more than two to start with?

html/gui/js/report_builder/ChartDateEditor.js
chakrabortyr: While debugging my own code, I noticed an error when chart_config.type was null, so I figured I'd take care of it as well.

html/gui/js/report_builder/ReportCreatorGrid.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
plessbd: We cant just disable the linting on an entire file.

html/gui/js/report_builder/Reporting.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
plessbd: we cant just disable the linting on an entire file.

html/gui/js/report_builder/ChartDateEditor.js
```diff
@@ -244,22 +244,16 @@ XDMoD.Reporting.ChartDateEditor = Ext.extend(Ext.Window,  {
...
 
-        if (chart_config.type.toLowerCase() == 'user defined') {
-
-           rdoGrpTimeframeMode.items.get(0).setValue(true);
-
-           this.start_date_field.setValue(chart_config.start);
-           this.end_date_field.setValue(chart_config.end);
-
+        /* eslint-disable */
+        if (chart_config.type && chart_config.type.toLowerCase() === 'user defined') {
```
plessbd: if the only changes are adding the `===` and checking the existance og `chart_config_type` this should not need linting disabled.

plessbd:
I think the rule changes should be limited in scope as much as possible instead of making these global, make them for the specific file (https://eslint.org/docs/user-guide/configuring#configuring-rules) with comments of why there were set the way they are, so that in the future if we want to try to fix them we can.

html/gui/js/report_builder/ChartDateEditor.js
```diff
@@ -244,22 +244,17 @@ XDMoD.Reporting.ChartDateEditor = Ext.extend(Ext.Window,  {
...
         }
 
-        if (chart_config.type.toLowerCase() == 'user defined') {
-
-           rdoGrpTimeframeMode.items.get(0).setValue(true);
-
-           this.start_date_field.setValue(chart_config.start);
-           this.end_date_field.setValue(chart_config.end);
-
+        /* eslint-disable */
```
jpwhite4: Please don't disable the linter. If there is a rule that you don't like then we can discuss and remove it from the project.

html/gui/js/report_builder/Reporting.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable indent*/
```
plessbd: why the different styles of this disable
`/* eslint-disable indent*/`
vs
`/* eslint-disable indent: 0`
I think we should try to be consistent

PR #250, opened 2017-09-20
Do not show future dates in lower right of XDMoD plots

smgallo:
This addresses [Tom Furlani's bug](https://app.asana.com/0/387470245161503/424407219764834).

## Description

The XDMoD UI presents data aggregated at particular units (day, month, quarter, year). It is
sometimes necessary to adjust the user-specified start or end date to align with aggregation
unit boundaries as we do not display partial aggregation units (the user must switch to a
finer-grained unit for this). This means that the start date may need to be adjusted to the
start of the unit that it falls into, similarly the end date may need to be adjusted to align
with the end date of the unit. For example, viewing a monthly aggregation unit the date range
2017-1-15 - 2017-08-20 must be adjusted to 2017-01-01 - 2017-08-31 to include whole months
since that is the data that will be returned.

XDMoD does not currently provide data for the future. For end dates that fall on the current date or in
the future, do not automatically adjust the date to the end of the aggregation period that it falls into.
Rather, adjust it to the end of the current day as this is the data that will be displayed.

## Motivation and Context

When viewing "to date" plots (e.g., "Year To Date"), adding a date in the future is misleading. If we were to plot 2 "Year to date" plots, one on 2017-9-4 and another on 2017-09-30 both plots would state that the date range was 2017-01-01 - 2017-09-30 but would contain different data.

## Tests performed

Compared my development branch to xdmod-dev for several duration selections (including an end date in the future) in both timeseries and aggregate mode. Also verified that report dates are correct in the Report Generator.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Visualization/HighChart2.php
+++ b/classes/DataWarehouse/Visualization/HighChart2.php
@@ -207,14 +207,28 @@ class HighChart2
         return $pointInterval * 24 * 3600 * 1000;
     } // getPointInterval()
 
-    // ---------------------------------------------------------
-    // setDuration()
-    //
-    // Format and adjust start and end date for chart, given
-    // selected aggregation unit.
-    //
-    // Note: Data related
-    // ---------------------------------------------------------
+    /**
+     * Format and adjust start and end date for chart, given selected aggregation unit.
+     *
+     * The XDMoD UI presents data aggregated at particular units (day, month, quarter, year). It is
+     * sometimes necessary to adjust the user-specified start or end date to align with aggregation
+     * unit boundaries as we do not display partial aggregation units (the user must switch to a
+     * finer-grained unit for this). This means that the start date may need to be adjusted to the
+     * start of the unit that it falls into, similarly the end date may need to be adjusted to align
+     * with the end date of the unit. For example, viewing a monthly aggregation unit the date range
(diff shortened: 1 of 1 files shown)


plessbd:
Do we have any tests for this?

smgallo:
There are no automated tests at the moment. All tests that I did were manual comparisons.

classes/DataWarehouse/Visualization/HighChart2.php
```diff
@@ -256,12 +275,22 @@ public function setDuration($startDate, $endDate, $aggregationUnit, $min_aggrega
             default:
             case 'Day':
             case 'day':
```
chakrabortyr: Nitpick: can't these just be converted to lowercase prior to the switch?

plessbd:
Could you add a screenshot of before and after so that in the future an automated test could be more easily made?

If possible can we use dates for data in the reference database, or maybe we need to add some more data to the reference db to do the test.

classes/DataWarehouse/Visualization/HighChart2.php
smgallo: That would require a much larger change that is not within the scope of this change. The aggregation units are typically set in `html/gui/js/DurationToolbar.js` and are capitalized but I'm not sure where else they may be fed from.

PR #265, opened 2017-09-26
Throw exception if Public User data is not found

jtpalmer:
## Description

See title.  Any suggestions for a better error message are welcome.

## Motivation and Context

If the public user data is missing, the portal will not be displayed and a blank screen is presented instead.  This happens if the database is not migrated from 6.6. to 7.0.  This PR changes the behaviour to display:

```
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"Public User not found","code":0}
```

Which at least includes an error message.

## Tests performed

- Installed 6.6
- Upgraded to 7.0
- (Did not run `xdmod-upgrade`)
- Loaded portal
- Observed error message

## Types of changes

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
+++ b/open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
@@ -412,20 +412,27 @@ class XDUserTest extends \PHPUnit_Framework_TestCase
         $this->assertNotNull($user->getUserID());
     }
 
+    /**
+     * @expectedException Exception
+     * @expectedExceptionMessage User "bilbo" not found
+     */
     public function testGetUserByUserNameInvalid()
     {
         $user = XDUser::getUserByUserName("bilbo");
-        $this->assertNull($user);
-    }
-
-    public function testGetUserByUserNameEmptyString()
-    {
-        $user = XDUser::getUserByUserName("");
-        $this->assertNull($user);
     }
 
     /**
      * @expectedException Exception
+     * @expectedExceptionMessage User "" not found
+     */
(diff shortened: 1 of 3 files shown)


jtpalmer:
Moved exception to the `getUserByUserName` method after discussion with @plessbd .

jpwhite4:
Please can you add a section to the FAQ (either here, or a a different pull request) that explains what to do if you see this message.

plessbd:
Maybe have the error point to the FAQ?

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -424,10 +424,12 @@ public function testGetUserByUserNameEmptyString()
         $this->assertNull($user);
     }
 
+    /**
+     * @expectedException Exception
+     */
     public function testGetUserByUserNameNull()
```
plessbd: should we add a test for `$user = XDUser::getUserByUserName('Ben is not really a user here...');`

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -424,10 +424,12 @@ public function testGetUserByUserNameEmptyString()
         $this->assertNull($user);
     }
 
+    /**
+     * @expectedException Exception
```
plessbd: the previous two tests need to have this as well I believe, as they will also now through an exception.

jtpalmer: I've updated those tests now

PR #266, opened 2017-09-26
JSON encode drill down data

jtpalmer:
## Description

See title.

## Motivation and Context

If any of the labels in `hierarchy.json` (really any dimension labels) contain a single quote (`'`) invalid JavaScript is returned from the `user_interface` `get_charts` controller.  These labels need to be properly encoded.

There are probably other issues with this code and the corresponding code in `html/gui/js/DrillDownMenu.js`. I can imagine that labels containing `-` or `,` characters will cause problems.  Fixing issues with those characters are not addressed in this PR.

## Tests performed

- Installed 7.1
- Edited a `hierarchy.json` label to contain both single and double quote characters
- Viewed chart in Usage tab
- Observed that the chart was displayed
- Inspected response from server to confirm properly encoded drill down data

## Types of changes

## Checklist:


jtpalmer:
Now that I've thought about it a bit more, it's probably best to replace all the string parsing (i.e. splitting on `,` and `-`) with the use of JSON arrays/objects.  I'll see what's necessary to do that.

jpwhite4:
I'm sure you have already thought of thi, but I'd remove the implode() statement and json_encode the php array directly.

classes/DataWarehouse/Query/Jobs/GroupBy.php
```diff
@@ -33,15 +33,20 @@ public function getDrillTargets($statistic_name, $query_classname)
 
 			if($group_by_instance->getAvailableOnDrilldown() !== false && array_search($statistic_name,$permitted_stats) !== false)
 			{
-				$drill_target_group_bys[] = $group_by_name.'-'.$group_by_instance->getLabel();
+				$drill_target_group_bys[] = array($group_by_name, $group_by_instance->getLabel());
```
jtpalmer: I just realized that if `getDrillTargets` changes, then it needs to change in the other modules. If `$group_by_name` is guaranteed to not contain `-` the string could be split in the caller.  Not sure if that's a safe assumption.

html/gui/js/DrillDownMenu.js
```diff
@@ -52,7 +52,7 @@ Ext.extend(CCR.xdmod.ui.DrillDownMenu, Ext.menu.Menu, {
                     items.push(
                         new Ext.menu.Item({
                             scope: this,
-                            drillDown: groupByDescripter[i],
+                            drillDown: gbd.join('-'),
```
jpwhite4: What is this 'drillDown' field used for? Would it not be better to keep this as an array?

jtpalmer:
@jpwhite4 Replaced the `drillDown` string with an array as you suggested.  What do you think about the change to `getDrillTargets` in the `GroupBy` classes?  Should we maintain compatibility or make a multi-repo change?

PR #268, opened 2017-09-26
Acl/user management7.0

ryanrath:
## Description
Changes to provide support for managing Acls via the internal dashboard User Management tab. 

## Motivation and Context
While Acls work alongside the current Role system. It does not allow for the addition of, and subsequent management of, an new Acl via the XDMoD Internal Dashboard. That is now supported.

## Tests performed
Manual Tests were performed. They were as follows: 

**Create a New User:**
  - Login to the Internal Dashboard
  - Click the 'User Management tab.
  - Click the 'Existing Users' sub-tab
  - Click the 'Create & Manage Users' Toolbar button
  - Ensure that the 'New User' tab is selected
  - Fill in the following fields:
    - First Name
    - Last Name
    - E-Mail Address
    - Username
    - Map To ( the selection does not matter )
    - Acls:
      - Select the 'User' acl
      - Select any other Acls that are appropriate
    - Optional:
      - Set a different 'User Type' 
  - Click the 'Create User' button
  - Ensure that the process complete's without error. 
    - **Note:** If you are testing on a limited system, sendmail needs to be setup or the creation process with throw an error. But, the user account will still be setup correctly. Just refresh the page and you should see the newly created User account.
  - If the user account is not visible click the 'Refresh' toolbar button. The user account should now be visible.
  - Double click the table row that represents the newly created User account. 
  - Ensure that the 'Create & Manage Users' modal window has been shown. 
  - Ensure that the user that was clicked is pre-selected on the left side of the window.
  - Ensure that the 'Acl Assignment' portion of the modal is showing the correct selections.
  - Click the 'Actions' Toolbar menu item
  - Select the 'Disable This Account' menu item
  - Ensure that a confirmation pop-up is displayed asking to confirm if you want to disable access for the user.
  - Click 'Yes'
  - Ensure that the operation completes successfully.
  - Ensure that the user entry in the modal windows 'Existing Users' table is now colored red.
  - Click the 'Actions' Toolbar menu item
  - Select the 'Enable This Account' menu item
  - Ensure that a confirmation pop-up is displayed asking to confirm if you want to enable access for the user.
  - Click 'Yes'
  - Ensure that the operation completes successfully.
  - Ensure that the user entry in the modal windows 'Existing Users' tale is now black
  - Add or remove some additional acls
  - Click 'Save Changes'
  - Ensure that the operation completes successfully
  - Close the 'User Management' modal window.
  - Force refresh the browser
  - Select the 'Existing Users' sub-tab.
  - double click the same user
  - Ensure that all of the user's information is as it was pre-refresh

**Modify an existing User:**
  - Login to the Internal Dashboard
  - Click the 'User Management tab.
  - Click the 'Existing Users' sub-tab
  - Double click on an existing user in the table of users.
  - Ensure that the 'User Management' modal window is displayed
  - Ensure that the user that was originally clicked is visible and selected in the 'Existing Users' table.
  - Ensure that the 'User Details' of the selected user are as expected.
  - Add or remove some acls. 
  - Click 'Save Changes' 
  - Ensure that the operation completes successfully
  - Close the 'User Management' modal window
  - double click the modified user
  - Ensure that acls displayed are as expected.
  - Force Refresh the browser.
  - Select the 'Existing Users' sub-tab
  - double click the modified user.
  - Ensure that the 'User Management' modal is displayed.
  - Ensure that the selected user is both visible and selected in the 'Existing Users' table.
  - Ensure that all of the information in the 'User Details' panel is as expected.

## Types of changes

## Checklist:


html/controllers/sab_user/enum_tg_users.php
```diff
@@ -38,13 +40,13 @@
 $user = \XDUser::getUserByID($_SESSION[$user_session_variable]);
 
 if (
-    $user->getActiveRole()->getIdentifier() == ROLE_ID_CAMPUS_CHAMPION
+   $user->hasAcl(ROLE_ID_CAMPUS_CHAMPION)
```
smgallo: I don't see a campus champion role in the `roles.json` file in this PR.

html/controllers/ui_data/summary3.php
```diff
@@ -64,8 +67,13 @@
             throw $e;
         }
     }
-
-    $summaryCharts = $logged_in_user->getMostPrivilegedRole()->getSummaryCharts();
+    $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($logged_in_user);
+    $summaryCharts = array_map(
```
smgallo: This function encodes the summary charts and then loops through them at least twice unencoding the JSON, setting a value, and re-encoding. Why not just encode them at the end?

html/controllers/user_admin/enum_roles.php
```diff
@@ -1,28 +1,39 @@
...
+
    $xda = new XDAdmin();
-			
-   $roles = $xda->enumerateRoles();
-			
+
+   $roles = $xda->enumerateAcls();
+
+   try {
+       $multipleServiceProviders = \xd_utilities\getConfiguration('features', 'multiple_service_providers') === 'on';
```
smgallo: I'd prefer to see statements like this use parens to make it easier to read. E.g., `$value = ( func() === $otherValue);`

html/controllers/user_admin/update_user.php
```diff
@@ -104,8 +100,22 @@
...
 
 // ===========================================
+    if (isset($_POST['acls'])) {
+       $acls = json_decode($_POST['acls'], true);
+       $user_to_update->setAcls(array());
+       foreach ($acls as $aclName => $centers) {
+           $acl = Acls::getAclByName($aclName);
+           $user_to_update->addAcl($acl);
+
+           if (count($centers) > 0) {
```
smgallo: Assuming that $centers is guaranteed to be an array you don't need the `if` statement here.

html/controllers/user_admin/update_user.php
```diff
@@ -104,8 +100,22 @@
...
+               foreach ($centers as $center) {
+                   $user_to_update->addAclOrganization($aclName, $center);
+               }
+           }
+       }
+    }
 
-	if (isset($_POST['roles'])) {
+	/*if (isset($_POST['roles'])) {
```
smgallo: Remove commented out code.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -491,21 +491,15 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
                     CCR.xdmod.ui.userManagementMessage('Please fix any fields marked as invalid.', false);
                     return;
                 }
+                var acls = newUserRoleGrid.getSelectedAcls();
+                var missingMappedUser = cmbUserMapping.getValue().length === 0;
 
-                if (!newUserRoleGrid.areRolesSpecified()) {
+                if (!acls.length > 0) {
```
smgallo: Why not just `if ( acls.length <= 0)` ?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -534,6 +525,13 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
                     return;
                 }
 
+                var populatedAcls = {};
+                // populate centers for any of the selected acls that have them
+                for ( var i = 0; i < acls.length; i++) {
+                    var acl = acls[i];
```
smgallo: Are these acl objects or ids?

smgallo:
@ryanrath This is set to merge into v7.0 but it should be v7.1

configuration/roles.json
```diff
@@ -349,39 +353,59 @@
...
+            "extends": "default",
+            "hierarchies": {
+                "level": 100,
+                "filter_override": false
+            }
         },
         "cd": {
             "display": "Center Director",
-            "type": "feature",
+            "type": "feature_requires_center",
```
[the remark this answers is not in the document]
ryanrath: so the intended purpose is to indicate that these acls require, when the install is using multiple service providers, additional information to function properly. What this translates to is a record being recorded in the user_acl_group_by_parameters table w/ a value of the service providers ( center ) id that should be associated with a user / acl / group by ( provider in this case ). 

But now that I've had a bit to look it over, a better way to indicate this requirement might be to do something like: 
```json
{
  "cd": {
    "display": "Center Director",
    "type": {
      "name": "feature",
      "requires_center": true
    }
  }
}
```

html/controllers/sab_user/enum_tg_users.php
ryanrath: Agreed, this was just a straight replace of the pattern
```php 
$user->getActiveRole()->getIdentifier() === SOME_HARD_CODED_ROLE;
```
with:
```php
$user->hasAcl(SOME_HARD_CODED_ROLE);
```
If the code was there before I didn't want to remove it and possibly break something that utilizes that logic / functionality / data down the line.

html/controllers/ui_data/summary3.php
ryanrath: Yeah, again I just wanted to sub out the deprecated function usage w/ the minimal amount of new to make it work. I didn't want to do any optimizations until I knew everything was still copacetic. But, it can most definitely be simplified. I'll put that in.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: sure

ryanrath:
Merge target changed to 7.1. Working on some disabled menus issues at the moment.

configuration/roles.json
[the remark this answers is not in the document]
smgallo: Or something more generic like this so multiple items could be required. Then the question is, where do the definition of the require values come from.
```
{
  "cd": {
    "display": "Center Director",
    "type": {
      "name": "feature",
      "requires": [ "center" ]
    }
  }
}

File: html/controllers/sab_user/enum_tg_users.php (hunk 1 above)
Comment (smgallo) [reply html/controllers/sab_user/enum_tg_users.php]: ok

File: classes/Models/Services/Acls.php [hunk 9]
```diff
@@ -780,4 +776,86 @@ public static function getDescriptorParamValues(XDUser $user, $aclName, $groupBy
...
+     *                                 constrain the most privileged acl search.
+     *                                 (optional)
+     * @return Acl|null If the user does not have an acl that satisfies the
+     *                  constraints then null will be returned. Else, the acl
+     *                  is returned as an instantiated Acl object.
+     * @throws Exception if the user does not have a user id.
+     */
+    public static function getMostPrivilegedAcl(XDUser $user, $moduleName = DEFAULT_MODULE_NAME, $aclHierarchyName = 'acl_hierarchy')
+    {
+        if (null == $user->getUserID()) {
```
smgallo: Use === here because null == 0 is true

configuration/roles.json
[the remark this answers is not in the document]
ryanrath: Ooooo, I like that. Ok, I'll work on updating the code / db to work with the new config layout.

bin/acl-config
```diff
@@ -864,43 +865,40 @@ SQL;
...
+
+    $modified = $statement->execute($params);
+    $count = $modified === true ? $statement->rowCount() : 0;
+    $inserted += $count;
+    if ($count === 0) {
+        $log->info("[FAIL] Did not Process Acl Hierarchy: $id");
+    } elseif ($count === 1) {
+        $log->info("[SUCCESS] Processed Acl Hierarchy: $id");
+    } else {
+        $log->info("[FAIL] Inserted more than one Acl Hierarchy Record for: $id");
```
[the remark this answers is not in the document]
smgallo: This should still be wrapped in a transaction.

classes/XDUser.php
```diff
@@ -3194,4 +3177,68 @@ public static function getUserByUserName($username)
...
+     *                               relationship created for it with the
+     *                               provided organization.
+     * @param string $organizationId the name of the organization
+     * @throws Exception if this user has not been saved yet
+     * @throws Exception if the provided acl cannot be found.
+     */
+    public function addAclOrganization($aclName, $organizationId)
+    {
+        if (empty($this->_id)) {
+            throw new \Exception("This user must be saved prior to calling this function.");
```
smgallo: Change "calling this function" to "calling " . `__FUNCTION__` for clarity.

classes/XDUser.php
```diff
@@ -3194,4 +3177,55 @@ public static function getUserByUserName($username)
...
+    public function addAclOrganization($aclName, $organization)
+    {
+        if (empty($this->_id)) {
+            throw new \Exception("This user must be saved prior to calling this function.");
+        }
+
+        $acl = Acls::getAclByName($aclName);
+
+        if ( null == $acl) {
+            throw new Exception("Unable to retrieve acl for: $aclName");
```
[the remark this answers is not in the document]
smgallo: There are still several places where `\Exception` is being used and also `Exception`. These should be consistent. If you want to do this in another PR that's fine.

bin/acl-config
```diff
@@ -243,7 +244,7 @@ SQL;
             $display = null;
 
             if (null === $hierarchy['name']) {
-                throw new Exception("Malformed hierarchy entry for module $moduleName. No name property.");
+                throw new \Exception("Malformed hierarchy entry for module $moduleName. No name property.");
```
chakrabortyr: This may should be just Exception without the backslash as per [this pr](https://github.com/ubccr/xdmod/pull/299).

classes/Models/Services/Acls.php
```diff
@@ -442,15 +445,8 @@ public static function getDisabledMenus(XDUser $user, array $realms)
...
-            $name = $row['name'];
-            if ($name != $previousName) {
-                $previousName = $name;
-            }
-            if (!isset($results[$name])) {
-                $results[$name] = array();
-            }
             if ($row['id'] != null) {
-                $results[$name] [] = array(
+                $results[] = array(
```
chakrabortyr: What's the behavior when you remove the index from the initial array? Does this always produce the desired result?

smgallo:
Looks like there are some still some issues with the tests.

bin/acl-config
```diff
@@ -914,6 +941,16 @@ SQL;
             $log->info("[FAIL] Inserted more than one Acl Hierarchy Record for: $id");
         }
     }
+    $committed = $db->commit();
+    if ($committed === false) {
+        $log->warning("[FAIL] Attempt to commit hierarchies for Acl [$acl] failed.");
+        $rolledBack = $db->rollBack();
+
+        // We throw an Exception so that we do not leave the db in an inconsistent state.
```
jpwhite4: This comment is at odds with the error message - if throwing the exception stops the database being left in an incosistent state then why does the exception claim that the database may be inconsistent?

jpwhite4:
Please can you add some tests for the new functionality. I suggest using the integration test harness to call the new/modified internal dashboard API calls to add/remove/modify some user accounts.

jpwhite4:
The "Public User" also incorrectly appears in the New User dialog box too.

Also I managed to get it to display this message:

image

but it still create an entry in the database for the user anyway. This probably needs to not add the rows to the database in this case.

Even better than that the user is undeletable now.

jpwhite4:
Also the Public User shows up in the Existing Users grid - this too should be hidden

classes/Models/Services/Acls.php
ryanrath: The behavior is that instead of having an associative array indexed by acl.name we now have a non-associative array that contains an record for each row that has a non-null id. I do need to update the comments as they no longer make sense, but this was changed so that the value of the `CCR.xdmod.ui.disabledMenus` property matched what it was before.

chakrabortyr:
User tests are still failing

ryanrath:
@jpwhite4 re: Public User showing in the existing users grid. I can certainly make that happen. For what it's worth the 'Public User' is visible in prod when you go to 'Create / Manage Users' and select 'Internal Users'. Would this be considered a bug? Or is this behavior desired? I just want to make sure before I start making changes.

ryanrath:
@jpwhite4 for the user creation exception you encountered, did you provide a username / mapped user and were there any acls assigned ( all I can see is that center director & center staff were not assigned )?

ryanrath:
Please note that there are some tests that were commented out during development. I'll be committing them shortly.

configuration/roles.json
```diff
@@ -361,15 +361,15 @@
         },
         "cd": {
             "display": "Center Director",
-            "type": "feature",
+            "extends": "default",
```
jpwhite4: Why did you change this?

configuration/roles.json
```diff
@@ -227,7 +227,7 @@
                                 },
                                 "group_by": "pi",
                                 "has_std_err": true,
-                                "id": 0.00000000000004,
```
jpwhite4: Why did this change?

smgallo:
In XDUser.php there are several exceptions being thrown as `Exception` and others as '\Exception`. Since this is not a namespaced file it doesn't matter but it should be consistent.

ryanrath:
@jpwhite4 I've removed `Public User` / the `pub` role from the internal_dashboard -> User Management tab. I've also updated the ControllerTests so that they take this removal into account.

PR #271, opened 2017-09-27
Add PDF support to XDMoD

jpwhite4:
## Types of changes

## Checklist:

### DIFF
--- a/html/gui/js/ExportPanel.js
+++ b/html/gui/js/ExportPanel.js
@@ -68,6 +68,8 @@ Ext.extend(CCR.xdmod.ui.ExportPanel, Ext.Panel, {
         this.imageExportAllowed = allow;
     },
     setupDisplay: function(format_type) {
+        // Define which fields are to be shown for each export type. (Note that the xml
+        // and csv exports have no fields shown).
         var fieldSettings = {
             width_inches: ['pdf'],
             height_inches: ['pdf'],
--- a/html/gui/js/ExportPanel.js
+++ b/html/gui/js/ExportPanel.js
@@ -29,7 +29,8 @@ Ext.apply(CCR.xdmod.ui.ExportPanel, {
         ['png', 'PNG - Portable Network Graphics'],
         ['svg', 'SVG - Scalable Vector Graphics'],
         ['csv', 'CSV - Comma Separated Values'],
-        ['xml', 'XML - Extensible Markup Language']
+        ['xml', 'XML - Extensible Markup Language'],
+        ['pdf', 'PDF - Portable Document Format']
     ],
     format_types_noimg: [
         ['csv', 'CSV - Comma Separated Values'],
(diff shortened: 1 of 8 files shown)


html/gui/js/ExportPanel.js
```diff
@@ -67,17 +68,19 @@ Ext.extend(CCR.xdmod.ui.ExportPanel, Ext.Panel, {
...
-            case 'svg':
-                this.showTitleCheckbox.show();
-                this.templateTypeCombo.show();
-                break;
-            case 'xml':
-            case 'csv':
-                this.showTitleCheckbox.hide();
-                this.templateTypeCombo.hide();
-                break;
+        var fieldSettings = {
```
smgallo: A brief comment as to what this data structure is for would help to clarify the code.

html/gui/js/ExportPanel.js
```diff
@@ -67,17 +68,19 @@ Ext.extend(CCR.xdmod.ui.ExportPanel, Ext.Panel, {
...
+            width_inches: ['pdf'],
+            height_inches: ['pdf'],
+            font_pt: ['pdf'],
+            show_title: ['pdf', 'png', 'svg'],
+            image_size: ['png', 'svg']
+        };
+
+        var field;
+        for (field in fieldSettings) {
+            if (fieldSettings.hasOwnProperty(field)) {
```
smgallo: If you create the `fieldSettings` object and iterate over it's data members is the `if` statement necessary?

libraries/charting.php
```diff
@@ -207,9 +207,94 @@ function exportHighchart($chartConfig, $width, $height, $scale, $format, $global
...
+    $descriptor_spec = array(
+        0 => array('pipe', 'r'),
+        1 => array('pipe', 'w'),
+        2 => array('pipe', 'w'),
+    );
+
+    $process = proc_open($command, $descriptor_spec, $pipes);
+
+    if (!is_resource($process)) {
+        throw new \Exception('Unable to create gs subprocess');
```
smgallo: Could you use `error_get_last()` to grab a useful error message? Or include the command that it was trying to execute.

libraries/charting.php
```diff
@@ -207,9 +207,94 @@ function exportHighchart($chartConfig, $width, $height, $scale, $format, $global
...
+    $out = stream_get_contents($pipes[1]);
+    $err = stream_get_contents($pipes[2]);
+
+    fclose($pipes[1]);
+    fclose($pipes[2]);
+
+    $return_value = proc_close($process);
+
+    if ($return_value != 0) {
+        $msg = "gs returned $return_value, stdout: $out stderr: $err";
```
smgallo: Channeling @plessbd do we need to create the variable or can we just pass the string to Exception?

html/gui/js/ExportPanel.js
plessbd: it is because of how javascript does a for in and will get prototype properties and you dont actually want to get them

and it think eslint will complain if it isnt used...

jpwhite4: Yes this is the way you have to iterate over an object in ES5. eslint will complain if you don't do this too.

PR #274, opened 2017-09-28
Improve robustness of metrics explorer GUI tests.

jpwhite4:
This click was missing a check for the mask


open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/loginPage.page.js
```diff
@@ -49,6 +49,7 @@ class LoginPage {
             it('Display Logged out State', function clickLogout() {
                 browser.waitForInvisible('.ext-el-mask-msg');
                 $('a[href*=actionLogin]').waitForVisible();
+                $('#main_tab_panel__about_xdmod').waitForVisible();
```
plessbd: Why are we waiting for the tab to be visible on this one?

jpwhite4: The reason is that we want the test to leave the browser in a stable state

PR #277, opened 2017-09-29
Updates shippable to image 7.0.1

plessbd:
This seems to work with my changes to the docker file to allow the base shippable image to be version 7.0.1
even though https://app.shippable.com/github/plessbd/xdmod/runs/9/summary/console failed, it was because of the UI tests which seem to have a few more bugs to work out and not because of the shippable updates


jtpalmer:
The 7.0 sync caused some conflicts.  Also, you might as well update the version numbers in `portal_settings.ini` and `package.json` as part of these changes.

PR #280, opened 2017-09-29
Add Export support to JobViewer

jpwhite4:
## Description
The chart export and printing in the job viewer is now updated to use the same Export and Print toolbar buttons as the rest of XDMoD. The main improvement is that the image size, font size and title presence are now configurable. Secondary improvement is that export UI is now consistent with the rest of XDMoD.

## Motivation and Context
Add configurable export to job viewer; Make Job viewer UI consistent with XDMoD.

## Tests performed
New tests will have been added in the SUPReMM module.

## Types of changes
- New feature (non-breaking change which adds functionality)


jpwhite4:
By the way I changed the export dialog so you can configure which export options appear for each portal module. This could have been done as a seperate pull request, but would not have been able to be tested without the changes here anyway.

PR #289, opened 2017-10-10
Update dependencies and support documentation

jtpalmer:
## Description

Updates the list of dependencies along with the RPM spec file and integration tests.  Also updates the support page to explain what versions we will support.

## Motivation and Context

If we aren't going to support it, it shouldn't be in the documentation.

## Tests performed

Ran automated tests.

## Types of changes

## Checklist:

### DIFF
--- a/docs/software-requirements.md
+++ b/docs/software-requirements.md
@@ -6,8 +6,8 @@ Open XDMoD requires the following software:
 
 - [Apache][]
     - [mod_rewrite][]
-- [MySQL][] 5.1 or 5.5
-- [PHP][] 5.3+
+- [MariaDB][]/[MySQL][] 5.5.3+
+- [PHP][] 5.4+
     - [PDO][]
     - [MySQL PDO Driver][pdo-mysql]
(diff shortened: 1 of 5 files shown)


jtpalmer:
@smgallo , @chakrabortyr I tried to incorporate everything from the email discussion, but please let me know if I missed anything.  The support page could probably use more work, any suggestions are always welcome.

docs/software-requirements.md
```diff
@@ -6,8 +6,8 @@ Open XDMoD requires the following software:
 
 - [Apache][]
```
smgallo: Since we list versions of several of the other software packages should we do the same for Apache or is any version of Apache2 OK?

jtpalmer: We've used Apache 2.2 and 2.4, not sure about 2.0.  It looks like [2.2 isn't maintained anymore](http://httpd.apache.org/#apache-httpd-2234-released-end-of-life-2017-07-11), so I'll change it to specify 2.4.

PR #290, opened 2017-10-10
Ignore duplicate LSF host names

jtpalmer:
## Description

Ignores duplicate hosts found in LSF accounting log files.

## Motivation and Context

Duplicate host names in the list of hosts for a given job break the job hosts ingestor.  Likewise, duplicate hosts would result in incorrect node counts.

This was discovered by an LSF user attempting to use SUPReMM which requires the job host data.  The cause is a previously unknown LSF accounting log format.

## Tests performed

Added unit test.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


jtpalmer:
The difference in LSF formats appears to be due to the `SHORT_EVENTFILE` setting.  Previously, it was assumed that this was set.

PR #294, opened 2017-10-11
Add polyfill for Array.prototype.includes() which is missing from IE.

jpwhite4:
## Description

Add polyfill for Array.prototype.includes() which is missing from IE.

## Motivation and Context
We started using .includes() in #271, turns out that it is not in IE. Rather than removing the calls to the function this just adds the capability to browsers that don't already have it.

## Tests performed
Manually checked that export works in IE. Ran the automated tests in SauceLabs.

## Types of changes


plessbd:
Do we actually need any of the other polyfills anymore?  I would say we should not be adding in things we dont need or use.

PR #296, opened 2017-10-11
updates to test conf

plessbd:
Get rid of having two places for .secrets file for tests
Update automated ui tests to use new file
update integration tests to use new file

update wdio.conf.js to be usable for both chrome headless and sauce labs


jpwhite4:
Given that you are changing all of the code that calls login, then how about taking the oppertunity to change the login function to access the credentials directly.

PR #297, opened 2017-10-12
Better Federated/Local UI Support

chakrabortyr:
## Description
Old Login UI (metrics-dev):
screen shot 2017-10-11 at 8 49 17 pm

New UI (final logo subject to change):
screenshot from 2017-10-30 13-01-38

## Motivation and Context
A concise look that fits in better with the rest of the GUI. More clearly associates the Forgot Password/Sign Up stuff with your local account rather than federated account.

On the federated side, addresses excess spacing issues by making the federated button and the window adapt itself to the login. We've also merged XSEDE/Non-XSEDE based logins so we have either a federated or a non-federated login. It's now also possible (and looks) better when you have either federated or local enabled.

## Tests performed
Verified in Travis + Shippable (ran continuous integration tests).

## Types of changes

## Checklist:


html/gui/js/CCR.js
```diff
@@ -929,54 +928,241 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+        items: [{
+            xtype: 'tbtext',
+            html: '<a href="' + CCR.xdmod.federationLoginLink.url + '"><img src="' + CCR.xdmod.federationLoginLink.icon + '" alt="Login here."></img></a>',
+            id: 'federatedLoginLink'
+        }]
+    }) : null;
+
+    var localLoginFrm = new Ext.form.FormPanel({
+        id: 'local_login_form',
+        title: 'Sign in with a local Account:',
```
smgallo: Lets change this to "Sign in with a local XDMoD account"

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -112,7 +112,7 @@ public function getXdmodAccount()
         return false;
     }
 
-    public function getLoginLink()
+    public function getLoginLink($returnTo = null)
```
smgallo: There are no comment blocks for any of the methods in this class. We should have a docblock for each method when writing new code as well as the class properties. For example, what is `$_as`???

html/gui/js/CCR.js
```diff
@@ -929,54 +928,241 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
-    // Adjsut window size depending on if the XSEDE login is enabled.
-    if (CCR.xdmod.features.xsede || CCR.xdmod.isFederationConfigured) {
-        width = 540;
-        promptHeight = 486;
-        iframeHeight = 500;
-    }
     //reset referer
     XDMoD.referer = document.location.hash;
 
+    // TODO: Actually get this from config
```
jpwhite4: Please do not add TODOs to the code. If you actually want to do the TODO then do it

html/gui/js/CCR.js
```diff
@@ -929,54 +928,241 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+
+                        switch (json.status) {
+                            case 'invalid_email_address':
+                                presentLoginResponse('A valid e-mail address must be specified.', false, 'reset_response');
+                                break;
+                            case 'no_user_mapping':
+                                presentLoginResponse('No XDMoD user could be associated with this e-mail address.', false, 'reset_response');
+                                break;
+                            case 'multiple_accounts_mapped':
+                                presentLoginResponse('Multiple XDMoD accounts are associated with this e-mail address.', false, 'reset_response');
```
smgallo: They should also contact the administrator.

html/index.php
```diff
@@ -311,7 +311,8 @@ function isReferrer($referrer)
                   // This will catch when a configuration directory does not exist if it is set in the environment level
                }
                if ($auth && $auth->isSamlConfigured()) {
-                  print "CCR.xdmod.isFederationConfigured = true;";
+                    print "CCR.xdmod.isFederationConfigured = true;";
```
smgallo: Do you want to add newlines into the printed code?

classes/Authentication/SAML/XDSamlAuthentication.php
chakrabortyr: Agreed. 

To quickly answer $_as is the variable we use to store whichever SimpleSAML auth source we'll be using. This code handles all federated auth. I'll go through and document the file.

html/gui/js/CCR.js
chakrabortyr: Np, removed.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -112,7 +121,14 @@ public function getXdmodAccount()
         return false;
```
plessbd: we should probably turn this into an exception

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -132,14 +148,18 @@ public function getLoginLink()
                 );
             }
             return array(
-                    'url' => $this->_as->getLoginUrl(),
+                    'url' => $this->_as->getLoginUrl($returnTo),
                     'organization' => $orgDisplay,
                     'icon' => $icon
                 );
         } else {
             return false;
```
plessbd: probably make this throw an exception rather than mixed return

plessbd:
I want to see PR's in xdmod-xsede to to add globus auth to composer as well as a PR for the configuration changes required to make XSEDE work.

Also, some screenshots of what the page looks like without federated login enabled

On the XSEDE PR we should have screenshots of what it will look like for the XSEDE Login as well.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -7,9 +7,8 @@
 
 class XDSamlAuthentication
 {
-    protected $_as = null;
-
-    protected $_sources = null;
+    protected $_as = null; // The selected auth source
```
smgallo: New code should include docblocks with variable types. I.e.,
```
/**
 * The selected auth source
 * @var \SAML\MyAuthSource
 */
```

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -45,6 +49,11 @@ public function isSamlConfigured()
         return $this->_isConfigured;
     }
 
+    /**
+     * Attempts to find a valid XDMoD user associated with the attributes we receive from SAML
+     * 
+     * @return mixed a valid XDMoD user if we have one, false otherwise
```
smgallo: Capitalize the first word of the sentence so it's easier to see that its the start of the description "A valid XDMoD user..."

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -112,6 +121,13 @@ public function getXdmodAccount()
         return false;
     }
 
+    /**
+     * Retrieves the login url we want to use with this authentication provider. 
+     * 
+     * @param returnTo the URI to redirect to after auth, if any. null by default.
```
smgallo: Please add the type of each parameter
```
* @param $returnTo string The URI to...
```

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -140,6 +156,10 @@ public function getLoginLink($returnTo = null)
             return false;
         }
     }
+
+    /**
+     * Sends an email notifying XDMoD admin of new account.
```
smgallo: Needs parameters

html/gui/general/login.php
```diff
@@ -16,17 +16,13 @@
...
             $formal_name = $xdmodUser->getFormalName();
         } else {
-            print($xdmodUser);
-            exit();
-            $samlError = $xdmodUser ? $xdmodUser : "INACTIVE";
+            $samlError = "INACTIVE";
         }
     } else {
-        $samlError = $xdmodUser ? $xdmodUser : null;
+        $samlError = null;
```
jpwhite4: Why is samlError set to null here? By code inspection it will already have the value false when this line of code is executed.

html/gui/general/login.php
```diff
@@ -16,17 +16,13 @@
...
     }
 }
 // Used for Federated login or samlErrors
-if (!empty($formal_name) || $samlError) {
-
```
jpwhite4: As per our discussion, the javascript on line 86 will not work because this new code is not in an iframe any more.

plessbd:
The text boxes have placeholder and labels, this seems off to me.
We have Sign Into XDMoD followed by Sign in with "ACCOUNT"  seems redundant to me

plessbd:
Need to be able to press enter after entering my password to submit the form

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -7,11 +7,32 @@
 
 class XDSamlAuthentication
 {
-    protected $_as = null;
+    /**
+     * The selected auth source
+     *
+     * @var \SimpleSAML_Auth_Simple
+     */
+    protected $_as = null; // The selected auth source
```
plessbd: Comment exists twice

plessbd:
Since we are adding the globus auth module we should add documentation on how to configure it to the docs.

plessbd:
Do you have updated screenshots to go with the changes?  Will be helpful when the user manual has to be updated

html/gui/js/CCR.js
```diff
@@ -268,7 +268,6 @@ XDMoD.GlobalToolbar.Help = function (tabPanel) {
             }
         });
     }
-
```
jpwhite4: Please don't edit this line - it is unrelated to the code changes and you will probably be penalised for it by the style checkers.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -35,16 +56,22 @@ public function __construct()
...
+     * @return boolean true if we have 1 or more auth sources. false otherwise
+     */
     public function isSamlConfigured()
     {
-        //TODO: Make this more robust by taking into account if the IDP MetaData does not exist.
-        // look at getLoginLink for details on getting idp info
-        if (is_null($this->_isConfigured)) {
-            $this->_isConfigured = count($this->_sources) > 0 ? true : false;
-        }
+        $this->_isConfigured = count($this->_sources) > 0 ? true : false;
```
jpwhite4: What is the purpose of the _isConfigured variable? It only appears to be set if you call isSamlConfigured() and is not used at all elsewhere. 

Surely the entire isSamlConfigured function could be simplified to:
```
return count($this->_sources) > 0;
```
and the superfluous member variable deleted.

html/gui/js/CCR.js
```diff
@@ -929,54 +928,250 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+        name: 'password',
+        inputType: 'password',
+        fieldLabel: 'Password'
+    });
+
+    var loginItems = [];
+    var accountName = CCR.xdmod.federationLoginLink && CCR.xdmod.federationLoginLink.organization.en !== 'Federation' ? CCR.xdmod.federationLoginLink.organization.en : CCR.xdmod.org_name;
+
+    var signInWithLocalAccount = function () {
+        if (txtLoginUsername.getValue().length === 0) {
```
jpwhite4: Just as a note for future development, the ExtJS form class has built in support for input validation, which saves manually writing the input validation.

html/gui/js/CCR.js
```diff
@@ -929,54 +928,250 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+                if (decodedResponse) {
+                    XDMoD.TrackEvent('Login Window', 'Successful login', txtLoginUsername.getValue());
+                    XDMoD.REST.token = data.results.token;
+                    XDMoD.TrackEvent('Login Window', 'Login from public session', '(Token: ' + XDMoD.REST.token + ')', true);
+                    presentLoginResponse('Welcome, ' + Ext.util.Format.htmlEncode(data.results.name) + '.', true, 'login_response');
+                    parent.location.href = '../../index.php' + parent.XDMoD.referer;
+                    parent.location.hash = parent.XDMoD.referer;
+                    parent.location.reload();
+                } else {
+                    XDMoD.TrackEvent('Login Window', 'Successful login', txtLoginUsername.getValue());
```
jpwhite4: We should not be sending a "Successful login" event if there was an error logging in.

html/gui/js/CCR.js
```diff
@@ -929,54 +928,250 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+                var decodedResponse;
+
+                if (success) {
+                    decodedResponse = CCR.checkDecodedJSONResponseSuccess(data);
+                }
+
+                if (decodedResponse) {
+                    XDMoD.TrackEvent('Login Window', 'Successful login', txtLoginUsername.getValue());
+                    XDMoD.REST.token = data.results.token;
+                    XDMoD.TrackEvent('Login Window', 'Login from public session', '(Token: ' + XDMoD.REST.token + ')', true);
```
jpwhite4: We should not be sending the private authentication token to google.

html/gui/general/login.php
```diff
@@ -76,15 +72,12 @@ function loadPortal() {
...
     ?>
-      <body class="error_message">
+      <body class="error_message" onload="loadPortal()">
         <p>
             <?php
             echo $message;
             ?>
         </p>
-        <p>
-          <a href="javascript:void(0);" onclick="parent.presentContactFormViaLoginError()">Click Here</a> to contact an administrator.
```
jpwhite4: Why is this feature being removed? Surely we still want to provide a UI to allow people to contact an admin if login fails?

html/gui/js/CCR.js
```diff
@@ -929,54 +928,250 @@ CCR.xdmod.ui.FadeInWindow = Ext.extend(Ext.Window, { //experimental
...
+            return;
+        }
+
+        var restArgs = {
+            username: txtLoginUsername.getValue(),
+            password: txtLoginPassword.getValue()
+        };
+
+        Ext.Ajax.request({
+            url: '/rest/v0.1/auth/login',
```
jpwhite4: Also the ExtJS form class already knows how to submit forms, you don't need to write your own Ajax handling stuff.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -112,7 +139,15 @@ public function getXdmodAccount()
...
-    public function getLoginLink()
+    /**
+     * Retrieves the login url we want to use with this authentication provider.
+     *
+     * @param string $returnTo the URI to redirect to after auth. default is null.
+     *
+     * @return mixed An array containing a login link + redirect, the name of the organization (eg. Twitter),
+     * and an icon (eg. A logo with the Twitter icon + 'Sign in with Twitter' ). false if none found.
+     */
+    public function getLoginLink($returnTo = null)
```
jpwhite4: Why does the returnTo have a default value? It is never called in the code with the default. What does it mean if there is a default? how have you tested that it works with the default if it is never called?

composer.json
```diff
@@ -23,7 +23,8 @@
         "apache/commons-logging": "^1.0.4",
         "egulias/email-validator": "^1.2",
         "taq/pdooci": "^1.0",
-        "phpmailer/phpmailer": "^5.2"
+        "phpmailer/phpmailer": "^5.2",
+        "ubccr/simplesamlphp-module-authglobus": "dev-master"
```
jpwhite4: I would expect to see the composer.lock file updated too. How come the lock file changes do not appear in the pull request?

html/gui/general/login.php
jpwhite4: Also it looks like the current code flashes up the error message for only 2 seconds before loading the portal again (which presumably will show the public user page). This is a quite a big change to the previous behaviour -   I would like it to not change, or if you do change it, to provide a reasonable justification.

classes/Authentication/SAML/XDSamlAuthentication.php
chakrabortyr: You're right that we don't have explicit calls where returnTo is null, however it's there to keep in line with SimpleSAMLphp functionality. When the returnTo parameter is null, SimpleSAMLphp returns to the current page. This is used when SimpleSAMLphp isn't doing a redirect. Makes more sense if you look at the behavior here: https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Auth/Simple.php#L317

composer.json
chakrabortyr: Not sure, think my branch might be in a weird state, I'll manually add it.

html/gui/general/login.php
chakrabortyr: The ability to contact the administrator is still extant, it just no longer opens the form up for you by default. This is in part because the parent has changed, but mostly because other prompts to contact your system administrator do not open the form for you by default. 

I think merely telling them something akin to 'Your account is disabled, please contact x' should suffice without opening up the form if that's what we do elsewhere.

html/gui/js/CCR.js
smgallo: If memory serves, that was being sent to be able to track what a user did during a session. I agree that it's a bad choice but think that we should leave it as is until we decide to revamp our analytics (and it was already there).

html/gui/general/login.php
jpwhite4: But the message is only flashed on the screen for 2 seconds. Most folks are not going to be able to parse and remember the error message and previously we DID have a link that they could click on. Now we don't.

smgallo:
A couple of general comments:
- The Globus button on your 9011 port does not look like the image above.
- Since we tend to have our action or "OK" buttons on the left and Cancel buttons on the right on most of our login/contact/profile dialogs, lets swap the position of the "Sign In" and "Forgot password/signup"  links.
- You'll need to change the text on the Sign Up dialog to "Users with an XSEDE User Portal account do not need to request an account and can log in via Globus. Local XDMoD accounts are created only for users affiliated with XSEDE, except with the express permission of XSEDE or NSF management."

html/gui/css/LoginPrompt.css
```diff
@@ -1,48 +1,80 @@
...
+ #federated_login_form form, #local_login_form form, #forgot_password_form form {
+	padding: 15px;
+ }
+ 
+#federatedLoginLink img {
+	 border: 1.33px; 
+	 border-style: ridge; 
+	 border-radius: 4px; 
+	 margin: 2.5px; 
+	 box-shadow: 0.25px;
```
plessbd: Invalid value for box shadow.
https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow?v=b

plessbd:
Please rebase and update tests for federated login

html/gui/js/Viewer.js
```diff
@@ -444,6 +444,7 @@ Ext.extend(CCR.xdmod.ui.Viewer, Ext.Viewport, {
 
         // The login dialog is presented if the user is not logged in
         // and the location requests an unavailable tab.
+        // eslint-disable-next-line indent
```
plessbd: why are we disabling indent instead of fixing it?

PR #300, opened 2017-10-13
Update TIme Aggregation Unit

plessbd:
## Description
Fix issue and update for later code changes to work.

## Motivation and Context

In updating code to work with cloud metrics found an issue where the get min and max date range would actually return a day previous to the allowed range.  This fixes that error and sets up for the new period tables by bounding the range to be bound to the min and max job time.

Also combined into one database query instead of a possible 4 from before.

## Tests performed

Ran old code and got duration returned which is the only places where getMinMax was called and compared to the new way.  Added tests to make sure this stays the same in the future.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/TimeAggregationUnit.php
+++ b/classes/DataWarehouse/Query/TimeAggregationUnit.php
@@ -40,77 +40,62 @@ abstract class TimeAggregationUnit
      */
     abstract public function getMaxPeriodPerYear();
 
-
     abstract public function getTimeLabel($timestamp);
-    /**
-     * getMinDateId
-     *
-     * @return int the minimum id from the corresponding time period table based on the
-     *         value of start_date. Returns 1, if start_date is less than all possible
-     *         values in the table.
-     */
-    public function getMinDateId($start_date)
-    {
-        $q = "select min(id) as id from ".$this->getUnitName()."s
-        where '$start_date'
-            between
-            ".$this->getUnitName()."_start
-            and ".$this->getUnitName()."_end";
-
-        $startDateResult = \DataWarehouse::connect()->query($q);
-
-        $start_date_id = 1;
-        if(count($startDateResult) > 0)
-        {
-            $start_date_id = $startDateResult[0]['id'];
-            if(!isset($start_date_id) || $start_date_id == '' )
-            {
-                $q = "select case when '$start_date' < min(".$this->getUnitName()."_start) then 1 else ".self::$_max_date_id." end as id  from ".$this->getUnitName()."s";
-                $startDateResult = \DataWarehouse::connect()->query($q);
-                if(count($startDateResult) > 0)
-                {
-                    $start_date_id = $startDateResult[0]['id'];
-                }
-            }
-        }
-
-
-        return $start_date_id;
-    } //getMinDateId
 
     /**
+     * geDateRangeIds
+     * Given the following possible combinations get the possible range of
+     * date ids that are possible forcing the bounds to be min and max job date,
+     * if less than or greater than jobmin or jobmax -1
+     * -1 is used as to force later functions to return the proper values.
      *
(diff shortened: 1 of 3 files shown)


classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -40,77 +40,62 @@ public function getUnitName()
...
+            SELECT
+                MIN(p.id) as minPeriodId,
+                MAX(p.id) as maxPeriodId
+            FROM
+                ' . $this->getUnitName() . 's p
+            JOIN
+                (
+                SELECT
+                    CASE
+                      WHEN "' . $start . '" > max_job_date THEN NULL
```
jpwhite4: Please use PDO params for the start and end to mitigate any possible sql injection

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -19,133 +18,124 @@ protected function __construct($unit_name)
...
     private static $_initialized = false;
 
-    /*
-    * This is the max date id that will be used if end_date is out of range and greater than the last
-    * date id avaiable in the data table.
-    */
+    /**
+     * This is the max date id that will be used if end_date is out of range and greater than the last
+     * date id avaiable in the data table.
+     */
```
jpwhite4: This is dead code now.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -19,133 +18,118 @@ protected function __construct($unit_name)
...
+     *                   minjob---------------------- maxjob
+     * (-1) start----end
+     * (min-end)   start-----------end
+     * (start-end)               start-----------end
+     * (start-max)                           start-----------end
+     * (-1)                                                      start----end
+     *
+     * @return array the miniumum and maximum date ids or -1, -1 if out of bounds
+     */
+    public function geDateRangeIds($start, $end){
```
smgallo: Should this be `getDateRangeIds`?

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -19,133 +18,118 @@ protected function __construct($unit_name)
...
-                {
-                    $end_date_id = $endDateResult[0]['id'];
-                }
-            }
-        }
-        return $end_date_id;
-    } //getMaxDateId
+    /**
+     * geDateRangeIds
+     * Given the following possible combinations get the possible range of
```
smgallo: How about this description?
Given a date range (start, end) return a new date range that has been normalized so that it includes only dates that overlap the time for which data is available. If the desired date range falls fully outside of the dates for which data is available (either before or after) this is considered an error.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -19,133 +18,118 @@ protected function __construct($unit_name)
...
+     * if less than or greater than jobmin or jobmax -1
+     * -1 is used as to force later functions to return the proper values.
+     *
+     *                   minjob---------------------- maxjob
+     * (-1) start----end
+     * (min-end)   start-----------end
+     * (start-end)               start-----------end
+     * (start-max)                           start-----------end
+     * (-1)                                                      start----end
+     *
```
smgallo: Docblock needs `@param $start string ...` entries.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -19,133 +18,118 @@ protected function __construct($unit_name)
...
+     * @return array the miniumum and maximum date ids or -1, -1 if out of bounds
```
smgallo: Capitalize "The minimum..."

PR #305, opened 2017-10-19
Make null config value an empty array

chakrabortyr:
## Description
Fixes an XDMoD crash where a key is present but no value is assigned in Config.

## Motivation and Context
Improper Configuration shouldn't impede the program as a whole; rather it should be apparent from missing functionality.

## Tests performed
Surely.

## Types of changes

## Checklist:

### DIFF
--- a/classes/Xdmod/Config.php
+++ b/classes/Xdmod/Config.php
@@ -310,11 +310,10 @@ class Config implements ArrayAccess
                 // If the key starts with a "+", merge the values.
 
                 $mainKey   = substr($key, 1);
+                $mainValue = array();
+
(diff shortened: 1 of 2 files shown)


classes/Xdmod/Config.php
```diff
@@ -314,7 +314,7 @@ protected function mergeData(array $data, array $partialData)
                     $mainValue = $data[$mainKey];
                 }
                 else {
-                    $mainValue = null;
+                    $mainValue = array();
```
smgallo: Thinking ahead to static analysis please set this prior to the `if()` block so it is properly scoped.

PR #307, opened 2017-10-23
Acl/config update

ryanrath:
## Description
- Updated the way in which Config.php constructs an internal data representation
  of data. Specifically, as viewed in the context of which module added which pieces of data. This can be
  used with the 'filterByMetaData' function to retrieve a module centric view of
  the required config section ( honoring the 'extends' directives and other
  modules 'inheriting' from xdmod ).
- Updated acl-config to work with this new method of retrieving module specific
  data.

## Motivation and Context
- Simplify the way in which we retrieve module specific data from the Config.php class.

## Tests performed
Manual Tests were performed using the following docker containers:
- tas-tools-ext-01.ccr.xdmod.org/centos7-xdmod6.6.0:version6
  - Installed Module Groups:
     - Plain Open XDMoD:
        - OpenXDMoD
      - OpenXDMoD w/ additional modules: 
        - OpenXDMoD
        - App Kernels
        - SUPREMM

- tas-tools-ext-01.ccr.xdmod.org/centos7-xdmod-supremm6.6.0:version:version3
    - Installed Module Groups:
        - XSEDE XDMoD:
            - OpenXDMoD
            - XSEDE
            - App Kernels
            - SUPREMM

**Test 1**
- Navigate to XDMoD running on container
- Ensure that you see three tabs: Summary, Usage, About
- Click the `Usage` tab
- Ensure that the tab loads successfully
- Ensure that the realm groupings in the left side tree contains all of the realms that are installed that public should have access to.
- Log In as a user that has the acl `usr`
- Ensure that the following tabs are present
  - `Metric Explorer`
  - `Allocations`
  - `Report Generator`
- Depending on which modules are installed, the following tabs may also be present:
  - XSEDE:
    - `Compliance`
    - `Sci Impact`
  - App Kernels:
    - `App Kernels`
  - SUPREMM
    - `Job Viewer`
- Click a few of the tabs and make sure that things are working as expected

**Test 2**
- spawn a shell in the container being tested
- login to mysql
- Execute `use moddb;`
- Execute the following and ensure there are no duplicate records 
```sql
SELECT * FROM acls;
```
- Execute the following and ensure there are no duplicate records:
```sql
SELECT * FROM tabs;
```

- Execute the following and ensure there are records present:
```sql
SELECT * FROM acl_hierarchies;
```
- Execute  the following and sure that the counts of tabs / acl are as expected.
```sql
SELECT 
  a.name, 
  COUNT(a.acl_id) as num 
FROM acl_tabs at 
JOIN acls a ON a.acl_id = at.acl_id 
JOIN tabs t ON t.tab_id = at.tab_id 
GROUP BY a.acl_id;
```

## Types of changes

## Checklist:


classes/Xdmod/Config.php
```diff
@@ -351,17 +376,158 @@ protected function mergeData(array $data, array $partialData)
...
-     * @return bool True if the array is associative.
+     * @param array $left
+     * @param array $right
+     * @return array
      */
-    protected function isAssocArray(array $arr)
+    protected function arrayRecursiveIntersect(array $left, array $right)
     {
-        return (bool)count(array_filter(array_keys($arr), 'is_string'));
+        $results = array();
```
jpwhite4: This code is mind-boggling complex. Please can you add some unit tests to confirm that it does what it is supposed to. Please run the code coverage tools to check that your unit tests cover all of the numerous code paths.

bin/acl-config
```diff
@@ -116,49 +116,82 @@ function main()
 
     // Module Retrieval
     $modules = $config['modules'];
-    if (null !== $modules['WARNING']) {
+    if (isset($modules['WARNING'])) {
```
plessbd: Why are there warning messages in the config file?
Should we be logging these warnings instead of just ignoring them (even if just logging to the screen)?

bin/acl-config
```diff
@@ -116,49 +116,82 @@ function main()
...
+        }
+        $moduleHierarchies[$module] = $hierarchies;
+    }
+
 
     // Acl Retrieval
     $roles = Roles::getRoleNames($blacklist);
-    $rolesData = $config->getModuleSection('roles');
 
     foreach ($modules as $module => $moduleData) {
```
plessbd: we are doing this loop twice, could we combine them?

bin/acl-config
```diff
@@ -116,49 +116,82 @@ function main()
...
                 }
             }
-        } else {
-            $log->notice("No role information for $module skipping...");
         }
     }
 
     // Datawarehouse Retrieval
-    $datawarehouseSource = $config->getModuleSection('datawarehouse');
     foreach ($modules as $module => $moduleData) {
```
plessbd: actually 3 times...  Could we merge them?

bin/acl-config
```diff
@@ -904,16 +932,18 @@ SQL;
...
         $inserted += $count;
         if ($count === 0) {
-            $log->info("[FAIL] Did not Process Acl Hierarchy: $id");
+            $log->notice("[SUCCESS] Acl Hierarchy: $id already exists");
         } elseif ($count === 1) {
-            $log->info("[SUCCESS] Processed Acl Hierarchy: $id");
+            $log->notice("[SUCCESS] Processed Acl Hierarchy: $id");
         } else {
-            $log->info("[FAIL] Inserted more than one Acl Hierarchy Record for: $id");
+            $log->warning("[FAIL] Inserted more than one Acl Hierarchy Record for: $id");
```
plessbd: When I think FAIL I think Fatal.  Probably should change this to a [WARNING] just to

bin/acl-config
```diff
@@ -1168,13 +1193,13 @@ SQL;
...
         if (!array_key_exists($realm, $realms)) {
-            $log->err("Unable to find a corresponding realm record for [$realm]. Skipping Query Descriptor for [ $module, $role, $groupBy ]");
+            $log->warning("Unable to find a corresponding realm record for [$realm]. Skipping Query Descriptor for [ $module, $acl, $groupBy ]");
             continue;
         }
 
         $realmData = $realms[$realm];
         if (!isset($realmData['statistics'])) {
-            $log->err("No statistics found for realm $realm. Skipping Query Descriptor for [ $module, $role, $groupBy ]");
+            $log->warning("No statistics found for realm $realm. Skipping Query Descriptor for [ $module, $acl, $groupBy ]");
```
plessbd: seems at times we are doing things like [SUCCESS] or [WARN] is there a reason we do or do not have that on all messages?

bin/acl-config
```diff
@@ -1586,7 +1611,7 @@ function verifyJsonSyntax()
  * the Acl system depends on. If an exception occurs while attempting to
  * retrieve as section then we can assume there is a problem with that section.
  *
- * @return 0 if all configuration sections were loaded successfully else 1.
+ * @return int 0 if all configuration sections were loaded successfully else 1.
```
plessbd: could we just change this to a boolean true or false?

bin/acl-config
ryanrath: It's a warning to users who may open the file that the contents have been automatically generated and that they modify them at their own risk.

plessbd: Can we add a comment in there with this infomration then just so someone like me doesn't ask this the next time I/they look at the code.

ryanrath: sure, not a problem

smgallo: I was going to say the same thing. Use the existing comment convention that we have and add a header section. Note that the ETL config file has a mechanism for stripping out comments prior to processing so we may want to adopt that at some point.

ryanrath: A header section?

smgallo: I.e., a block of comments at the top of the file.

ryanrath: I should say, I understand what a header section is, but I'm not sure how you mean it in this context. Do you mean for comments etc. to be placed in the header section?

plessbd:
```json
{
    "#" : "This file is auto generated",
    "#" : "Dont touch or you break all teh things"
}
```

bin/acl-config
```diff
@@ -116,49 +116,82 @@ function main()
...
+    foreach($modules as $module => $moduleData) {
+        $hierarchies = $config->filterByMetaData(
+            $config->getModuleSection('hierarchies'),
+            array(
+                'modules' => array(
+                    $module
+                )
+            )
+        );
+        if (isset($hierarchies['hierarchies'])) {
```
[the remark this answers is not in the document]
ryanrath: This was actually there to handle the differences between OpenXDMoD && XSEDE. I've since simplified the code so that this isn't needed.

bin/acl-config
```diff
@@ -116,49 +116,82 @@ function main()
...
+            $config->getModuleSection('roles'),
+            array(
+                'modules' => array(
+                    DEFAULT_MODULE_NAME,
+                    $module
+                )
+            )
+        );
+
+        if (isset($moduleData['roles'])) {
```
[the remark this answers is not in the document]
ryanrath: This is similar to the hierarchies one ( in that it was meant to handle the differences between OpenXDMoD && XSEDE ) but there was another wrinkle in that roles.json is laid out differently for OpenXDMoD and XSEDE. I have a branch for ccr-private-xdmod that includes the changes to roles.json that fixed this ( the problem was 'cs' extended 'cd' which led to some of the merged properties of 'cs' having unintended values).  This has also been simplified to no longer be needed.

classes/Xdmod/Config.php
ryanrath: I can ( will ) absolutely do that.

bin/acl-config
ryanrath: This has been updated and will be committed shortly

bin/acl-config
ryanrath: yup, has been updated / committed and will be pushed shortly.

bin/acl-config
ryanrath: yup, has been updated / committed and will be pushed shortly.

bin/acl-config
ryanrath: I can, but that would require more complicated calling code ( @ 1563 ). Either way it's cool.

smgallo: More complicated as in `$result += ( ! verifyJsonSyntax() ? 1 : 0 );`?

smgallo: Or better yet... `$numInvalidFiles += ( ! verifyJsonSyntax() ? 1 : 0 );`

ryanrath: I've updated the verify* functions in the latest commit.

jpwhite4:
In acl-config lines 204 and 205 you have the following:
```
    // We want to process 'xdmod' first.
    array_multisort($results, SORT_ASC);
```
Sorting the array in ascending alphabetical order will put appkernels at the beginning and xsede at the end so this clearly has no impact on xdmod being processed first. Please explain why the array is sorted and update the comment appropriately.

classes/Xdmod/Config.php
```diff
@@ -351,17 +376,158 @@ protected function mergeData(array $data, array $partialData)
...
+     */
+    protected function isAssocArray($value)
+    {
+        if (!is_array($value)) {
+            return false;
+        }
+        return (bool)count(array_filter(array_keys($value), 'is_string'));
+    }
+
+    public function filterByMetaData(array $data, array $metaData)
```
plessbd: missing docblock

ryanrath:
The latest commit contains tests for the new Config functionality. 

The code coverage for the tests is sitting at 84% of the lines in Config.php and 100% of the newly added / modified functions.

These functions include: 
- getModuleSection
- loadModuleSection
- mergeData ( not new, but tested )
- filterByMetaData
- arrayRecursiveIntersect
- isAssocArray

jpwhite4:
I'm trying to understand the code. I created the test case below:

```php
    public function testArrayRecurse()
    {
        $config = Config::factory();

        $left = array(
            'moo' => array(
                'meta-data' => array(
                    'blurk' => 'blark'
                ),
                'foo',
                array(
                    'bar' => 'baz'
                )
            )
        );

        $metadata = array(
            'blurk' => 'blark'
        );

        $expected = array(
            'moo' => array(
                'foo',
                array(
                    'bar' => 'baz'
                )
            )
        );

        $this->assertEquals($expected, $config->filterByMetaData($left, $metadata));

    }
```
I was expecting this test to pass, but it fails because the 'bar'=>'baz' is stripped from the result array. Is this the indented behaviour? If so please can you add some documentation that explains the different scenarios where the filterByMetaData will include or exclude tree entries.

jpwhite4:
I note that this test passes:
```php
   public function testArrayRecurse1()
      {
          $config = Config::factory();
  
          $left = array(
              'moo' => array(
                  'meta-data' => array(
                      'blurk' => 'blark'
                  ),
                  'foo',
                  array('one', 'two', 'three')
              )
          );
  
          $metadata = array(
              'blurk' => 'blark'
          );
  
          $expected = array(
              'moo' => array(
                  'foo',
                  array('one', 'two', 'three')
              )
          );
  
          $this->assertEquals($expected, $config->filterByMetaData($left, $metadata));
  
      }
```
which makes me think that the previous test failure is a bug rather than by design.

ryanrath:
@jpwhite4 That is actually by design ( or at least the current design, if we need to revise things that that's not a problem). If your original $left had been tagged by the `addMetaDataRecursive` function it would have looked like this: 
```php
$left = array(
            'moo' => array(
                'meta-data' => array(
                    'blurk' => 'blark'
                ),
                'foo',
                array(
                    'bar' => 'baz',
                    'meta-data' => array(
                         'blurk' => 'blark'
                    )
                )
            )
        ); 
````
and when you filtered on `array('blurk' => 'blark')` you would have received what you expected. 

Generally speaking, If the filtering encounters an associative array, it will only be included if there is at least one commonality between it's meta-data property and the meta-data being used as a filter. A concrete example being: Say you want to filter for everything that was added by the SUPREMM module. You're meta-data would look like: `array( 'modules' => array( 'xdmod', 'supremm') )`. `xdmod` is included so as to model the way in which our modules 'extend' xdmod. The results of the filtering w/ this meta-data will return all data that is tagged as being added by 'xdmod' or 'supremm' but will ensure that data added by any of the other modules are not included.

bin/acl-config
ryanrath: Typically I include the "tags" if there was some action that was executed and I want to call attention to the outcome.

PR #309, opened 2017-10-24
Storage Realm (Alpha)

jtpalmer:
## Description

Add storage realm and basic ETL for storage data.

## Motivation and Context

Storage alpha is overdue.

## Tests performed

Ran tests in docker container with and without storage realm enabled.

## Types of changes

## Checklist:


plessbd:
Would like to see some tests added, but since this is alpha I guess it is ok...

classes/DataWarehouse/Query/Storage/GroupBy.php
```diff
@@ -0,0 +1,224 @@
...
+/**
+ * Storage query GroupBy base class.
+ */
+abstract class GroupBy extends \DataWarehouse\Query\GroupBy
+{
+
+    /**
+     * @var \DataWarehouse\Query\Model\Schema;
+     */
+    protected $_schema; // @codingStandardsIgnoreLine
```
smgallo: There are enough of these underscore variables used that I think we should remove this sniff. @jpwhite4 @plessbd thoughts?

configuration/etl/etl.d/storage.json
```diff
@@ -0,0 +1,124 @@
...
+            "truncate_destination": false,
+            "namespace": "ETL\\Ingestor",
+            "class": "StructuredFileIngestor",
+            "options_class": "IngestorOptions",
+            "endpoints": {
+                "destination": {
+                    "type": "mysql",
+                    "name": "Storage",
+                    "config": "datawarehouse",
+                    "schema": "modw_storage"
```
smgallo: You may want to add `"create_schema_if_not_exists": true` here to automatically create the destination schema for you (assuming that the database user has the proper permissions).

configuration/etl/etl.d/storage.json
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+    },
+    "storage": [
+        {
+            "name": "StorageFileSystemIngestor",
+            "description": "Ingests file system data",
+            "definition_file": "storage\/file_system.json",
```
smgallo: Do you need to escape the forward slash here? I haven't had to do that for paths.

configuration/etl/etl_specs.d/storage/user_usage.schema.json
```diff
@@ -0,0 +1,60 @@
+{
+    "$schema": "http:\/\/json-schema.org\/draft-04\/schema#",
```
smgallo: Same here, do we need to escape the forward slashes?

configuration/etl/etl_tables.d/storage/directory.json
```diff
@@ -0,0 +1,39 @@
...
+        "columns": [
+            {
+                "name": "directory_id",
+                "type": "int(11)",
+                "nullable": false,
+                "extra": "AUTO_INCREMENT"
+            },
+            {
+                "name": "path",
+                "type": "varchar(255)",
```
jpwhite4: According to the internet: Linux has a maximum filename length of 255 characters for most filesystems (including EXT4), and a maximum path of 4096 characters. Would turning this into a TEXT field or a longer varchar cause problems elsewhere?

classes/DataWarehouse/Query/Storage/GroupBy.php
jpwhite4: I'd prefer to not selectively turn off the sniffs: Either (1) we allow underscores or (2) we don't. If we don't allow them then we shouldn't be adding new code that uses underscores. If we do allow them then we shouldn't be forbidding them in the linter.

In this case this is a new variable, so could easily be named without an underscore prefix. @jtpalmer  Are there any cases where one has to create a variable with an underscore due to the existing API?

In any case I don't have a strong opinion on whether we pick option (1) or (2) only that we should pick option (1) or (2).

smgallo: I'd vote for (1)

jtpalmer: I used underscores in this class to be consistent with properties inherited from `DataWarehouse\Query\GroupBy`, but will change them.

configuration/etl/etl.d/storage.json
jtpalmer: [Escaping forward slashes is optional](https://tools.ietf.org/html/rfc7159#section-7).  I ran all of the new JSON files through a formatter (i.e. output from `json_encode`) which added those.  Easy enough to remove if you'd prefer they not be included.

configuration/etl/etl_tables.d/storage/directory.json
jtpalmer: I put a unique key on that column and am having some issues with the maximum key length when I try to increase the size of the column.  @smgallo Is there a way to specify the length of an index in the ETL configuration?

I could just remove the index for now.

configuration/etl/etl.d/storage.json
smgallo: I'd rather keep them consistent with everything else that we have

configuration/etl/etl_tables.d/storage/directory.json
smgallo: In it's current state it assumes that the entire string is a column name. I'll add to my shortlist to support the "column(length)" syntax, which should be pretty simple since its only checking to make sure that the column exists in the definition before sending it to the database.

PR #310, opened 2017-10-26
Update MailWrapper.php

plessbd:
catch sendmail not working and continue on

This is specifically for making sure tests pass in docker.


plessbd:
Requested changes to be made in a different pull request

PR #313, opened 2017-10-31
Handling acls that require centers to be provided

ryanrath:
## Description
- Actually added the code that handles processing / reporting when an acl
  requires additional information to function properly. In particular, when an
  acl requires a 'provider' ( center / service_provider ).

## Motivation and Context
Kinda required.

## Tests performed
Manual Tests:
- Spin up a the supremm docker image / install xdmod (this branch), appkernels, supremm / setup & ingest everything
- open a browser and point it at XDMoD / the internal dashboard
- behold, things be less broke then before.

## Types of changes

## Checklist:

### DIFF
--- a/classes/Models/Services/Acls.php
+++ b/classes/Models/Services/Acls.php
@@ -900,7 +900,7 @@ SQL;
 
         if (count($rows) > 0) {
             return array_reduce($rows, function ($carry, $item) {
-                $carry []= new Acl($item);
+                $carry[] = new Acl($item);
                 return $carry;
             }, array());
         }
--- a/classes/XDAdmin.php
+++ b/classes/XDAdmin.php
@@ -109,14 +109,11 @@ class XDAdmin
     public function enumerateAcls()
     {
         $sql = <<<SQL
-SELECT a.*,
-  req.acl_id IS NOT NULL AS requires_center
-FROM acls a LEFT JOIN
-  (
-    SELECT a2.acl_id 
-    FROM acls a2 WHERE a2.name IN ('cs','cd')
-  ) req ON req.acl_id = a.acl_id
-  ORDER BY a.display;
+SELECT
+a.*,
+  CASE WHEN a.name IN ('cd', 'cs') THEN 1 ELSE 0 END requires_center
+FROM acls a
+ORDER BY a.display;
 SQL;
 
         return $this->moddb->query($sql);
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -970,7 +970,7 @@ SQL;
         foreach ($this->_roles as $role) {
             $roleId = $this->_getRoleID($role);
             if ($roleId === null) {
-                throw new Exception("Unable to find an id for: $roleId. Unable to save user.");
+                continue;
             }
             $this->_pdo->execute(
                 "INSERT INTO UserRoles VALUES(:id, :roleId, '0', '0')",
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -1450,7 +1450,7 @@ FROM user_acls ua
 -- we only want records that are related to a specific user
 -- the original sql implicitly left out the flag or feature acls
 -- so we need to filter these out here
-WHERE ua.user_id = :user_id AND at.name != 'feature'
+WHERE ua.user_id = :user_id AND at.name = 'data'
 -- In this ordering we use coalesce so that any acl that does not participate
 -- in a hierarchy will be sent to the bottom of the list
 ORDER BY COALESCE(aclh.level, 0) DESC, a.name
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -579,7 +579,7 @@ SQL;
     public function isDeveloper()
     {
 
-        return (in_array(ROLE_ID_DEVELOPER, $this->_acls));
+        return (array_key_exists(ROLE_ID_DEVELOPER, $this->_acls));
 
     }//isDeveloper
 
--- a/configuration/portal_settings.ini
+++ b/configuration/portal_settings.ini
@@ -1,15 +1,15 @@
 [general]
 title = "Open XDMoD"
-site_address = "http://127.0.0.1:8080/"
-user_manual = "http://127.0.0.1:8080/user_manual/"
-contact_page_recipient = "ccr-xdmod-help@buffalo.edu"
-tech_support_recipient = "ccr-xdmod-help@buffalo.edu"
+site_address = "http://localhost:8080/"
+user_manual = "http://localhost:8080/user_manual/"
+contact_page_recipient = ""
+tech_support_recipient = ""
 
 ; The version number is updated during the upgrade process.
 version = "7.1.0"
 
 debug_mode = "off"
-debug_recipient = "ccr-xdmod-help@buffalo.edu"
+debug_recipient = ""
 
 exceptions_logfile = "exceptions.log"
 rest_logfile = "rest_exceptions.log"
@@ -66,26 +66,26 @@ basic_auth = "on"
 
 [mailer]
 sender_name = "Open XDMoD Mailer"
-sender_email = "ccr-xdmod-help@buffalo.edu"
+sender_email = ""
 captcha_public_key = ""
 captcha_private_key = ""
 
 [reporting]
 base_path = "/usr/share/xdmod/reporting/jasper_builder"
-java_path = "/usr/bin/java"
-javac_path = "/usr/bin/javac"
-phantomjs_path = "/usr/local/bin/phantomjs"
+java_path = ""
+javac_path = ""
+phantomjs_path = ""
 
 [logger]
 db_engine = "MySQLDB"
-host = "127.0.0.1"
-port = "3308"
+host = "localhost"
+port = "3306"
 user = "xdmod"
-pass = "xdmod123"
+pass = ""
 database = "mod_logger"
 table = "log_table"
-email_from = "ccr-xdmod-help@buffalo.edu"
-email_to = "ccr-xdmod-help@buffalo.edu"
+email_from = ""
+email_to = ""
 email_subject = "Open XDMoD Logger"
 default_level_console = "NOTICE"
 default_level_file = "WARNING"
@@ -94,36 +94,36 @@ default_level_mail = "ERR"
 
 [database]
 db_engine = "MySQLDB"
-host = "127.0.0.1"
-port = "3308"
+host = "localhost"
+port = "3306"
 user = "xdmod"
-pass = "xdmod123"
+pass = ""
 database = "moddb"
 
 ; This section applies to modw, modw_aggregates, and modw_filters.
 [datawarehouse]
 db_engine = "MySQLDB"
-host = "127.0.0.1"
-port = "3308"
+host = "localhost"
+port = "3306"
 user = "xdmod"
-pass = "xdmod123"
+pass = ""
 database = "modw"
 rest_logfile = "rest_datawarehouse.log"
 
 [shredder]
 db_engine = "MySQLDB"
-host = "127.0.0.1"
-port = "3308"
+host = "localhost"
+port = "3306"
 user = "xdmod"
-pass = "xdmod123"
+pass = ""
 database = "mod_shredder"
 
 [hpcdb]
 db_engine = "MySQLDB"
-host = "127.0.0.1"
-port = "3308"
+host = "localhost"
+port = "3306"
 user = "xdmod"
-pass = "xdmod123"
+pass = ""
 database = "mod_hpcdb"
 
 [slurm]
--- a/html/controllers/user_admin/create_user.php
+++ b/html/controllers/user_admin/create_user.php
@@ -31,9 +31,11 @@ if (isset($_POST['acls'])) {
     // are not used for data access.
     $aclNames = array();
     $featureAcls = Acls::getAclsByTypeName('feature');
-    if (count($featureAcls) > 0) {
+    $tabAcls = Acls::getAclsByTypeName('tab');
+    $uiOnlyAcls = array_merge($featureAcls, $tabAcls);
+    if (count($uiOnlyAcls) > 0) {
         $aclNames = array_reduce(
-            $featureAcls,
+            $uiOnlyAcls,
             function ($carry, Acl $item) {
                 $carry []= $item->getName();
                 return $carry;
@@ -41,13 +43,8 @@ if (isset($_POST['acls'])) {
             array()
         );
     }
-    $found = false;
-    foreach(array_keys($acls) as $acl) {
-        if (!in_array($acl, $aclNames)) {
-            $found = true;
-            break;
-        }
-    }
+    $diff = array_diff(array_keys($acls), $aclNames);
+    $found = !empty($diff);
     if (!$found) {
         \xd_response\presentError('Please include a non-feature acl ( i.e. User, PI etc. )');
     }
--- a/html/controllers/user_admin/create_user.php
+++ b/html/controllers/user_admin/create_user.php
@@ -84,13 +84,10 @@ try {
             $centerConfig = array();
             $count = 0;
             foreach($centers as $center) {
-                switch($count) {
-                    case 0:
-                        $config = array('primary' => 1, 'active' => 1);
-                        break;
-                    default:
-                        $config = array('primary' => 0, 'active' => 0);
-                        break;
+                if ($count === 0) {
+                    $config = array('primary' => 1, 'active' => 1);
+                } else {
+                    $config = array('primary' => 0, 'active' => 0);
                 }
                 $centerConfig[$center] = $config;
                 $count += 1;
--- a/html/controllers/user_admin/create_user.php
+++ b/html/controllers/user_admin/create_user.php
@@ -94,6 +94,10 @@ try {
             }
             $newuser->setOrganizations($centerConfig, $acl);
         } elseif (in_array($acl, array('cd', 'cs'))) {
+            // This block pertains to OpenXDMoD. Specifically, no centers are
+            // returned with acls in OpenXDMOD as there is only one to choose
+            // from. So we catch that use case here and provide the one center
+            // for 'center' related acls.
             $currentCenters = Centers::getCenters();
             if (count($currentCenters) > 0) {
                 $center = $currentCenters[0]['id'];
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -167,8 +167,8 @@ $params = array('uid' => RESTRICTION_UID);
          $user_to_update->disassociateWithInstitution();
       }
       else {
-        $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
-           $user_to_update->setInstitution($_POST['institution'], $isPrimary);
+            $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
+            $user_to_update->setInstitution($_POST['institution'], $isPrimary);
       }
 
    }//if (isset($_POST['institution']))
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -8,15 +8,15 @@ use Models\Acl;
 $params = array('uid' => RESTRICTION_UID);
 
    $isValid = xd_security\secureCheck($params, 'POST');
-	
+
    if (!$isValid) {
       $returnData['success'] = false;
       $returnData['status'] = 'invalid_id_specified';
       xd_controller\returnJSON($returnData);
    };
-	
+
    // -----------------------------
-	
+
    $user_to_update = XDUser::getUserByID($_POST['uid']);
 
    if ($user_to_update == NULL) {
@@ -24,9 +24,9 @@ $params = array('uid' => RESTRICTION_UID);
       $returnData['status'] = 'user_does_not_exist';
       xd_controller\returnJSON($returnData);
    }
-	
+
    // -----------------------------
-	
+
    $params = array(
       'first_name'    => RESTRICTION_FIRST_NAME,
       'last_name'     => RESTRICTION_LAST_NAME,
@@ -36,9 +36,9 @@ $params = array('uid' => RESTRICTION_UID);
    );
 
    \xd_security\assertEmailParameterSet('email_address');
-		
+
    $qualifyingParams = xd_security\secureCheck($params, 'POST', false);
-	
+
    if ($qualifyingParams == 0) {
       $returnData['success'] = false;
       $returnData['status'] = 'need_update_information';
@@ -46,7 +46,7 @@ $params = array('uid' => RESTRICTION_UID);
    }
 
    // -----------------------------
-   
+
    $me = XDUser::getUserByID($_SESSION['xdDashboardUser']);
 
    if ($me->getUserID() == $user_to_update->getUserID()) {
@@ -58,7 +58,7 @@ $params = array('uid' => RESTRICTION_UID);
       }
 
    	if (isset($_POST['acls'])) {
-   	
+
          $role_config = json_decode($_POST['acls'], true);
          if (!array_key_exists(ROLE_ID_MANAGER, $role_config)) {
                 $returnData['success'] = false;
@@ -67,36 +67,36 @@ $params = array('uid' => RESTRICTION_UID);
          }
       }//if (isset($_POST['acls']))
 
-   } 
+   }
 
    if (isset($_POST['first_name']))       $user_to_update->setFirstName($_POST['first_name']);
    if (isset($_POST['last_name']))        $user_to_update->setLastName($_POST['last_name']);
-   
+
    if (isset($_POST['email_address'])) {
-      
+
       $email_address = (strlen($_POST['email_address']) > 0) ? $_POST['email_address'] : NO_EMAIL_ADDRESS_SET;
 
       if ( ($user_to_update->getUserType() != XSEDE_USER_TYPE) && ($email_address == NO_EMAIL_ADDRESS_SET) ) {
             $returnData['success'] = false;
             $returnData['status'] = 'This XDMoD user must have an e-mail address set.';
-            xd_controller\returnJSON($returnData);        
+            xd_controller\returnJSON($returnData);
       }
-      
+
       $user_to_update->setEmailAddress($email_address);
-   
+
    }
-   
-   if (isset($_POST['assigned_user']))    $user_to_update->setPersonID($_POST['assigned_user']);				
+
+   if (isset($_POST['assigned_user']))    $user_to_update->setPersonID($_POST['assigned_user']);
    if (isset($_POST['is_active']))        $user_to_update->setAccountStatus($_POST['is_active'] == 'y' ? ACTIVE : INACTIVE);
-   
+
    if (isset($_POST['user_type'])) {
-   
+
       if ($user_to_update->getUserType() != XSEDE_USER_TYPE) {
-      
+
          $user_to_update->setUserType($_POST['user_type']);
 
       }
-            
+
    }
 
 
@@ -160,19 +160,19 @@ $params = array('uid' => RESTRICTION_UID);
     }
 
    // -----------------------------
-       
+
    if (isset($_POST['institution'])) {
 
       if ($_POST['institution'] == -1) {
          $user_to_update->disassociateWithInstitution();
       }
       else {
-         $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
-         $user_to_update->setInstitution($_POST['institution'], $isPrimary);
+        $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
+           $user_to_update->setInstitution($_POST['institution'], $isPrimary);
       }
 
    }//if (isset($_POST['institution']))
-   
+
    // -----------------------------
 
    try {
@@ -181,19 +181,19 @@ $params = array('uid' => RESTRICTION_UID);
    catch(Exception $e) {
       $returnData['success'] = false;
       $returnData['status'] = $e->getMessage();
-      xd_controller\returnJSON($returnData);	
+      xd_controller\returnJSON($returnData);
    }
 
    $returnData['success'] = true;
-   
+
    $statusPrefix = $user_to_update->isXSEDEUser() ? 'XSEDE ' : '';
    $displayUsername = $user_to_update->isXSEDEUser() ? $user_to_update->getXSEDEUsername() : $user_to_update->getUsername();
-   
+
    $returnData['status'] = $statusPrefix."User <b>$displayUsername</b> updated successfully";
-   
+
    $returnData['username'] = $user_to_update->getUsername();
    $returnData['user_type'] = $user_to_update->getUserType();   //if isset()...
-   
-   xd_controller\returnJSON($returnData);	
+
+   xd_controller\returnJSON($returnData);
 
 ?>
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -148,7 +148,7 @@ $params = array('uid' => RESTRICTION_UID);
          $user_to_update->disassociateWithInstitution();
       }
       else {
-         $user_to_update->setInstitution($_POST['institution'], array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $role_config));
+         $user_to_update->setInstitution($_POST['institution'], array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls));
       }
 
    }//if (isset($_POST['institution']))
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -125,13 +125,10 @@ $params = array('uid' => RESTRICTION_UID);
                 $centerConfig = array();
                 $count = 0;
                 foreach($centers as $center) {
-                    switch($count) {
-                        case 0:
-                            $config = array('primary' => 1, 'active' => 1);
-                            break;
-                        default:
-                            $config = array('primary' => 0, 'active' => 0);
-                            break;
+                    if ($count === 0 ) {
+                        $config = array('primary' => 1, 'active' => 1);
+                    } else {
+                        $config = array('primary' => 0, 'active' => 0);
                     }
                     $centerConfig[$center] = $config;
                     $count += 1;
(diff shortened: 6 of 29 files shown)


classes/XDUser.php
```diff
@@ -1639,22 +1639,25 @@ public function setOrganizations($organization_ids = array(), $role = ROLE_ID_CE
...
-        $this->_pdo->execute(
-            "DELETE FROM user_acl_group_by_parameters WHERE user_id = :user_id AND acl_id = :acl_id AND group_by_id IN (SELECT gb.group_by_id FROM group_bys gb WHERE gb.name = 'provider')",
-            array(
-                ':user_id' => $this->_id,
-                ':acl_id' => $acl->getAclId()
-            )
-        );
+
+        $acl = Acls::getAclByName($role);
+        if (isset($acl)) {
```
jpwhite4: Under what circumstances would `$acl` not be set? Should you be logging a warning in this case? or perhaps throwing an exception if this indicates corrupt data in the database?

classes/Models/Services/Acls.php
```diff
@@ -445,7 +447,15 @@ function ($carry, $item) use ($handle) {
...
         );
 
+        $previousName = null;
         foreach ($rows as $row) {
+            $name = $row['name'];
+            if ($name != $previousName) {
+                $previousName = $name;
+            }
+            if (!isset($results[$name])) {
+                $results[$name] = array();
```
jpwhite4: Please can you explain the reason for this code. From manual code inspection, it looks to me like the empty array added to `$results` will get filtered out in line 468 so will not change the return value of the function at all.

jpwhite4: Please can you explain what is the purpose of this code? It appears to me that this code does nothing since the empty array that you add to `$results` here is being stripped out on line 468.

classes/XDUser.php
ryanrath: It will not be set when an acl is not found that has a `name` value = `$role`. I do agree that an exception should probably be thrown.

jpwhite4: To clarify my question: Under what circumstances will an acl not be found with a `name` value = `$role`

configuration/roles.json
```diff
@@ -366,6 +366,9 @@
                 "level": 400,
                 "filter_override": false
             }],
+            "requires": [
```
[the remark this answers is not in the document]
ryanrath: I'm certainly open to suggestions on how to store / represent the relationship between some acls and some additional data that will be required when a user is assigned the acl. Concrete example: In an XDMoD install that supports multiple centers, when a user is assigned the Center Director acl, an associated center must be supplied. 

I do need to make sure that there is a `ConfigFileMigration`, thank you for reminding me.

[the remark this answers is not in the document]
smgallo: In this case, since the dimension already exists, could there simple be a "multiple: true" flag with an appropriate name?

jpwhite4:
I just tried to test on xdmod-dev (@20:38 on Wednesday) and found that it is not behaving as expected.

I created a new user (Tarquin Farqar) and made him a Center Staff at TACC and mapped to TAS Communitiy user. This user account should be able to see all TACC jobs in addition to TAS jobs in the SUPReMM realm, but it can only see the TAS jobs. See e.g. screenshot

image

smgallo:
I don't know if this is related to this PR, but I was accessing the internal dashboard to test adding a center role to one of my test users. I accidentially opened the wrong user (nikolaystest) who happens to be a Campus Staff (which should be Center Staff).

local_dashboard_nikolaystest

I then selected my test user (z3gallo@yahoo.com) and upon opening the Campus Staff dialog it has the same Center selected as Nikolay's user. It appears that the dialog is not properly resetting itself when a new user is selected. The same is true of the Center Director dialog, I selected wilkinsn who is an SDSC center director and when I viewed the dialog for my test user SDSC was also selected. Saving the data does indeed add these ACLs to my user when they should not.

local_dashboard_z3gallo

Also, simply opening and closing any of the Campus Champion, Campus Staff (which should be Center Staff), or Center Director dialogs, the UI tells me that data has changed when it has not.

unsaved_changes

configuration/roles.json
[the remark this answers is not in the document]
jpwhite4: Given that the only supported dimension is "provider", you could always just require multiple if the dimension is "provider" and not change the config file at all. The fact of multiple providers is hard coded throughout the current code (including the new ACL stuff). We can then stick to the original design plan (from Steve's email dated May 24 2017) which moved the configuration API design and implementation to Phase 4 (recall this code change is part of Phase 1.4).

A few selected examples of where the multiple providers is hardcoded in the current code:
addAclOrganization() in XDUser
controllers/user_admin/update_user.php lines 110-115
html/internal_dashboard/js/admin_panel/AclGrid.js lines 197-206, 224-236, 259-

jpwhite4:
Also I just checked on metrics-dev (which does not have the multiple centers change) and it also has the problem with the permissions update.

Steps to reproduce:
Update an existing user that did not previously have center director ACL.
Login as the user and view SUPReMM data. 
Observe that instead of being able to see all data, they can only view the same data that they could before the update.

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -202,7 +212,12 @@ XDMoD.Admin.AclGrid.CenterSelector = Ext.extend(Ext.menu.Menu, {
             baseParams: {
                 operation: 'enum_resource_providers'
             },
-            autoLoad: true
+            autoLoad: true,
+            listeners: {
+                load: function() {
+
```
jpwhite4: Please can you double check this: the function is empty.

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -176,6 +183,9 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
 
     reset: function () {
         this.ccInclude.reset();
+        for (var key in this.aclCenters) {
+            this.aclCenters[key] = [];
```
jpwhite4: needs an 
```javascript
if(this.aclCenters.hasOwnProperty(key)) {
```

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -15,14 +14,14 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
 
     initComponent: function () {
         var self = this;
-        var isDirty = true;
+        this.isDirty = true;
```
jpwhite4: Shouldn't this be initialized to false? The grid can only be Dirty once a user has interacted with it and that cannot have happened at this point in the code.

smgallo:
@ryanrath I tested on xdmod-dev and the behavior that I saw previously with the Center Director and Center Staff dialogs not being properly initialized between users is no longer occurring.

classes/XDUser.php
```diff
@@ -1103,7 +1106,7 @@ private function _getRoleID($role_abbrev)
             ':abbrev' => $role_abbrev,
         ));
 
-        return $roleResults[0]['role_id'];
+        return count($roleResults) > 0 ? $roleResults[0]['role_id'] : null;
```
chakrabortyr: If 'role_id is guaranteed to exist at this point and we're only ever referencing the initial index, why are we using count? Couldn't this just be:

`return !empty($roleResults) ? $roleResults[0]['role_id'] : null;`

Also, why is there a null check here and not in other get_Parameter_ methods?

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -66,15 +65,37 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
...
+                var aclCenterExists = self.aclCenters.hasOwnProperty(record.data.acl_id);
+                var aclHasCenters = aclCenterExists && self.aclCenters[record.data.acl_id].length > 0;
+
+                // If this acl requires center information
+                // and we have an entry for this acl in our center list
+                // and the entry for this acl has centers currently associated
+                // with it.
+                if (requires_center && aclCenterExists && aclHasCenters) {
+                    record.set(this.dataIndex, true);
+                } else if (requires_center && aclCenterExists && !aclHasCenters) {
```
chakrabortyr: Readability nitpick: Can you please move the empty control statement to the top of the conditionals and put the more verbose explanations for the behavior in a block comment above that?

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -150,7 +173,14 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
 
     setSelectedAcls: function (acls) {
         this.store.each(function (record) {
-            record.set('include', acls.indexOf(record.data.acl_id) >= 0);
+            var hasAcl = acls.indexOf(record.data.acl_id) >= 0;
```
smgallo: Wouldn't it be clearer here if you explicitly checked for the invalid return value of -1?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -494,14 +497,13 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
...
                 var acls = newUserRoleGrid.getSelectedAcls();
-                var missingMappedUser = cmbUserMapping.getValue().length === 0;
 
                 if (acls.length <= 0) {
                     CCR.xdmod.ui.userManagementMessage('This user must have at least one role.', false);
                     return;
                 }
 
-                if ((acls.indexOf('usr') >= 0 || acls.indexOf('pi') >= 0) && missingMappedUser) {
+                if ((acls.indexOf('usr') >= 0 || acls.indexOf('pi') >= 0) && cmbUserMapping.getValue().length === 0) {
```
smgallo: Similarly here, if you're concerned if the value is in the array but don't care about where it is why not check `!= -1`?

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/TestFiles.php
```diff
@@ -0,0 +1,61 @@
...
+     * The current environment setting for this utility class.
+     * @var string
+     */
+    private $env;
+
+    /**
+     * TestFiles constructor.
+     * @param string $baseDir the base directory to use when retrieving test
+     * files.
+     */
```
smgallo: Need to add a `@param` for the 2nd optional parameter.

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/TestFiles.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __construct($baseDir, $defaultEnvironment = null)
+    {
+        $this->baseDir = $baseDir;
```
smgallo: Shouldn't you check that `$baseDir` is a valid directory?

html/controllers/user_admin/update_user.php
```diff
@@ -101,18 +101,46 @@
...
-                    $user_to_update->addAclOrganization($aclName, $center);
+                $centerConfig = array();
+                $count = 0;
+                foreach($centers as $center) {
+                    switch($count) {
+                        case 0:
+                            $config = array('primary' => 1, 'active' => 1);
+                            break;
+                        default:
+                            $config = array('primary' => 0, 'active' => 0);
```
chakrabortyr: Piggy backing off of what Steve said, is it even worth the conditional? You're only setting it to zero after the first iteration.

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -255,7 +297,7 @@ XDMoD.Admin.AclGrid.CenterSelector = Ext.extend(Ext.menu.Menu, {
             ]
         });
 
-        var grid = new Ext.grid.GridPanel({
+        this.grid = new Ext.grid.GridPanel({
```
chakrabortyr: Why have we changed the context of the grid?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -526,6 +528,17 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
...
                 }
+                var hasOnly = function (data, required) {
+                    var count = 0;
+                    for (var i = 0; i < required.length; i++) {
+                        count += data.indexOf(required[i]) !== -1 ? 1 : 0;
+                    }
+                    return count === data.length;
+                };
+                if (hasOnly(acls, ['dev', 'mgr'])) {
+                    CCR.xdmod.ui.userManagementMessage('You must select a non-flag acl for the user. ( i.e. anything not Manager or Developer ');
```
chakrabortyr: If the goal is to ensure that non-flag ACLs are selected, is there a simpler way to achieve that?

classes/XDUser.php
ryanrath: - `why are we using count?` 
  - because we can? Why use empty?  
- `Couldn't this just be:`
  - Yup, sure could. 
- `why is there a null check here...`
  - which other `get_Parameter_methods` ones that were updated in this PR or just in general?

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: Yup yup, update incoming

classes/Models/Services/Acls.php
```diff
@@ -859,4 +858,38 @@ public static function getMostPrivilegedAcl(XDUser $user, $moduleName = DEFAULT_
...
+        $rows = $db->query(
+            $query,
+            array(
+                ':acl_type_name' => $aclTypeName
+            )
+        );
+
+        if (count($rows) > 0) {
+            return array_reduce($rows, function ($carry, $item) {
+                $carry []= new Acl($item);
```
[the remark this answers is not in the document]
ryanrath: yup, change made.

configuration/portal_settings.ini
```diff
@@ -1,15 +1,15 @@
 [general]
 title = "Open XDMoD"
-site_address = "http://localhost:8080/"
-user_manual = "http://localhost:8080/user_manual/"
-contact_page_recipient = ""
-tech_support_recipient = ""
+site_address = "http://127.0.0.1:8080/"
+user_manual = "http://127.0.0.1:8080/user_manual/"
+contact_page_recipient = "ccr-xdmod-help@buffalo.edu"
```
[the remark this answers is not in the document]
ryanrath: yeah, I'm not sure what's going on here. I'll make sure that portal_settings.ini isn't in the change set.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: Do you have a suggestion?

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: So that it can be referenced easily elsewhere within the code.

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -66,15 +65,29 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
...
+
+                var requires_center = record.data.requires_center;
+                var aclCenterExists = self.aclCenters.hasOwnProperty(record.data.acl_id);
+                var aclHasCenters = aclCenterExists && self.aclCenters[record.data.acl_id].length > 0;
+
+                // If this acl requires center information
+                // and we have an entry for this acl in our center list
+                // and the entry for this acl has centers currently associated
+                // with it.
+                if (requires_center && aclCenterExists && aclHasCenters) {
```
smgallo: I think that you need to remove the 3rd condition `&& aclHasCenters` from the outer `if` here. IIRC, You want to set the record if `requires_center && aclCenterExists && aclHasCenters` but not if `requires_center && aclCenterExists && ! aclHasCenters`

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/TestFiles.php
```diff
@@ -0,0 +1,66 @@
...
+     * TestFiles constructor.
+     * @param string $baseDir the base directory to use when retrieving test
+     * @param string $defaultEnvironment the value to use as the default
+     * environment if one is not found
+     * files.
+     */
+    public function __construct($baseDir, $defaultEnvironment = null)
+    {
+        if (!is_dir($baseDir)) {
+            throw new Exception("Base Dir: $baseDir is not a directory. Unable to continue");
```
smgallo: This needs to be `\Exception` since we are in a namespace.

chakrabortyr:
LGTM once you fix the stuff Steve mentioned.

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: yeah, forgot to remove it when I moved it inside the outer if. Updating now

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/exportDialog.js
```diff
@@ -22,7 +22,6 @@ describe('Export Dialog', function () {
         ];
         expect(browser.getText(xdmod.selectors.exportDialog.comboListItems)).to.deep.equal(expected);
         browser.waitAndClick(xdmod.selectors.exportDialog.formatDropdown());
-        browser.waitForInvisible(xdmod.selectors.exportDialog.comboList);
```
jpwhite4: Why was this line removed?

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
```diff
@@ -158,7 +158,7 @@ class MetricExplorer {
         browser.waitUntilAnimEnd(this.selectors.catalog.collapseButton);
     }
     setDateRange(start, end) {
-        browser.waitForAllInvisible('.ext-el-mask');
+        browser.waitUntilNotExist('.ext-el-mask');
```
jpwhite4: Why was this changed?

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
```diff
@@ -389,11 +390,10 @@ class MetricExplorer {
...
     }
     addDataViaToolbar() {
-        browser.click(this.selectors.addData.button);
-        // Click on Jobs
-        browser.waitAndClick(this.selectors.buttonMenu.firstLevel + ' ul li:nth-child(3)');
-        // click on CPU Hours: Per Job
-        browser.waitAndClick(this.selectors.addData.secondLevel + ' ul li:nth-child(2)');
+        var n = 'CPU Hours: Per Job';
+        browser.waitAndClick(this.selectors.addData.button);
+        browser.waitAndClick('//div[@id="metric-explorer-chartoptions-add-data-menu"]//span[contains(text(), "Jobs")]');
```
jpwhite4: Please don't add hardcoded selectors - these should be added as a variable to the page class and then referenced.

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
```diff
@@ -26,6 +26,7 @@ then
     xdmod-import-csv -t names -i $REF_DIR/names.csv
     xdmod-ingestor
     php /root/bin/createusers.php
+    acl-import
```
jpwhite4: How come you run acl-import here? If this is now a required thing to do then we will need to update the user documentation

open_xdmod/modules/xdmod/integration_tests/lib/Controllers/UsageExplorerTest.php
```diff
@@ -122,7 +136,7 @@ public function testAggregateView()
     "group_by": "none",
     "statistic": "avg_wallduration_hours",
     "start_date": "2016-01-01",
-    "end_date": "2017-12-31",
+    "end_date": "2017-01-01",
```
jpwhite4: How come the date range changed?

open_xdmod/modules/xdmod/automated_tests/artifacts/update-artifacts.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
plessbd: why are we adding this file again?

It looks like it already exists twice (i take a bit of blame for that) but we should really only have it once.

ryanrath: Because this is the currently supported way of adding test artifact support to a set of tests. 

Per our conversation, I 100% agree that the test artifacts should be centralized and will be the first to submit a pr with those changes... after this PR is merged.

html/controllers/user_admin/create_user.php
```diff
@@ -48,10 +79,30 @@
...
+                    $config = array('primary' => 0, 'active' => 0);
+                }
+                $centerConfig[$center] = $config;
+                $count += 1;
+            }
+            $newuser->setOrganizations($centerConfig, $acl);
+        } elseif (in_array($acl, array('cd', 'cs'))) {
+            $currentCenters = Centers::getCenters();
+            if (count($currentCenters) > 0) {
+                $center = $currentCenters[0]['id'];
```
jpwhite4: How come we are hardcoding the first center here? what happens if there are multiple centers?

classes/Models/Services/Centers.php
```diff
@@ -0,0 +1,69 @@
...
+{
+
+    public static function getCenters()
+    {
+        $db = DB::factory('database');
+
+        return $db->query("SELECT o.* FROM modw.organization o");
+    }
+
+    public static function getCenterStaffMembers(XDUser $user)
```
jpwhite4: Why is this not a member function of the XDUser class?

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -66,15 +65,29 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
...
                     this.grid.getView().findRowIndex(node)
                 );
-
-                record.set(this.dataIndex, !record.data[this.dataIndex]);
-
-                isDirty = true;
-
-                if (self.selectionChangeHandler) {
-                    self.selectionChangeHandler();
+                // if this record requires a center AND
```
jpwhite4: This comment appears to be trunca

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/exportDialog.js
ryanrath: This was a dev change that shouldn't have made it in. I was experiencing what seemed like inordinately long run times for this set of tests ( upwards of two minutes ) after doing some debugging it looked like it was the waitingForVisible lines. I removed 'um to see if they had any effect on the speed ( they seemed to drop the runtime into the 10s of seconds ). After re-running these tests a number of times I only had the 2+ minute happen once. Every other run was 15-30 seconds.

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
ryanrath: see prior comments on expotDialog.js. I'll re-add.

html/internal_dashboard/js/admin_panel/AclGrid.js
jpwhite4: Where is it accessed elsewhere in the code? I cannot see any code that accesses this new member variable.

open_xdmod/modules/xdmod/integration_tests/lib/Controllers/UsageExplorerTest.php
ryanrath: 'nother dev change, reversion incoming.

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
ryanrath: duly noted, I'll refactor this into the page object. Commit incoming

html/controllers/user_admin/create_user.php
ryanrath: This was added to catch a case that occurs in OpenXDMoD in which there will be no centers supplied w/ 'cd', 'cs' ( due to being OpenXDMoD and there being only one center they can possibly pick ). But I do think, after looking at this, that it wouldn't hurt to add an additional check to the `elseif` that ensures that multipleServiceProviders === true.

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: lol, I'll update the comment.

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: Lines: 65, 66, 100, 109, 243, 244, 292, 323, 338

jpwhite4: the "this" on line 65 is the Ext.grid.CheckColumn object. The 'grid' variable inside the XDMoD.Admin.AclGrid.CenterSelector is not accessed on this line. Same for the others.

ryanrath: Ok, then I retract my initial statement and instead say that it was changes made in the course of debugging. Since there is no functional difference between the original and current state of the code the update obviously should not have been included. My apologies for rushing my initial assessment to both @chakrabortyr and @jpwhite4

classes/Models/Services/Centers.php
ryanrath: per our conversation, I will queue these changes up immediately post-merge.

classes/XDAdmin.php
```diff
@@ -109,14 +109,14 @@ public function updateAccountRequestStatus($id = -1, $creator = '')
...
-    JOIN acl_types aclt 
-        ON aclt.acl_type_id = a.acl_type_id 
-    ORDER BY a.display
+SELECT a.*,
+  req.acl_id IS NOT NULL AS requires_center
+FROM acls a LEFT JOIN
+  (
+    SELECT a2.acl_id 
+    FROM acls a2 WHERE a2.name IN ('cs','cd')
+  ) req ON req.acl_id = a.acl_id
```
jpwhite4: FYI this same query can be simplified to:

SELECT a.*, 
   CASE WHEN name = 'cd' THEN 1 WHEN name = 'cs' THEN 1 ELSE 0 END as requires_center 
FROM acls a ORDER BY a.display;

html/controllers/user_admin/update_user.php
```diff
@@ -123,7 +163,7 @@
          $user_to_update->disassociateWithInstitution();
       }
       else {
-         $user_to_update->setInstitution($_POST['institution'], array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $role_config));
+         $user_to_update->setInstitution($_POST['institution'], array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls));
```
jpwhite4: The $acls variable my not be set at this point since it is only created inside an earlier conditional statement. Also in the current code the $role_config variable contains exactly the same thing as the $acls variable (lines 62 and 106 are redundant).

html/internal_dashboard/js/admin_panel/AclGrid.js
```diff
@@ -85,7 +97,9 @@ XDMoD.Admin.AclGrid = Ext.extend(Ext.grid.EditorGridPanel, {
              */
             reset: function () {
                 this.grid.store.each(function (record) {
-                    record.set(this.dataIndex, false);
+                    if (record.get(this.dataIndex) !== false) {
```
jpwhite4: Note that this new code check is not needed since the ExtJs record set function already does this. (so in fact this ``optimization'' will be slower than the old code.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -526,6 +528,18 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
                     CCR.xdmod.ui.userManagementMessage('This user must have a type associated with it.', false);
                     return;
                 }
+                // these have their own variable so that they can be easily
+                // swapped out for the results of a call to the backend to retrieve
+                // the current list of feature acls. And to make it clear what this
+                // array of strings represents.
+                var featureAcls = ['dev', 'mgr'];
```
jpwhite4: You should not hard code these strings here - the list of non-flag ACLs is already available in the CCR.xdmod.UserTypes object: `Object.values(CCR.xdmod.UserTypes)`

html/controllers/user_admin/update_user.php
ryanrath: While `$acls` is defined in an earlier conditional statement, if that condition is not met the request errors out. So if the request processing get's to the line 166 then it will have had to pass the conditional on line 144 successfully ( and $acls will be set ). 
**NOTE:** *Unless I've missed something, which is entirely possible*

I agree that `$role_config` and `$acls` contain exactly the same information and as long as it's not important to maintain the order in which validation occurs then it could certainly be moved / merged with the code block starting at 104.

html/internal_dashboard/js/admin_panel/AclGrid.js
ryanrath: Yeah I've got nothing on why I added this bit of code. My best guess was that it was left in while trying to debug the problem that 00a47e1d122d0fc340c5682cb100b34ee07a2bb5 was supposed to fix. Definitely not an 'optimization' though. Commit incoming w/ this change.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: Agreed, and thanks for pointing out `CCR.xdmod.UserTypes` I hadn't seen that one before. Update committed.

jpwhite4:
Uncaught TypeError: Cannot read property 'value' of null
    at fn (SectionExistingUsers.js:565)
    at ext-base.js:1

seen when trying to disable an existing user account

html/controllers/user_admin/update_user.php
jpwhite4: You cannot exit this controller if POST['acls'] is not set - this will break the user disable/enable code and probably some other stuff too.

ryanrath:
And now I know why I added that 'id' field to the existingUserEmailField... Resolved in the latest commit

smgallo:
When trying to disable a user on xdmod-dev, I get the following error as well:
```
<br />
<b>Notice</b>:  Undefined offset: 0 in <b>/data/www/xdmod/share/classes/XDUser.php</b> on line <b>1113</b><br />
{"success":false,"status":"SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'role_id' cannot be null"}
```

ryanrath:
Only related in that it's resolved in the changed code base.

html/controllers/user_admin/update_user.php
ryanrath: It will unless update_user.php is updated so that we only check the acls if 'is_active' is not set ( we're not enabling / disabling a user ). 

Some research: 
A search for 'update_user' ( executed from the root of an xdmod working copy w/ modules checked out to open_xdmod/moduels/*)
- grep -iRn update_user --exclude-dir=.git --exclude-dir=.idea

```
open_xdmod/modules/xdmod/integration_tests/lib/Controllers/ControllerTest.php:438:            'operation' => 'update_user',
open_xdmod/modules/xdmod/integration_tests/lib/Controllers/BaseUserAdminTest.php:262:            $this->getTestFiles()->getFile('user_admin', 'test.update_user')
open_xdmod/modules/xdmod/integration_tests/lib/Controllers/UserAdminTest.php:227:            'operation' => 'update_user',
html/internal_dashboard/js/admin_panel/SectionExistingUsers.js:446:                        objParams.operation == 'update_user'
html/internal_dashboard/js/admin_panel/SectionExistingUsers.js:453:                        if (objParams.operation == 'update_user') {
html/internal_dashboard/js/admin_panel/SectionExistingUsers.js:564:                            operation: 'update_user',
html/internal_dashboard/js/admin_panel/SectionExistingUsers.js:813:                    operation: 'update_user',
html/controllers/user_admin/update_user.php:3:// Operation: user_admin->update_user
html/controllers/user_admin.php:8: *     update_user: uid + at least one of the following: first_name, last_name, email_address, roles, primary_role, assigned_user, is_active
html/controllers/user_admin.php:26:$controller->registerOperation('update_user');
```

These results show that the only front end files where this is found is in `SectionExistingUsers.js`

Of the four hits in this file, two of them are concerned with what to do after a user management action has been taken
- line 446 and 453

That leaves line 564 and 813
- Line 813 is in the 'Save Changes' button handler ( Updating a User ) 
- Line 564 is the enable / disable user code

So it seems, at least on initial inspection, that the two use cases for `update_user.php` is when updating a user via the 'Save Changes' button and Enabling / Disabling a user. And as long as those two cases are covered in `update_user.php` we should be good. 

Of course if there's some other interaction / use case that I didn't uncover during my research. Please let me know and I'll make the appropriate change right away.

Thank you again for taking the time to look through everything

jpwhite4: The latest code still has the problem that if POST['acls'] is not set but POST['institution'] and 'active' are then it will reference an undefined variable

jpwhite4:
Something funny is going on with the user status message text:

Old code firefox text shows and is colour coded (green for active):
image

New code on firefox text appears in black only:

image

New code on chrome no text appears at all:

image

ryanrath:
Hi @jpwhite4, I wasn't able to reproduce the no status text but I was able to reproduce the non-colorization of the status text. The issue was ( at least while exploring this on OSX Chrome ) that the Component returned via `Ext.getCmp()` did not have a `el` property. This meant that when the code attempted to add / remove the classes that colored the status text no actual changes were made to the underlying dom element.

Which OS did you experience the missing status text? Just want to make sure I'm testing with the same setup.

Thanks,

Ryan

ryanrath:
@jpwhite4 Ok, I was finally able to reproduce the missing user status. I'm not sure if it was via the same set of steps that you took but for me it involved utilizing the "Create / Manage" button on both the "XDMoD Account Requests" and the "Existing Users" tab. Which ever tab is utilized second is the one that fails to display the status text. This seems to be due to the fact that there is a separate instantiation of SectionExistingUsers.js for each tab which doesn't play well with the `id: 'txtAccountStatus'` definition of the status text. 

I've actually run into other issues when going back and forth between the two tabs "Create & Manage Users". In particular, when switching between each of the parent tabs and switching between the "New User" & "Modify User" child tabs the layout of the window can get all sorts of messed up.

PR #323, opened 2017-11-07
Tool for copying filters between metric explorer charts

smgallo:
## Description

While running an analysis I found that I had several charts with the same set of 20+ filters. Rather than need to add each of these filter sets to multiple charts and, in some cases multiple data series per chart, I developed a tool to provide a developer with the ability to copy filters from one chart to another.  Given a particular user, the following scenarios are supported:
- List all saved metric explorer charts
- List the global filters for all charts (or a specific chart)
- List the local filters for each data series for all charts (or a specific chart)
- Copy filters between charts and/or data series in a chart. This includes
    - Copy global filters from one chart to another
    - Copy global filters from one chart to a data series in another (or the same) chart
    - Copy data series filters to another data series (in the same chart or another)
    - Copy data series filters to the global chart filters of another (or the same) chart

Note that filters are checked so duplicates are not added.

Chart and series identifiers are discoverable using this tool:
```
php manage_saved_charts.php -r ~/xdmod-gw-7.1-current -u smgallo -l chart-names
chart[43] WL - GW Usage
chart[44] WL - Core Count by Gateway
chart[45] WL - Jobs Ended by Gateway
chart[47] WL - GW Usage sans Zhanglab
chart[48] WL - GW Comm User vs Allocation
chart[49] WL - GW Comm User vs Allocation 2
chart[50] Test Chart
```

```
php manage_saved_charts.php -r ~/xdmod-gw-7.1-current -u smgallo -l series -s 37
chart[37] CPU Hours: Total by Queue series[0] Jobs: total_cpu_hours by queue
chart[37] CPU Hours: Total by Queue series[1] Jobs: total_su by queue
```
To copy the filters from series 3 in chart 49 to the global filters of chart 50 use
```
php manage_saved_charts.php -r ~/xdmod-gw-7.1-current -u smgallo -o copy-filters --source 37,2 --destination 50 -v
```

```
php manage_saved_charts.php -h
        Usage: manage_saved_charts.php

        -d <chart_id>[,<series_id>], --destination <chart_id>[,<series_id>]
        The destination for the specified operation. Only used if an operation has been specified.
        If only a chart identifier has been provided then global options for that chart will be
        used as the destination. If a series has been provided then options for that chart series
        will be used as the destination. The identifiers can be obtained using the --list option.

        -l <item>, --list <item>
        Display a list of the various items. If a source chart id was specified list only values
        for that chart.
        Supported items are: chart-names, series, global-filters, local-filters

        -o <operation>, --operation <operation>
        Operation to perform. Supported operations are:
        copy-filters: Copy the filters from the source to the destination

        -r <path>, --xdmod-root <path>
        The XDMoD root path where we will find share/configuration/linker.php

        -s <chart_id>[,<series_id>], --source <chart_id>[,<series_id>]
        The source for the specified operation. Only used if an operation has been specified.
        If only a chart identifier has been provided then global options for that chart will be
        used as the source. If a series has been provided then options for that chart series will
        be used as the source. The identifiers can be obtained using the --list option.

        -t, --dry-run
        Operate in DRYRUN mode and do not modify the database.

        -u <username>, --user <username>
        The user whose profile will be loaded.

        -v, --verbose
        Provide verbose output.
```
## Motivation and Context

Got tired of typing 24+ filter names into the dialog for multiple charts and data series.

## Tests performed

Tested the following cases:

## Types of changes

## Checklist:

### DIFF
--- a/tools/dev/manage_saved_charts.php
+++ b/tools/dev/manage_saved_charts.php
@@ -476,4 +476,3 @@ function generateFilterName(array $filter)
 {
     return sprintf("(%s) %s = %s", $filter['id'], $filter['dimension_id'], $filter['value_name']);
 }
-?>
--- a/tools/dev/manage_saved_charts.php
+++ b/tools/dev/manage_saved_charts.php
@@ -1,0 +1,479 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Saved metric explorer charts are stored in each user's profile. The profile is stored as a
+ * serialized PHP object with a 'queries_store' property containing a 'data' array that stores
+ * objects describing each chart. The 'data' array contains the chart name and a 'config'
+ * property containing an encoded JSON representation of the chart configuration. For example:
+ *
+ * {
+ *   'queries_store' => {
+ *     'data' => [
+ *       {
+ *         'name' => 'my chart',
+ *         'config' => '{"featured":false,"trend_line":false, ...'
+ *       }
+ *     ]
+ *   }
+ * }
+ *
+ * This tool provides a means to manipulate individual chart configurations and save them
+ * back to a user's profile.
+ */
+
+// Process arguments
+
+$scriptOptions = array(
+    'destination' => null,
+    'dryrun' => false,
+    'list' => null,
+    'operation' => null,
+    'source' => null,
+    'user' => null,
+    'verbose' => false,
+    'xdmod-root' => '/home/smgallo/xdmod/'
+);
+
+$supportedOperations = array(
+    'copy-filters'
+);
+
+$options = array(
+    'd:' => 'destination:',
+    'h'  => 'help',
+    'l:' => 'list:',
+    'o:' => 'operation:',
+    'r:' => 'xdmod-root:',
+    's:' => 'source:',
+    't'  => 'dry-run',
+    'u:' => 'user:',
+    'v'  => 'verbose'
+);
+
+$args = getopt(implode('', array_keys($options)), $options);
+
+foreach ($args as $arg => $value) {
+    switch ($arg) {
+        case 'd':
+        case 'destination':
+            $scriptOptions['destination'] = $value;
+            break;
+
+        case 'h':
+        case 'help':
+            usage_and_exit();
+            break;
+
+        case 'l':
+        case 'list':
+            $scriptOptions['list'] = $value;
+            break;
+
+        case 'o':
+        case 'operation':
+            $scriptOptions['operation'] = $value;
+            break;
+
+        case 'r':
+        case 'xdmod-root':
+            $scriptOptions['xdmod-root'] = $value;
+            break;
+
+        case 's':
+        case 'source':
+            $scriptOptions['source'] = $value;
+            break;
+
+        case 't':
+        case 'dry-run':
+            $scriptOptions['dryrun'] = true;
+            break;
+
+        case 'u':
+        case 'user':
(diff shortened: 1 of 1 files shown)


jpwhite4:
Looks good to me. I note that you could modify this code to hit the rest interface to access the charts rather than the internal API call, not sure if there is a good use case for this except as part of a command-line interface to XDMoD. 

P.S I'm probably going to need a script/mechanism that automatically exports all metrics explorere charts for a user. I also want an automatic way to generate plots that have multiple dataseries each showing the same metric grouped by a given dimension, but each dataseries has a filter applied to show one set of values for a different dimension.(E.g. Chart showing XD SUs by CPU Usage with a different dataset for each HPC resource).

smgallo:
I thought about using the REST interface but decided on the internal API since I had that part of this tool already kicking around. I can modify it to dump the charts and make the modifications that you need. Those would be a good use case for hitting the REST endpoints to select available options.

tools/dev/manage_saved_charts.php
```diff
@@ -0,0 +1,479 @@
...
+
+$scriptOptions = array(
+    'destination' => null,
+    'dryrun' => false,
+    'list' => null,
+    'operation' => null,
+    'source' => null,
+    'user' => null,
+    'verbose' => false,
+    'xdmod-root' => '/home/smgallo/xdmod/'
```
plessbd: we should default this to `/usr/share/xdmod/`

tools/dev/manage_saved_charts.php
```diff
@@ -0,0 +1,479 @@
...
+require_once $linker;
+
+if ( $scriptOptions['dryrun'] ) {
+    print "Running in DRYRUN mode" . PHP_EOL;
+}
+
+// --------------------------------------------------------------------------------
+// Verify user
+
+if ( $scriptOptions['verbose'] ) {
```
plessbd: instead of doing this 9 times, maybe create a "log" function that you just pass the message to and it does the check.

tools/dev/manage_saved_charts.php
```diff
@@ -0,0 +1,479 @@
...
+ * @param array $filter An associative array based on the decoded JSON object representing a chart filter.
+ *
+ * @return string The filter name.
+ */
+
+function generateFilterName(array $filter)
+{
+    return sprintf("(%s) %s = %s", $filter['id'], $filter['dimension_id'], $filter['value_name']);
+}
+?>
```
plessbd: remove closing tag

tools/dev/manage_saved_charts.php
smgallo: In this case it would mean things may get computed when they are not needed. For example,
```
                if ( $scriptOptions['verbose'] ) {
                    print "Destination filters:" . PHP_EOL;
                    foreach ($destFilterList as $filterIndex => $filter) {
                        print sprintf(
                            "filter[%d] %s",
                            $filterIndex,
                            generateFilterName($filter)
                        ) . PHP_EOL;
                    }
                }

```

and

```
            if ( $scriptOptions['verbose'] && count($filtersToAdd) > 0 ) {
                $filterNames = array_reduce(
                    $filtersToAdd,
                    function ($carry, array $filter) {
                        return $carry . PHP_EOL . generateFilterName($filter);
                    },
                    ''
                );
                print sprintf("Adding %d filters:%s", count($filtersToAdd), $filterNames) . PHP_EOL;
            }
```

PR #324, opened 2017-11-08
Completing the addition of GroupByProvider to OpenXDMoD

ryanrath:
## Description
- The GroupByProvider.php class already exists in OpenXDMoD but does not exist
  in datawarehouse.json. This caused a problem replicating the behavior of
  records being inserted into UserRoleParameters for users being assigned the
  Center Director or Center Staff. The reason for this problem is that the
  param_name column in UserRoleParameters holds what amounts to hard coded
  group_by names, while in the new user_acl_group_by_parameters the equivalent
  column is group_by_id ( an actual fk to the group_by table ). The problem then
  becomes, how do you create a reference to a group_by that doesn't exist? Well,
  you don't, hence the creation of this PR.

  We also ensure that the user experience doesn't change by *not* adding a
  record to roles.json ( a user needs an entry in query_descripters to have any
  access to a realms group_by ). This was per a conversation with @jtpalmer. 

- **NOTE:**
  Other Updates that support these changes:
  - xdmod-test-artifacts: files were added to support the ConfigTest changes.

## Motivation and Context
To ensure that the new acl framework provides the same behaviors as Roles.

## Tests performed
The following Tests were performed on: OpenXDMoD, OpenXDMoD w/ App Kernels, SUPREMM
- component_tests
- integration_tests
- automated_tests
- manual inspection of the db tables: 
```sql
SELECT COUNT(*) as count, 'UserRoleParameters' as 'table'
FROM UserRoleParameters
UNION ALL
SELECT COUNT(*) as count, 'user_acl_group_by_parameters' as 'table'
FROM user_acl_group_by_parameters
```
*Ensure that the counts of both tables are the same*
```sql
SELECT urp.*, uagbp.user_acl_parameter_id IS NOT NULL AS 'exists'
FROM UserRoleParameters urp
  JOIN Roles r ON r.role_id = urp.role_id
  LEFT JOIN acls a ON a.name = r.abbrev
LEFT JOIN user_acl_group_by_parameters uagbp
  ON uagbp.user_id = urp.user_id AND
     uagbp.acl_id = a.acl_id AND 
     uagbp.value = urp.param_value;
```
*Ensure that the content in UserRoleParameters is present in user_acl_group_by_parameters*

## Types of changes

## Checklist:

### DIFF
--- a/classes/OpenXdmod/Migration/Version701To710/ConfigFilesMigration.php
+++ b/classes/OpenXdmod/Migration/Version701To710/ConfigFilesMigration.php
@@ -5,6 +5,7 @@
 
 namespace OpenXdmod\Migration\Version701To710;
 
+use CCR\Json;
 use OpenXdmod\Migration\ConfigFilesMigration as AbstractConfigFilesMigration;
 
 /**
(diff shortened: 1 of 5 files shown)


jtpalmer:
Is there already a configuration file migration for the changes to `datawarehouse.json`?

ryanrath:
There is not. I'll add that right now. Thanks!

ryanrath:
@jtpalmer section added to ConfigFilesMigration & tested w/ docker via upgrade.

PR #325, opened 2017-11-08
XDUser Construction: Primary Role

ryanrath:
## Description
- A slight change that modifies XDUser so that during construction it honors the
  `primary_role` parameter which is used to populate the `activeRole` property.
- Updated the create_user.php to take this change into account ( it now provides
  ROLE_ID_USER )

## Motivation and Context
Ensure that current uses of XDUser::_construct do not break and behave as intended.

## Tests performed
Added a component test to ensure we test for the desired behavior implemented by this change.
- i.e. that the primaryRole supplied during XDUser construction is taken into consideration when retrieving / populating the activeRole. 

## Types of changes

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -465,47 +465,13 @@ class XDUser implements JsonSerializable
         // See: http://stackoverflow.com/questions/1197005/how-to-get-numeric-types-from-mysql-using-pdo
         $user->_user_type = (int)$userCheck[0]['user_type'];
 
-        $user->_roles = array();
+        // We retrieve the most privileged acl for this user and use it for the
+        // active / primary role.
+        $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($user);
+        $activeRoleFormalName = self::_getFormalRoleName($mostPrivilegedAcl->getName());
 
-        $rolesResult = $pdo->query("
-         SELECT 
-            r.abbrev,
-            r.description,
-            IF(ur.is_primary, COALESCE(urp.is_primary, ur.is_primary), ur.is_primary) AS is_primary,
-            IF(ur.is_active, COALESCE(urp.is_active, ur.is_active), ur.is_active) AS is_active,
-            urp.param_value
-         FROM
-            UserRoles AS ur
-            JOIN Roles AS r ON ur.role_id = r.role_id
-            LEFT JOIN UserRoleParameters AS urp ON ur.user_id = urp.user_id AND ur.role_id = urp.role_id
-         WHERE ur.user_id = :user_id
-      ", array(
-            ':user_id' => $user->_id,
-        ));
-
-        foreach ($rolesResult as $roleSet) {
-
-            if (!in_array($roleSet['abbrev'], $user->_roles)) {
-                $user->_roles[] = $roleSet['abbrev'];
-            }
-
-            if ($roleSet['is_active'] == '1') {
-                $user->_active_role = \User\aRole::factory($roleSet['description']);
-                $user->_active_role->configure($user, $roleSet['param_value']);
-            }
-
-        }//foreach
-
-        if (!isset($user->_active_role)) {
-            $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($user);
-            $activeRoleFormalName = self::_getFormalRoleName($mostPrivilegedAcl->getName());
-
-            $user->_active_role = aRole::factory($activeRoleFormalName);
-            $user->_active_role->configure($user);
-
-            $user->_primary_role = aRole::factory($activeRoleFormalName);
-            $user->_primary_role->configure($user);
-        }
+        $user->_primary_role = $user->_active_role = aRole::factory($activeRoleFormalName);
+        $user->_active_role->configure($user);
 
         // BEGIN: ACL population
         $query = <<<SQL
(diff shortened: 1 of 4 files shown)


html/controllers/user_admin/create_user.php
```diff
@@ -41,7 +41,7 @@
         '',
         $_POST['last_name'],
         array_keys($acls),
-        null,
+        ROLE_ID_USER,
```
jpwhite4: What happens if you create a user account that does not have the "usr" role? This would happen if you do not check the "User" check box in the internal dashboard GUI.

ryanrath: Then the user would be created without an association to the 'usr' role. 

Now, when the user is saved it would attempt to Update a record in UserRoles, if one exists, such that record.user_id = newUser.id AND record.role_id = activeRole.role_id ( in this case 'usr' ). But that attempt would fail as there won't be a record for this user / 'usr' role. 

Caveat, for the lifetime of this user object any calls to `$user->getActiveRole()` will return the 'usr' role. But seeing as how it goes out of scope at the end of create_user.php I don't imagine this is an issue.
  - **Note:** Nor is `$user->getActiveRole()` used in any of the functions called after the user is saved: ` $newuser->addAclOrganization`, `$newuser->setInstitution`, `$newuser->getUsername`, `$newuser->getPasswordLastUpdatedTimestamp`

jpwhite4:
Please can you explain in a bit more detail why this change is needed. You state that the motivation is to 'Ensure that current uses of XDUser::_construct do not break and behave as intended.' Please can you explain what is broken and why.

ryanrath:
@jpwhite4 I've added the changes that we talked about the other day in regards to `activeRole`. I've tested it out on OpenXDMoD and everything passed. I'm going to go ahead and deploy to my xdmod-dev port and test things out there (still working on that docker image).

ryanrath:
@smgallo Just a quick update the `_primary_role` changes have been added and I've updated the XDUser component Tests so that they can be run on an XSEDE system.

classes/XDUser.php
```diff
@@ -496,6 +497,17 @@ public static function getUserByID($uid, &$targetInstance = NULL)
 
         }//foreach
 
+        if (!isset($user->_active_role)) {
+            $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($user);
+            $activeRoleFormalName = self::_getFormalRoleName($mostPrivilegedAcl->getName());
+
+            $user->_active_role = aRole::factory($activeRoleFormalName);
```
smgallo: There are already methods `setActiveRole()` and `setPrimaryRole()`. Why are these not called here?

classes/XDUser.php
```diff
@@ -496,6 +497,17 @@ public static function getUserByID($uid, &$targetInstance = NULL)
 
         }//foreach
 
+        if (!isset($user->_active_role)) {
```
smgallo: In the loop above this block of code, `$_active_role` is being set based on data in the UserRoles, UserRolesParameters, and Roles tables. Here, if it didn't get set in the loop above it is being set based on the return value from `Acls::getMostPrivilegedAcl()`. Are we mixing the old Roles and new ACL infrastructure? That sounds like a recipe for problems...

classes/XDUser.php
```diff
@@ -962,21 +974,49 @@ public function saveUser()
...
+        // Retrieve this users most privileged acl as it will be used to set the
+        // the _active_role property.
+        $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($this);
 
-        $active_role_id = $this->_getRoleID($this->_active_role->getIdentifier());
+        // NOTE: It is possible that a user may not have any acls, in this case
+        // there cannot be a most privileged one and so we test to make sure
+        // that an acl was returned before utilizing it.
+        if (isset($mostPrivilegedAcl)) {
+            $activeRoleName = $this->_getFormalRoleName($mostPrivilegedAcl->getName());
```
smgallo: `getFormalRoleName()` is a static method yet here we are calling it in object context. Why?

classes/XDUser.php
```diff
@@ -962,21 +974,49 @@ public function saveUser()
             );
         }
 
-        // If the updater (e.g. Manager) has pulled out the (recently) active role for this user, reassign the active role to the primary role.
+        // Retrieve this users most privileged acl as it will be used to set the
+        // the _active_role property.
+        $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($this);
 
-        $active_role_id = $this->_getRoleID($this->_active_role->getIdentifier());
+        // NOTE: It is possible that a user may not have any acls, in this case
```
smgallo: If the user has no ACLs, can they see anything at all? Is there a "Public" ACL that is the default? If so then won't all users start with "Public" and be elevated if they have ACLs in the database?

classes/XDUser.php
```diff
@@ -155,7 +156,7 @@ function __construct($username = NULL, $password = NULL, $email_address = NO_EMA
 
         // =================================
 
-        $primary_role_name = $this->_getFormalRoleName(ROLE_ID_USER);
+        $primary_role_name = $this->_getFormalRoleName($primary_role);
```
smgallo: The constructor sets `$_active_role` on line 165 but it does not set the primary role. I thought they going to both be the same?

classes/XDUser.php
```diff
@@ -2121,6 +2161,13 @@ public function getActiveRole()
             throw new Exception('You must call saveUser() on this newly created XDUser prior to using getActiveRole()');
         }
 
+        if (!isset($this->_active_role)) {
```
smgallo: It looks like there are 2 ways to create a user, via `__constructor()` and `getUserByID()` both of which set the active user. Why is it being set again here?

classes/XDUser.php
ryanrath: A user cannot be created w no acls. It's prevented on both the front end ( via a a validation check in SectionNewUser.js@ line 499 ) and on the backend via a check in `XDUser::__construct` that throws an exception if the number of items in the acl array is 0. Also, you're not able to assign the 'Public' acl to a user via the internal_dashboard. Nor are you able to modify the 'Public' user.

classes/XDUser.php
ryanrath: This was an oversight, it hadn't been static prior to this PR.

classes/XDUser.php
smgallo: If they must always have an ACL, then they will always have a most privileged ACL, correct? If that's the case should you not be throwing an exception here? And if they always have an ACL why check to see if the active and primary roles are set, shouldn't they be set based on the ACL? See line 981 "NOTE: It is possible that a user may not have any acls" which is why I asked.

classes/XDUser.php
ryanrath: yeah, I vaguely remember there being a reason for this... but I think that was before we setup the active / primary role code in the constructor / getUserByID

classes/XDUser.php
ryanrath: @smgallo This should be updated in the most recent commit

smgallo:
`_getFormalRoleName()` is still being called in object context (`$this->_getFormalRoleName()`), in static context as it should be now that it is a static method, and also as `XDUser::_getFormalRoleName()`. These need to be fixed.

ryanrath:
@smgallo those calls have been updated in the latest commit. I also found one other instance of a `static` function being called via `$this` ( in XDUser::reloadUser ). I'll start a new PR to get that change in.

smgallo:
Still  have one instance being called as `XDUser::_getFormalRoleName()` on line 2173. Any reason for this not to be `static::`?

PR #326, opened 2017-11-08
Update XDUser::enumAllAvailableRoles

ryanrath:
## Description
Updated this XDUser function to use 1 query as opposed to 6. Besides reducing the number of queries run it also means that we can add additional acls without adding code / sql to this function. 

## Motivation and Context
shrinking / simplifying the way in which this function accomplishes it's goal.

## Tests performed
Added user specific tests for both `enumAllAvailableRoles` as well as the functions that use `enumAllAvailableRoles`
- `getMostPrivilegedRole`
- `getAllRoles`
-  ~~`getRoleCategories`~~ NOTE: this function is only called once and the caller itself is never called.

## Types of changes

## Checklist:


ryanrath:
NOTE: the shippable tests failed because the changes from https://github.com/ubccr/xdmod/pull/324 have not been integrated yet.

classes/XDUser.php
```diff
@@ -1338,73 +1338,49 @@ public function enumAllAvailableRoles()
...
+              SELECT
+                ah.acl_id,
+                ah.level
+              FROM acl_hierarchies ah
+                JOIN hierarchies h
+                  ON ah.hierarchy_id = h.hierarchy_id
+                JOIN modules m
+                  ON h.module_id = m.module_id
+              WHERE h.name = :acl_hierarchy_name
+                    AND m.name = :module_name
```
jpwhite4: Why is the module name included in this query. Surely the module name is irrelevant at this point in the application logic.

ryanrath: ~~So the answer is 'it depends', if we only ever want to support one acl_hiearchy then this can be removed. If at some point we want to allow for multiple acl_hiearchies ( which the db tables support ) that have the same name then we'd want to have the module supplied. ~~

After thinking about it, it's only ever required if we want to support acl_hiearchies that have the same name but were added by different modules. So: 
- Module: XDMoD
  - acl_hiearchy.name: acl_hierarchy
- Module: <Company X>
  - acl_hiearchy.name: acl_hiearchy
If I removed the `:module_name` and associated joins then we'd run into problems. But really all that would have to happen is that the acl_hiearchy record from <Company X> would need to be named something different.

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -824,15 +824,32 @@ public function testGetRoleWithNull()
...
+        $user = XDUser::getUserByUserName($userName);
 
         $allAvailableRoles = $user->enumAllAvailableRoles();
         $this->assertEquals($expected, $allAvailableRoles);
     }
 
+    public function provideEnumAllAvailableRoles()
+    {
+        return array(
+            array(self::CENTER_DIRECTOR_USER_NAME, 'center_director_all_available_roles.json'),
```
chakrabortyr: Is there a better way to do this than hardcoding filenames in here?

ryanrath: This can be something that's addressed in a future PR. But thank you for the question.

smgallo: For the component tests I think it's OK to hardcode the file names.

jpwhite4: These tests only check user accounts that have one or two roles. There are no tests for accounts that have more than two roles, nor are there tests for accounts that have role modifiers set (such as manager or developer). Please can you test these other possibilities too.

classes/XDUser.php
```diff
@@ -1352,73 +1352,98 @@ public function enumAllAvailableRoles()
...
-            $this->_pdo->query($role_query_1, $role_query_1_params),
-            $this->_pdo->query($role_query_2, $role_query_2_params),
-            $this->_pdo->query($role_query_3, $role_query_3_params),
-            $this->_pdo->query($role_query_4, $role_query_4_params),
-            $this->_pdo->query($role_query_5, $role_query_5_params),
-            $this->_pdo->query($role_query_6, $role_query_6_params)
+        // NOTE: previously we had no DB concept of modules / realms
+        // the values that are provided for :module_name, :realm_name, and
+        // :group_by_name simulate the behavior of the old system.
+        $available_roles = $this->_pdo->query(
```
smgallo: The calling code that I found in XDUser.php, html/index.php, and  html/controllers/multi_role_manager/enum_active_roles.php does not try/catch PDOExceptions so you should handle them here.

classes/XDUser.php
```diff
@@ -1511,7 +1538,8 @@ public function setInstitution($institution_id, $is_primary = false)
         $this->_pdo->execute($aclInsert, array(
```
smgallo: I found `setInstitution()` used in several places that did not try/catch a `PDOException` (html/controllers/user_admin/create_user.php and classes/XDUser.php) so exceptions need to be checked here.

classes/XDUser.php
```diff
@@ -1352,73 +1352,98 @@ public function enumAllAvailableRoles()
...
-
-        $role_query_6 = "SELECT r.description, r.abbrev AS param_value, urp.is_primary, urp.is_active " .
-            "FROM moddb.UserRoles AS urp, moddb.Roles AS r " .
-            "WHERE r.role_id = urp.role_id AND user_id=:user_id " .
-            "AND r.description = 'Public'";
-        $role_query_6_params = array(
-            ':user_id' => $this->_id,
-        );
+        $query = <<<SQL
+SELECT DISTINCT
```
smgallo: I'd like to see inline SQL comments for each of the subqueries here to explain what they are doing and also a comment on the overall query describing what it's doing so someone looking at it later will have an easier time figuring it out.

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterDirectorRoleTest.php
```diff
@@ -67,14 +68,13 @@ public function testGetFormalNameWithNoUser()
...
-        $user = XDUser::getUserByUserName(XDUserTest::CENTER_DIRECTOR_USER_NAME);
+        $user = XDUser::getUserByUserName(self::CENTER_DIRECTOR_USER_NAME);
         $expected = 'Center Director';
         $cd = new CenterDirectorRole();
         $cd->configure($user);
 
         $actual =  $cd->getFormalName();
-        $found = strpos($actual, $expected);
-        $this->assertTrue($found !== false);
+        $this->assertNotFalse(strpos($actual, $expected), "Expected to find '$expected' in '$actual'");
```
smgallo: Why are you checking for "Center Director" in the returned string instead of an exact match? Please add a comment explaining this.

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterDirectorRoleTest.php
```diff
@@ -105,12 +105,12 @@ public function testEnumStaffMembersNoUser()
...
-        $expected = Json::loadFile(XDUserTest::getTestFile('center_director_staff_members.json'));
+        $expected = Json::loadFile($this->getTestFile('center_director_staff_members-update_enumAllAvailableRoles.json'));
 
-        $user = XDUser::getUserByUserName(XDUserTest::CENTER_DIRECTOR_USER_NAME);
+        $user = XDUser::getUserByUserName(self::CENTER_DIRECTOR_USER_NAME);
         $cd = new CenterDirectorRole();
         $cd->configure($user);
         $actual = $cd->enumCenterStaffMembers();
-        $this->assertEquals($expected, json_decode(json_encode($actual), true));
+        $this->assertEquals(count($expected), count($actual));
```
smgallo: I don't know the answer to this, but in addition to the number of staff members returned does it make sense to also check to see if they are the correct ones?

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterStaffRoleTest.php
```diff
@@ -21,8 +24,8 @@ public function testGetIdentifierAbsoluteNoUser()
 
     public function testGetIdentifierAbsolute()
     {
-        $expected = XDUserTest::CENTER_STAFF_ACL_NAME . ';1';
-        $user = XDUser::getUserByUserName(XDUserTest::CENTER_STAFF_USER_NAME);
+        $expected = self::CENTER_STAFF_ACL_NAME . ';1';
```
smgallo: Why is the ";1" added? Comments would be useful here.

configuration/roles.json
```diff
@@ -399,7 +399,7 @@
             "display": "Manager",
             "type": "data",
             "hierarchies": [{
-                "level": 301,
+                "level": 1,
```
jpwhite4: How come the level changed?

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -846,41 +819,149 @@ public function testGetRoleIDForValidRole()
...
-     */
     public function testGetRoleIDForInvalidRole()
     {
-        $user = XDUSer::getUserByUserName(self::CENTER_STAFF_USER_NAME);
-        $reflection = new ReflectionClass($user);
-        $method = $reflection->getMethod('_getRoleID');
-        $method->setAccessible(true);
-        $method->invoke($user, self::INVALID_ACL_NAME);
+        try {
+            $user = XDUSer::getUserByUserName(self::CENTER_STAFF_USER_NAME);
```
jpwhite4: s/XDUSer/XDUser/

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -846,41 +819,149 @@ public function testGetRoleIDForValidRole()
...
+                $results []= array(
+                    $userName,
+                    $fileName
+                );
+                $count += 1;
+            }
+        } catch(Exception $e) {
+            echo $e->getMessage();
+        }
+        $this->assertTrue(true);
```
jpwhite4: This line is not needed.

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -846,41 +819,149 @@ public function testGetRoleIDForValidRole()
...
+                $userName = $user->getUsername();
+                $fileName = implode('_', $aclCombination) . "_acls.json";
+                $results []= array(
+                    $userName,
+                    $fileName
+                );
+                $count += 1;
+            }
+        } catch(Exception $e) {
+            echo $e->getMessage();
```
jpwhite4: Is this code supposed to thow an exception? If so then use the test framework to check that the exception is thrown. If not then don't catch and ignore it.

classes/XDUser.php
```diff
@@ -1741,7 +1772,8 @@ public function setOrganizations($organization_ids = array(), $role = ROLE_ID_CE
                     ':user_id' => $this->_id,
                     ':acl_id' => $acl->getAclId(),
                     ':value' => $organization_id,
-                    ':module_name' => DEFAULT_MODULE_NAME
+                    ':module_name' => DEFAULT_MODULE_NAME,
+                    ':realm_name' => 'jobs'
```
jpwhite4: Please can you clarify why you are hard coding the jobs realm here. What about the SUPReMM realm data and the other realms in the XSEDE module?

classes/XDUser.php
ryanrath: Most definitely, I'll add those in there.

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterDirectorRoleTest.php
ryanrath: (np, I'll add that ) This is part of the effort to make more tests 'cross compatible' which in this context means on OpenXDMoD and XSEDE. Specifically, the 'centers' are going to be different on OpenXDMoD and XSEDE ( which is returned as a part of `getFormalName`. On XSEDE system it would look something like: `Center Director - TG`. While on OpenXDMoD w/ the ref db it will probably look like `Center Director - screw`  ).

classes/XDUser.php
jpwhite4: The question about the hardcoded jobs realm also applies to the code on line 1444

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterDirectorRoleTest.php
ryanrath: I would say yes but I'm not sure how to nail down which users the center director will be associated with ( due to the dynamic user creation / association w/ centers for users with the 'cd'/'cs' acl. I'm certainly open to suggestions.

jpwhite4: We fully control the entire test environment so we already know what users exist and what centers they are associated with.

open_xdmod/modules/xdmod/component_tests/lib/Roles/CenterStaffRoleTest.php
ryanrath: *nod* adding those now

ryanrath:
@smgallo Added those bits of documentation for you in 
- https://github.com/ubccr/xdmod/pull/326/commits/a4dd4777679ce4e4a57a3c58b7ebe533a02782c2
- https://github.com/ubccr/xdmod/pull/326/commits/aaf12640dab4b5d81ab0443506cb4152ca1a22bc

configuration/roles.json
ryanrath: ~After some talks w/ @smgallo about the ordering of acls w/ regards to hierarchical vs. non-hierarchical...~
Actually, thank you for this catch. It's not that the level should have changed. It's that the `type` should have changed ( to 'feature' ) and the 'hierarchies' property removed. I'll update that right now

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
ryanrath: yup thanks, this was code left in from dev.

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
ryanrath: Yup sorry, more dev code here. I was running into Exceptions when I was using this code to create the expected artifacts and threw a try-catch around it to figure out what was going on. I'll remove it.

classes/XDUser.php
ryanrath: Yup, I'll do my best to explain 

*( **WARNING** tired explanation / ramble follows, I can follow this up tomorrow morning with an in person explanation )*

So, we start with the goals:
- Make sure that whatever is done, the results "mean" the same thing as previously. ( i.e. a record in `UserRoleParameters` should "mean" the same thing as a record in `user_acl_group_by_parameters`).

To understand if the end result of setOrganizations as it applies to `user_acl_group_by_parameters` we need to understand
the similarities differences between it and it's parent `UserRoleParameters`. 

There's the obvious acl_id vs. role_id, the removal of is_primary, is_active, and param_op.
But the difference that matters in this case is UserRoleParameters.param_name == user_acl_group_by_parameters.group_by_id.

So first, why is user_acl_group_by_parameters.group_by_id == UserRoleParameters.param_name?

In looking at the distinct list of values in UserRoleParameters.param_name we find two: 'provider' and 'institution'.

Ok, so what do records in UserRoleParameters "mean"? 

Well, it's hard to deduce just from looking at the database but basically they're there to provide data filtering. In particular the 'param_name' is just a flag that is used by various 'GroupBy*' classes to determine which dynamic bit of sql it should add. It (a GroupBy) has no explicit run time concept of which 'module' or 'realm' ( a case could be made for runtime realm as that is contained in the directory structure ) it is associated with other then those provided implicitly by the directory structure it is contained within ( it does have an implicit dev time concept of module / realm based on file location ). 

So since there is no concept of which module / realm for a particular 'param_name' we run into a problem identifying which 'group_by' equates to param_name 'provider'.

Naively we could try: `SELECT group_by_id FROM group_bys where name = 'provider';`
The problem with that is two-fold. One, there can be more than one GroupByProvider in an XDMoD Install dependent on which modules you have installed. Two, within
a module you can have more then one 'group_by' with the same name if they are part of different realms.
A concrete example is that if you look at tas-db1 and run the following query:
```sql
SELECT 
  m.name module, 
  r.name realm, 
  gb.name group_by 
FROM group_bys gb 
  JOIN modules m ON m.module_id = gb.module_id 
  JOIN realms r ON r.realm_id = gb.realm_id 
WHERE  gb.name = 'provider';
```
you should see something like:
```
module,  realm,               group_by
xdmod,   jobs,                provider
xdmod,   resourceallocations, provider
supremm, supremm,             provider
```

So, since the original UserRoleParameters.param_name (or how it is used ) care not one jot for module / realm then as long as we can pin the reference to a group_by that produces the same 'provider' / 'institution' value then we should be set.

This is where the default values of module='xdmod' and realm='jobs' comes in. Since in XDMoD / Jobs realm we have a 'provider' group_by.

jpwhite4: The original code does have an implicit assumption that and group by with the same identifier applies to all realms. So then the UserRoleParameters specify 'provider' it actually means all group bys that have the 'provider' identifier in all realms (in the xsede case this is supremm, jobs and resourceallocations). Based on your explanation it looks like this is no longer honoured.

ryanrath:
@smgallo Alrighty, PDOException caught / sql logged utilizing the `ETL\Loggable` class. I also fixed up the Shippable automated tests as they apparently required another dependency be added to run.

configuration/etl/etl_sql.d/acls/xdmod/update_admin_users.sql
```diff
@@ -0,0 +1,31 @@
+-- =============================================================================
+-- NAME:      update_admin_users.sql
+-- EXECUTION: once on installation
+-- PURPOSE:   Ensure that any users that only have 'mgr' have 'usr' added.
```
smgallo: Does the admin UI also enforce this?

classes/User/Roles/DeveloperRole.php
```diff
@@ -0,0 +1,19 @@
...
+
+class DeveloperRole extends \User\aRole
+{
+    public function __construct()
+    {
+        parent::__construct(ROLE_ID_DEVELOPER);
+
+    }//__construct
+
+}//PublicRole
```
smgallo: Lets fix the comment here.

configuration/etl/etl_sql.d/acls/xdmod/update_admin_users.sql
ryanrath: It does now, I've added a new commit a2f7165d2b634d19c9b385c6a061c40a105c4d05 to https://github.com/ubccr/xdmod/pull/313 that enforces this on both the front and back end.

classes/XDUser.php
```diff
@@ -2223,7 +2223,14 @@ public function getMostPrivilegedRole()
         $mostPrivilegedAcl = Acls::getMostPrivilegedAcl($this);
         $roleName = self::_getFormalRoleName($mostPrivilegedAcl->getName());
         $role = aRole::factory($roleName);
-        $role->configure($this);
+
+        $organizationId = $mostPrivilegedAcl->getOrganizationId();
```
smgallo: Does `getOrganizationId()` return a value or `null`? If so, why not just call `$role->configure($this, $organizationId)` without the conditional? `configure` handles a `null` second parameter already.

configuration/etl/etl_sql.d/acls/xdmod/update_admin_users.sql
jpwhite4: The admin user account that is created in the `xdmod-setup` script does not have the 'usr' role because the users have not been ingested at the point in time that the admin user is created. How does this change impact this?

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
```diff
@@ -26,6 +26,7 @@ then
     xdmod-import-csv -t names -i $REF_DIR/names.csv
     xdmod-ingestor
     php /root/bin/createusers.php
+    acl-import
```
jpwhite4: How come you have to do an acl import here now?

ryanrath: This change is related to the file you commented on previously, `update_admin_users.sql`.

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -1362,6 +1441,81 @@ public function testGetFormalRoleNameEmptyString()
...
+    }
+
+    public function provideRoleGetParameters()
+    {
+        # SDSC: 856
+        # PSC: 848
+        # TACC: 476
+        $centerPermutations = $this->allCombinations(array(856, 848, 476));
+        $rolePermutations = $this->allCombinations(array('cd', 'cs'));
+        /*$rolePermutations = $this->allCombinations(array('cd', 'cs', 'cc'));*/
```
chakrabortyr: please remove if not needed.

ryanrath:
@smgallo @jpwhite4 I've pushed the changes we spoke about earlier this morning. 
- The default 'admin' user will now be created w/ both 'mgr' and 'usr' acls and a person_id of -1 ( as opposed to 0 ). 
- The default values for modw.person will be present both pre and post ingestion ( Unassociated (-2) and Unknown (-1) ). 
- Users that only have the 'mgr' acl will have the 'usr' role added. 
- Users that have an invalid person_id reference will have this value changed to -1 ( Unknown ). 

All of these changes apply / have been tested with both a fresh install and an upgrade.

classes/XDUser.php
```diff
@@ -2172,7 +2223,14 @@ public function getMostPrivilegedRole()
...
+        if (!isset($organizationId)) {
+            $role->configure($this);
+        } else {
+            $role->configure($this, $organizationId);
```
chakrabortyr: Why do we need the above line if it's okay to call this function with just one param? Looks like you could just do:
`$role->configure($this, $mostPrivilegedAcl->getOrganizationId());` and save some lines of code.

open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -1362,6 +1441,81 @@ public function testGetFormalRoleNameEmptyString()
...
+                $this->assertTrue(in_array($item, $expected), "Expected [". json_encode($expected) . "] Received: [" . json_encode($actual) . "]");
+            }
+        }
+    }
+
+    public function provideRoleGetParameters()
+    {
+        # SDSC: 856
+        # PSC: 848
+        # TACC: 476
```
jpwhite4: Please can you move the XSEDE specific stuff to the XSEDE module.

configuration/etl/etl_sql.d/acls/xdmod/update_admin_users.sql
chakrabortyr: ^^ What's the answer to the above?

ryanrath: Sorry for the delay in responding @chakrabortyr . To handle fresh installs ( `xdmod-setup` ) `classes/OpenXdmod/Setup/AdminUserSetup.php` was updated so that the admin account is now created with the 'usr' acl. This script will handle the upgrade scenario.

PR #328, opened 2017-11-08
Update _getFormalRoleName to use the acl table

ryanrath:
## Description
- Just another incremental PR for converting XDUser over to acls. 
- Also reduced the lines of code where I could.

## Motivation and Context
To continue migrating XDUser to acls.

## Tests performed
Tests have been included for this function as well as functions that utilize this function. They include: 
- `getMostPrivilegedRole`
- `getAllRoles`
- `setActiveRole`
- `__construct` ( via the testCreateUser* functions )

These tests reside within `component_tests/lib/XDUserTest.php`

## Types of changes

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -2583,7 +2583,8 @@ FROM (
        AND a.name = :abbrev;
 SQL;
 
-        $roleData = $pdo->query($query,
+        $roleData = $pdo->query(
+            $query,
             array(
                 ':abbrev' => $role_abbrev,
                 ':pub_abbrev' => ROLE_ID_PUBLIC
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -2566,20 +2566,29 @@ SQL;
 
     public function _getFormalRoleName($role_abbrev)
     {
-        $publicAclDisplay = 'Public';
-        if ($role_abbrev == ROLE_ID_PUBLIC || $role_abbrev == NULL) {
-            return $publicAclDisplay;
-        }
-
         $pdo = DB::factory('database');
+        $query = <<<SQL
+SELECT CASE WHEN a.acl_id IS NULL
+  THEN pub.description
+       ELSE a.display END description
+FROM (
+       SELECT
+         acl_id,
+         display AS description
+       FROM acls
+       WHERE name = :pub_abbrev
+     ) pub
+  LEFT JOIN acls a
+    ON a.acl_id != pub.acl_id
+       AND a.name = :abbrev;
+SQL;
 
-        $roleData = $pdo->query("SELECT display as description FROM acls WHERE name = :abbrev", array(
-            ':abbrev' => $role_abbrev,
-        ));
-
-        if (count($roleData) == 0) {
-            return $publicAclDisplay;
-        }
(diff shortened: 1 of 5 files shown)


jpwhite4:
The pull request description mentions that there are new tests - did you forget to git add the test file(s)?

ryanrath:
I was actually just thinking about that this morning. I didn't forget per se, I was actually a little over zealous last week and added the tests for a great many of the functions that would be refactored. The ones that cover this function are:
- XDUserTests.php::testGetFormalRoleName
  - test cases include: 'cd', 'cs', 'pi', 'usr', and 'pub'
  - **NOTE:** There will be a slight change to the `dataProvider` function. I'll be adding entries for null / empty string while removing: `XDUserTest.php::testGetFormalRoleNameNull()` and `XDUserTest.php::testGetFormalRoleNameEmptyString`

ryanrath:
@jpwhite4 My thought behind adding the tests first was to be able to assert that the tests passed before the changes.

classes/XDUser.php
```diff
@@ -2551,23 +2551,19 @@ private function _formalizeTimestamp($db_timestamp)
...
     {
-
-        if ($role_abbrev == ROLE_ID_PUBLIC) {
-            return 'Public';
-        }
-
-        if ($role_abbrev == NULL) {
-            return 'Public';
+        $publicAclDisplay = 'Public';
+        if ($role_abbrev == ROLE_ID_PUBLIC || $role_abbrev == NULL) {
```
smgallo: `NULL` should be `null` as per style guide.

classes/XDUser.php
```diff
@@ -2551,23 +2551,19 @@ private function _formalizeTimestamp($db_timestamp)
...
     public function _getFormalRoleName($role_abbrev)
     {
-
-        if ($role_abbrev == ROLE_ID_PUBLIC) {
-            return 'Public';
-        }
-
-        if ($role_abbrev == NULL) {
-            return 'Public';
+        $publicAclDisplay = 'Public';
```
smgallo: Why is `'Public'` hardcoded when there is already a public role in `moddb.Roles` (role_id = 9) and in the `moddb.acls` table (acl_id = 8)? Shouldn't this information be coming from the database? What if we update the database? Then we also need to update this function.

configuration/etl/etl_sql.d/acls/xdmod/create_public_user.sql
```diff
@@ -22,6 +22,9 @@ LEFT JOIN Users cur
         BINARY cur.email_address = BINARY inc.email_address
 WHERE cur.id IS NULL;
 
+-- Update public acl display
+UPDATE acls SET display = 'Public' WHERE name = 'pub';
+
```
smgallo: Don't forget that we'll also need a migration as well.

ryanrath: Yup, we have one. That's a big reason why I included this change here as opposed to somewhere else. This sql file is executed as part of the Migration from 701to710.

smgallo: OK, sounds good.

PR #334, opened 2017-11-22
update tests to catch and throw errors properly

plessbd:
The expect script can "gobble" the spawned processes exit status.  Lets not do that...


open_xdmod/modules/xdmod/integration_tests/scripts/xdmod-setup.tcl
```diff
@@ -133,3 +133,6 @@ confirmFileWrite yes
 enterToContinue
 
 selectMenuOption q
+
```
jpwhite4: Do you need an expect eof here?

plessbd: in this case i believe wait and eof will accomplish the same goals

PR #336, opened 2017-11-27
test updates

plessbd:
The webdriverio dependency was set to be ^4.9.4 (my bad) this allows it to be updated to newer versions, the newer versions have changed some things though and until we can go through the tests and make them work with the newer version just force it to be 4.9.4 and only that known version.  We should probably do this for all versions of things going forward.


jpwhite4:
Why does this fix work?

PR #337, opened 2017-11-28
add build-filter-list

plessbd:
add a build-filter-lists script  not currently used.  Will be used later.

### DIFF
--- a/bin/xdmod-build-filter-lists
+++ b/bin/xdmod-build-filter-lists
@@ -1,0 +1,153 @@
+#!/usr/bin/env php
+<?php
+
+require_once __DIR__ . '/../configuration/linker.php';
+
+use CCR\Log;
+// Disable memory limit.
+ini_set('memory_limit', -1);
+
+$opts = array(
+    array('h', 'help'),
+
+    // Logging levels.
+    array('v', 'verbose'),
+    array('d',  'debug'),
+    array('q', 'quiet'),
+
+    array('r::', 'realms::'),
+);
+
+$shortOptions = implode(
+    '',
+    array_map(
+        function ($opt) {
+            return $opt[0];
+        },
+        $opts
+    )
+);
+$longOptions = array_map(
+    function ($opt) {
+        return $opt[1];
+    },
+    $opts
+);
+
+$args = getopt($shortOptions, $longOptions);
+
+if ($args === false) {
+    fwrite(STDERR, "Failed to parse arguments\n");
+    exit(1);
+}
+if (array_key_exists('h', $args) || array_key_exists('help', $args)){
+    displayHelpText();
+    exit();
+}
+$realms = array();
+$logLevel = Log::NOTICE;
+foreach ($args as $key => $value) {
+    if (is_array($value) && 'v' !== $key) {
+        fwrite(STDERR, "Multiple values not allowed for '$key'\n");
+        exit(1);
+    }
+
+    switch ($key) {
+        case 'd':
+        case 'debug':
+            $logLevel = max($logLevel, Log::DEBUG);
+            break;
+        case 'q':
+        case 'quiet':
+            $logLevel = max($logLevel, Log::WARNING);
+            break;
+        case 'r':
+        case 'realms':
+            $realms = explode(',', $value);
+            break;
+        case 'verbose':
+            $logLevel = max($logLevel, Log::DEBUG);
+            break;
+        case 'v':
+            $logLevel = max($logLevel, Log::INFO);
+            if(is_array($value)){
+                switch(count($value)){
+                    case 2:
+                        $logLevel = max($logLevel, Log::DEBUG);
+                        break;
+                    case 3:
+                        // This should be implemented as Log::TRACE
+                        // once TRACE is Implemented
+                        $logLevel = max($logLevel, Log::DEBUG);
+                        break;
+                    default:
+                        $logLevel = max($logLevel, Log::INFO);
+                }
+            }
+
+            break;
+        case 'debug':
+            $logLevel = max($logLevel, Log::DEBUG);
+            break;
+        default:
+            fwrite(STDERR, "Unexpected option '$key'\n");
+            exit(1);
+            break;
+    }
+}
+$logger = Log::factory(
+    'xdmod-build-filter-lists',
+    array(
+        'file'            => false,
+        'mail'            => false,
+        'consoleLogLevel' => $logLevel,
+    )
+);
+
+build($realms, $logger);
+
+function build($realms, $logger){
+    try {
+        $filterListBuilder = new FilterListBuilder();
+        $filterListBuilder->setLogger($logger);
+        if (count($realms) > 0){
+            foreach($realms as $realm){
+                if(!empty($realm)){
+                    $filterListBuilder->buildRealmLists($realm);
+                }
+            }
+        }
+        else {
+            $filterListBuilder->buildAllLists();
+        }
+    } catch (Exception $e) {
+        $logger->crit(array(
+            'message'    => 'Filter list building failed: ' . $e->getMessage(),
+            'stacktrace' => $e->getTraceAsString(),
+        ));
+    }
+}
+function displayHelpText()
+{
+    echo <<<'EOF'
+
+Usage: xdmod-build-filter-lists [-v] [-r]
+
+    -h, --help
+        Display this message and exit.
+
+    -v, --verbose
+        Output info level and above log messages.
+
+    -vv, -d, --debug
+        Output debug level and above log messages.
+
+    -q, --quiet
+        Output warning level and above log messages.
+
+    -r[Jobs,SUPREMM]
+        Comma Separated list of realms to build filter list for.  If none specified, auto discover and build all.
+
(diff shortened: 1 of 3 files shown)


bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+    )
+);
+$longOptions = array_map(
+    function ($opt) {
+        return $opt[1];
+    },
+    $opts
+);
+
+$args = getopt($shortOptions, $longOptions);
```
smgallo: Lines 10-37 can be simplified by using an associative array.

```
$opts = arrray(
    'h' => 'help',
    'v' => 'verbose',
    ...
);
$args = getopt(implode('', array_keys($opts)), $opts);
```

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+
+if ($args === false) {
+    fwrite(STDERR, "Failed to parse arguments\n");
+    exit(1);
+}
+if (array_key_exists('h', $args) || array_key_exists('help', $args)){
```
smgallo: Why not just add this as another case below?

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+                        // once TRACE is Implemented
+                        $logLevel = max($logLevel, Log::DEBUG);
+                        break;
+                    default:
+                        $logLevel = max($logLevel, Log::INFO);
+                }
+            }
+
+            break;
+        case 'debug':
```
smgallo: This clause is duplicated.

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+    function ($opt) {
+        return $opt[1];
+    },
+    $opts
+);
+
+$args = getopt($shortOptions, $longOptions);
+
+if ($args === false) {
+    fwrite(STDERR, "Failed to parse arguments\n");
```
smgallo: Call `displayHelpText()` here instead and pass it an optional message.

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+}
+if (array_key_exists('h', $args) || array_key_exists('help', $args)){
+    displayHelpText();
+    exit();
+}
+$realms = array();
+$logLevel = Log::NOTICE;
+foreach ($args as $key => $value) {
+    if (is_array($value) && 'v' !== $key) {
+        fwrite(STDERR, "Multiple values not allowed for '$key'\n");
```
smgallo: Call `displayHelpText()` here instead and pass it an optional message.

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+            $filterListBuilder->buildAllLists();
+        }
+    } catch (Exception $e) {
+        $logger->crit(array(
+            'message'    => 'Filter list building failed: ' . $e->getMessage(),
+            'stacktrace' => $e->getTraceAsString(),
+        ));
+    }
+}
+function displayHelpText()
```
smgallo: This can be a multi-purpose function if you also accept an optional message. You can also output the proper exit code. For example:
```
function usage_and_exit($msg = null)
{
    if ($msg !== null) {
        fwrite(STDERR, "\n$msg\n\n");
    }
```

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+{
+    echo <<<'EOF'
+
+Usage: xdmod-build-filter-lists [-v] [-r]
```
smgallo: I find that `Usage: {$argv[0]}` is better here because then you don't have to change your code if you change the file name.

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+    -v, --verbose
+        Output info level and above log messages.
+
+    -vv, -d, --debug
+        Output debug level and above log messages.
+
+    -q, --quiet
+        Output warning level and above log messages.
+
+    -r[Jobs,SUPREMM]
```
smgallo: Shouldn't this be: `[-r Jobs,SUPREMM]` since it is optional?

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+        else {
+            $filterListBuilder->buildAllLists();
+        }
+    } catch (Exception $e) {
+        $logger->crit(array(
+            'message'    => 'Filter list building failed: ' . $e->getMessage(),
+            'stacktrace' => $e->getTraceAsString(),
+        ));
```
smgallo: Add `exit(1)` here?

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+        case 'r':
+        case 'realms':
+            $realms = explode(',', $value);
+            break;
+        case 'verbose':
+            $logLevel = max($logLevel, Log::DEBUG);
+            break;
+        case 'v':
+            $logLevel = max($logLevel, Log::INFO);
+            if(is_array($value)){
```
smgallo: Normalize `$value` to an array and you can lose line 73 and the `if` statement. `$value = ( is_array($value) ? $value : array($value) );`

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,153 @@
...
+        case 'debug':
+            $logLevel = max($logLevel, Log::DEBUG);
+            break;
+        case 'q':
+        case 'quiet':
+            $logLevel = max($logLevel, Log::WARNING);
+            break;
+        case 'r':
+        case 'realms':
+            $realms = explode(',', $value);
```
smgallo: If we want to maintain consistency, I've been supporting multiple options rather than a single one with a list of comma separated values. With multiple options specified they come in as an array. See https://github.com/ubccr/xdmod/blob/fa80766a18756bcbb5216af99dbfbf482fca35c4/tools/etl/etl_overseer.php#L251-L258

bin/xdmod-build-filter-lists
plessbd: I should have started from a different base :)

bin/xdmod-build-filter-lists
plessbd: I like micro-optimizations...
I actually had it as part of the switch, will move it back.

bin/xdmod-build-filter-lists
plessbd: Yeah since I am already doing that for -v I can do that here as well.  Will switch to multiple for consistency.

bin/xdmod-build-filter-lists
plessbd: agreed, again, should have used a different base.

bin/xdmod-build-filter-lists
smgallo: Optimizations are good for bottlenecks, but here... Meh.

bin/xdmod-build-filter-lists
smgallo: Slacker.

bin/xdmod-build-filter-lists
```diff
@@ -0,0 +1,132 @@
...
+        'mail'            => false,
+        'consoleLogLevel' => $logLevel,
+    )
+);
+build($realms, $logger);
+
+function build($realms, $logger){
+
+    try {
+        throw Exception('AHHH');
```
smgallo: Why not "Arrrggghhh"? https://www.youtube.com/watch?v=LJfowXTXOfU

plessbd: crap dumb debug code...

plessbd: like it never happened....

jtpalmer:
If you're adding a new command to do what is already part of `xdmod-ingestor`, you should remove that functionality from `xdmod-ingestor` and update the documentation.

jtpalmer:
Also, why does this script need to exist?  What's the motivation and context?  When we discussed this you said you would add filter list building to the ETL pipeline.

smgallo:
Some additional context in response to @jtpalmer's question. Since we will now have 2 ETL pipelines in Open XDMoD (and already do for XSEDE) the filters need to be run after both complete as a separate process. Also, since creating the filters can take 15-20 minutes it is useful to only update  filters for specific realms while developing and debugging.

smgallo:
@jtpalmer There is a plan for a separate (multiple) PR to remove the filter list code from the core and XSEDE modules and add it into the appropriate cron jobs

bin/xdmod-ingestor
```diff
@@ -73,7 +73,7 @@ function main()
     }
 
     $help = $ingest = $aggregate = $noAppend = $ingestAll = $ingestShredded
-        = $ingestStaging = $ingestHpcdb = $buildFilterLists = false;
+        = $ingestStaging = $ingestHpcdb = false;
```
smgallo: Also need to remove the command line option `array('', 'build-filter-lists'),`

PR #338, opened 2017-11-29
Updating update_user to handle new acl structure

ryanrath:
## Description
- the structure of acls as they're sent over from the front end has changed from
  ['acl1', 'acl2', 'acl3'] -> {'acl1': ['center1', 'center2'], 'acl2': []}
  and as such the function used to test for the absence of the 'mgr' role needs
  to be changed from `in_array` to `array_key_exists`.
## Motivation and Context
Because you should be able to modify your own user.

## Tests performed
Manual Test:
-  Login to the internal dashboard 
- Select 'User management' tab
- Select the 'Existing Users' tab
- Double click on the account that you are currently logged in as.
- Click the checkbox next to the 'Principal Investigator' acl. ( It doesn't matter if it was checked previously or not )
- Click 'Save Changes' 
- It should save successfully. 
- To double check force refresh the page and navigate to the acct again.
- Ensure that the 'Principal Investigator' acl is the same as it was after the previous update.

## Types of changes

## Checklist:

### DIFF
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -59,7 +59,7 @@ $params = array('uid' => RESTRICTION_UID);
    	if (isset($_POST['acls'])) {
    	
          $role_config = json_decode($_POST['acls'], true);
-         if (!in_array(ROLE_ID_MANAGER, $role_config)) {
+         if (!array_key_exists(ROLE_ID_MANAGER, $role_config)) {
                 $returnData['success'] = false;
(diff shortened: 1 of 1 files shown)


jpwhite4:
It looks like there is at least one other place in the code that must be fixed due to the new acl structure. On line 126 the code references the 'primaryRole' key from the acl structure, which does not exist anymore.

ryanrath:
@jpwhite4 Thank you for this catch it definitely needed to be changed. I would be lying if I said I didn't have a case of tunnel vision sometimes so it's much appreciated. I've added that change in the most recent commit.

PR #342, opened 2017-11-30
No sending tokens during login event tracking

chakrabortyr:
## Description
As a part of the Globus UI changes, we noticed an XDMoD.TrackEvent() method sending the user's REST token over as part of a successful login. This is removed. Additionally, a minor update to unsuccessful login tracking (to denote that it actually was unsuccessful).

## Motivation and Context
We shouldn't be sending user tokens over to Google Analytics.

## Tests performed
Surely.

## Types of changes

## Checklist:


smgallo:
How will this affect tracking of individual user movement within the XDMoD UI via google analytics?

chakrabortyr:
To the above (won't let me reply directly), it doesn't. We're ostensibly using categories and actions to quantify events. The token appears to be sent as part of the event details, which you can't meaningfully use during drilldown.

PR #346, opened 2017-12-04
Update MetricExplorerTest.php

plessbd:
make the expected and actual be in the right order so we get valid error messages
https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertEquals

Actual
```bash
1) IntegrationTests\Controllers\MetricExplorerTest::testGetDimensionFilters
Failed asserting that 200 matches expected 500.
```
expected
```bash
1) IntegrationTests\Controllers\MetricExplorerTest::testGetDimensionFilters
Failed asserting that 500 matches expected 200.
```


plessbd:
Well if you dont like it, tell the phpunit people to change the order of their api....

PR #347, opened 2017-12-05
Fix Component Tests

ryanrath:
## Description
- Fixing the component tests (Acl related) so that they utilize the new
  environment specific location for files that contain the expected results.

## Motivation and Context
Not broken tests are good.

## Tests performed
the tests?

## Types of changes

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
+++ b/open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
@@ -59,15 +59,17 @@ class XDUserTest extends \PHPUnit_Framework_TestCase
 
     public static function setUpBeforeClass()
     {
+        self::setupEnvironment();
+    }
+    private static function setupEnvironment()
+    {
         $testEnvironment = getenv('test_env');
         if ($testEnvironment !== false) {
             self::$ENV = $testEnvironment;
         } else {
             self::$ENV = self::DEFAULT_TEST_ENVIRONMENT;
         }
-
     }
-
 
     /**
      * @dataProvider provideGetUserByUserName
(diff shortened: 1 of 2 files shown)


open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php
```diff
@@ -1297,4 +1355,78 @@ public function provideGetFormalRoleName()
...
+        while (array_key_exists($username, self::$users)) {
+            $suffix = rand(self::MIN_USERS, self::MAX_USERS);
+            $username = "$username$suffix";
+        }
+        return $username;
+    }
+
+    public static function getTestFile($fileName)
+    {
+        if (!isset(self::$ENV)){
```
plessbd: is this actually needed?  for any of this to get called doesnt setUpBeforeClass get called which should initialize this variable?

ryanrath: Yup, if this function is to be called outside of this class then we need to ensure that the environment variable is set before returning a value.

plessbd:
seems like this is failing...
```
There was 1 failure:

1) ComponentTests\XDUserTest::testCreateUserWithNonStandardPrimaryRole
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'mgr'
+'usr'

/root/src/github.com/ubccr/xdmod/open_xdmod/modules/xdmod/component_tests/lib/XDUserTest.php:629

FAILURES!
Tests: 170, Assertions: 289, Failures: 1.
```

ryanrath:
Yup, it was but is passing now.

PR #353, opened 2017-12-13
Add 'Select All' button to filter dialog.

jpwhite4:
- This change adds a select all button to the filter dialog box. Now that
  a large number of filters can potetially added or removed in one go,
  this change also performance optimizations when large numbers of filters
  change.
- Select all button is grayed out if there are more than 150 filters
  since more that this number has a detrimental performance impact on
  several parts of the XDMoD UI (the database seems to cope fine though).


jpwhite4:
Note that the build failure is caused because the main branch is broken at the moment and is not due to these proposed changes.

PR #363, opened 2018-01-12
Dont regress just test!

plessbd:
this should be adding some regression tests...

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -18,6 +18,7 @@ build:
         - ./open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
         - composer install --no-progress
         - cp ~/assets/secrets open_xdmod/modules/xdmod/integration_tests/.secrets.json
+        - ./open_xdmod/modules/xdmod/tests/artifacts/update-artifacts.sh
         - ./open_xdmod/modules/xdmod/integration_tests/runtests.sh --log-junit `pwd`/shippable/testresults/xdmod-integration.xml
         - cp ./configuration/portal_settings.ini ./configuration/portal_settings.ini.old
         - cp -f /etc/xdmod/portal_settings.ini ./configuration/portal_settings.ini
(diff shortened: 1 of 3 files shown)


shippable.yml
```diff
@@ -18,6 +18,12 @@ build:
         - ./open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
         - composer install --no-progress
         - cp ~/assets/secrets open_xdmod/modules/xdmod/integration_tests/.secrets.json
+        - ./open_xdmod/modules/xdmod/tests/artifacts/update-artifacts.sh
+        - REG_TEST_USER_ROLE=user ./open_xdmod/modules/xdmod/regression_tests/runtests.sh --log-junit `pwd`/shippable/testresults/xdmod-regression-user.xml
```
jpwhite4: I think I'd like the runtests.sh script automatically run all of the tests (for all of the users). This would then be consistent with all of the other test types.

If you really want to be able to only run a sub-set of tests then you could make the runtests script accept  commandline args to restrict the tests that are run (but I wouldn't even bother doing that since a develper can just run phpunit manually themselves.

plessbd: makes sense, I was more focused on getting it to run.  once I shippable to cooperate ill change it to this.  Though I am thinking I might just have to disable the chrome yum repo...

plessbd: This has now been "resolved"
Ideally I want to and might resturcture all of `open_xdmod/modules/xdmod/*` to be something more like `open_xdmod/modules/xdmod/tests/{shared,unit,regression,browser,component,integration}` and in there we just have a runtests.sh that will go and hit all of the individual tests so we only have one script to call.

Also, I want the regression tests to be refactored to just use the .secrets.json (which I guess we have two of and need to fix with next docker build) and not have to be specified individually.

But having something that works is better than having no regression tests...

jpwhite4:
The only slight problem with this is that it looks like it now takes 40 minutes to run the tests.

PR #377, opened 2018-01-23
Force artifacts to be relative to test

plessbd:
This is needed for easier support of multi environment / module regression tests.


open_xdmod/modules/xdmod/regression_tests/lib/Controllers/UsageExplorerTest.php
```diff
@@ -117,7 +118,7 @@ public function csvExportProvider()
         self::$baseDir = __DIR__ . '/../../../tests/artifacts/xdmod-test-artifacts/xdmod/regression/current/';
         $envBaseDir = getenv('REG_TEST_BASE');
         if(!empty($envBaseDir)){
-            self::$baseDir = $envBaseDir;
+            self::$baseDir = __DIR__ . $envBaseDir;
```
jpwhite4: realpath()?

PR #378, opened 2018-01-23
Add image tests; only run a random subset of regression tests by defa…

jpwhite4:
…ult.

export REG_TEST_ALL=1 to have all tests run (this should take 15 - 20 minutes or so).

also fix the test script so that test errors cause the script to have a non-zero exit code.

### DIFF
--- a/open_xdmod/modules/xdmod/regression_tests/lib/Controllers/UsageChartsTest.php
+++ b/open_xdmod/modules/xdmod/regression_tests/lib/Controllers/UsageChartsTest.php
@@ -1,0 +1,172 @@
+<?php
+
+namespace RegressionTests\Controllers;
+
+class UsageChartsTest extends \PHPUnit_Framework_TestCase
+{
+    protected static $helper;
+
+    // Used when running in hash-generation mode.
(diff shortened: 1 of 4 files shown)


open_xdmod/modules/xdmod/regression_tests/lib/Controllers/UsageChartsTest.php
```diff
@@ -0,0 +1,172 @@
...
+            $hash = $expectedHashes[$testName];
+        }
+
+        return array($testName, $reference, $hash);
+    }
+
+    public function chartSettingsProvider()
+    {
+        self::$helper = new \TestHarness\XdmodTestHelper();
+        self::$helper->authenticate('cd');
```
plessbd: do we only want to test one user or multiple?

jpwhite4: For the purposes of these tests one user is sufficient since I'm testing the different combinations of chart settings (the code paths of which are independent of the user account). The SUPReMM equivalent of these test will need to check different user accounts though.

PR #381, opened 2018-01-24
Add test for SQL parser

smgallo:
## Description

Add a test for the SQL parser. Also
- Allow Loggable::setLogger() to accept a null argument and use the empty logger
- Improve handling of SQL reserved words such as DISTINCT

## Motivation and Context

Test upgraded SQL parser

## Tests performed

Ran existing test suite

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/aRdbmsDestinationAction.php
+++ b/classes/ETL/aRdbmsDestinationAction.php
@@ -767,7 +767,7 @@ abstract class aRdbmsDestinationAction extends aAction
                  && array_key_exists('name', $item['alias'])
             ) {
                 $columnNames[] = $item['alias']['name'];
-            } else {
+            } elseif ( 'reserved' != $item['expr_type'] ) {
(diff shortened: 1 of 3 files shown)


classes/ETL/aRdbmsDestinationAction.php
```diff
@@ -767,7 +767,7 @@ public function getSqlColumnNames($sql)
                  && array_key_exists('name', $item['alias'])
             ) {
                 $columnNames[] = $item['alias']['name'];
-            } else {
+            } elseif ( 'reserved' != $item['expr_type'] ) {
```
jpwhite4: Is this here to enable testing or is it an bug fix seen when testing?

smgallo: That was a bug found during testing. Otherwise it treats "DISTINCT organization" as 2 individual column names instead of just "organization".

PR #383, opened 2018-01-24
User's acls were being cleared on save

ryanrath:
## Description
After updating a users acls, the acl display would be cleared, thus no longer showing the selected users acls. 

## Motivation and Context
A users acls should be displayed after changes are saved. 

## Tests performed
Manual Tests: 
**Test 1:**
- Login to the internal dashboard
- Select 'User Management' -> 'Existing Users'
- Select a user that has more than one acl
- Remove an acl from that user
- Save changes
- Expected: 
  - The current acls for the user are displayed. 

## Types of changes

## Checklist:

### DIFF
--- a/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
+++ b/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
@@ -687,9 +687,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
         };
 
         self.resetDirtyState = function () {
-            settingsAreDirty = false;
-            roleGrid.updateDirtyState();
-            roleGrid.reset();
+            roleGrid.setDirtyState(false);
         };
 
         // ------------------------------------------


html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -687,9 +687,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
         };
 
         self.resetDirtyState = function () {
-            settingsAreDirty = false;
-            roleGrid.updateDirtyState();
```
smgallo: This function is still used in the `onMouseDown` event in the AclGrid, does it work in that context?

ryanrath: That it does, in the onMouseDown handler for AclGrid ( and previously in RoleGrid ) we found the following: 
```js
isDirty = true;
```
Which meant that whenever the 'Include' column was clicked the whole component was considered dirty. This code failed to take into account the starting state of the components ( i.e. which acls were pre-selected when a user is viewed ). The current code / function basically redefines being `dirty` as when any acl record's 'include' property is different then it started with.

PR #384, opened 2018-01-25
Adding Object.values polyfill

ryanrath:
## Description

Internet Explorer ( non-Edge ) does not support `Object.values` so this
polyfill provides that functionality if it does not already exist.

## Motivation and Context
It's nice to have all javascript methods used in an app to work in all supported browsers.

## Tests performed
Manual Test in xdmod docker image
**Test 1:**
- Login to internal dashboard
- Navigate to 'User Management' -> 'Existing Users'
- Click 'Create & Manage Users'
- Select the 'New User' Tab
- Fill out the required information
- Click the 'Save' button
- The user should be successfully created
  - *NOTE: on a docker instance an error will be returned as `sendmail` is not setup. But if you go to / refresh the 'Existing Users' tab then the newly created user will be present*
## Types of changes

## Checklist:

### DIFF
--- a/html/gui/lib/ie-object-values-polyfill.js
+++ b/html/gui/lib/ie-object-values-polyfill.js
@@ -1,9 +1,9 @@
 if (!Object.values) {
-    Object.values = function(obj) {
+    Object.values = function (obj) {
         return Object.keys(obj).map(
-          function(key) {
+          function (key) {
               return obj[key];
           }
         );
-    }
+    };
 }
--- a/html/gui/lib/ie-object-values-polyfill.js
+++ b/html/gui/lib/ie-object-values-polyfill.js
@@ -1,0 +1,9 @@
+if (!Object.values) {
+    Object.values = function(obj) {
+        return Object.keys(obj).map(
+          function(key) {
+              return obj[key];
+          }
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


html/gui/lib/ie-object-values-polyfill.js
```diff
@@ -0,0 +1,9 @@
+if (!Object.values) {
+    Object.values = function(obj) {
+        return Object.keys(obj).map(
+          function(key) {
+              return obj[key];
+          }
+        );
+    }
+}
```
smgallo: Need a newline here

html/index.php
```diff
@@ -112,6 +112,7 @@ function isReferrer($referrer)
       <link rel="shortcut icon" href="gui/icons/favicon_static.ico" />
       <script type="text/javascript" src="gui/lib/oldie-console-patch.js"></script>
       <script type="text/javascript" src="gui/lib/oldie-array-includes-polyfill.js"></script>
+      <script type="text/javascript" src="gui/lib/ie-object-values-polyfill.js"></script>
```
smgallo: We may want to consider bundling all of these small patches together for performance reasons.

ryanrath: That's certainly a possibility, but I'm not sure we gain a lot by bundling just the polyfills / patches. I'm certainly open to seeing some benchmarks though. I would hazard a guess that concatenating / minifying / zipping *all* of our js might lead to some performance gains but then again I'm not sure.

ryanrath:
re: style checks:
I have setup the xdmod-qa repo + lint-diff and have a method for running this pre-commit now.

PR #387, opened 2018-01-26
Fix incorrect data filtering

jpwhite4:
Fixes issue with data filtering when an unpriv user tries to view
timeseries data for a realm / stat that is protected (i.e. has
additional role-specific filters applied).

Added integration test that confirms the correct behaviour.

### DIFF
--- a/classes/DataWarehouse/Query/Query.php
+++ b/classes/DataWarehouse/Query/Query.php
@@ -761,7 +761,7 @@ class Query
      * The where conditions and role parameters from the other class will
      * overwrite any existing settings in this class.
      */
-    public function cloneParameters($other)
+    public function cloneParameters(Query $other)
     {
         $this->_where_conditions = $other->_where_conditions;
         $this->parameters = $other->parameters;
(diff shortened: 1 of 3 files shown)


classes/DataWarehouse/Query/Query.php
```diff
@@ -756,6 +756,21 @@ public function setParameters(array $parameters = array())
         }
     }
 
+    /**
+     * Copy the parameters and role parameters from another query class
+     * The where conditions and role parameters from the other class will
+     * overwrite any existing settings in this class.
+     */
+    public function cloneParameters($other)
```
smgallo: We should use `(Query $other)` in the definition if we are always expecting an object of that class.

smgallo:
Other than the comment above, LGTM.

jpwhite4:
I note that the shippable build failed because the expected results were changed as a consequence of a different pull request that has not yet been approved. We will have to come up with a better way of dealing with external test artefact dependencies.

ryanrath:
That's my bad, I've reverted the test artifact so that your test should pass now and created a new artifact with the new expected test results for my branch ( and the test that uses it ).

PR #388, opened 2018-01-26
XSEDE_USER_TYPE to FEDERATED_USER_TYPE

ryanrath:
## Types of changes

## Checklist:


html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -687,6 +676,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
         };
 
         self.resetDirtyState = function () {
+            settingsAreDirty = false;
```
jpwhite4: Should this change be in this pull request?

html/controllers/user_admin/update_user.php
```diff
@@ -186,8 +186,8 @@ function ($carry, Acl $item) {
 
    $returnData['success'] = true;
 
-   $statusPrefix = $user_to_update->isXSEDEUser() ? 'XSEDE ' : '';
-   $displayUsername = $user_to_update->isXSEDEUser() ? $user_to_update->getXSEDEUsername() : $user_to_update->getUsername();
+   $statusPrefix = $user_to_update->isFederatedUser() ? 'Federated ' : '';
```
jpwhite4: Will the displayUsername be correct now? you are changing the logic..

classes/NewRest/Controllers/UserControllerProvider.php
```diff
@@ -131,7 +131,7 @@ private function extractUserData(XDUser $user)
             'first_name' => $user->getFirstName(),
             'last_name' => $user->getLastName(),
             'email_address' => $emailAddress,
-            'is_xsede_user' => $user->isXSEDEUser(),
+            'is_federated_user' => $user->isFederatedUser(),
```
smgallo: `is_federated_user` is already set on the following line. Why can't we just use that?

ryanrath: yeah, we can. I was in replace mode at that point. Removal incoming

html/controllers/user_admin/update_user.php
ryanrath: per our conversation: yes, the displayUsername will contain the correct value for Federated Users ( although the value will be different than before. This is expected. )

classes/XDUser.php
```diff
@@ -2845,7 +2844,7 @@ public static function initializeXSEDEUser($username)
             $last_name
         );
 
-        $user->setUserType(XSEDE_USER_TYPE);                   // XSEDE User
+        $user->setUserType(FEDERATED_USER_TYPE);                   // XSEDE User
```
jpwhite4: Suggest remove the (now) inaccurate comment // XSEDE User

classes/XDUser.php
```diff
@@ -2906,7 +2905,7 @@ public static function deriveUserFromXSEDEUser($username)
             "SELECT id FROM Users WHERE username=:username AND user_type=:user_type",
             array(
                 'username' => $username,
-                'user_type' => XSEDE_USER_TYPE
+                'user_type' => FEDERATED_USER_TYPE
```
jpwhite4: Rather than changing this variable I would just remove this (unused) function altogether (unless it is being used somewhere).

classes/XDUser.php
```diff
@@ -2940,7 +2939,7 @@ public static function XSEDEUserExists($username)
             'SELECT id FROM moddb.Users WHERE username = :username AND user_type=:user_type',
             array(
                 'username' => $username,
-                'user_type' => XSEDE_USER_TYPE
+                'user_type' => FEDERATED_USER_TYPE
```
jpwhite4: This function needs to be gone too.

html/controllers/user_admin/delete_user.php
```diff
@@ -24,8 +24,8 @@
        
       $username = $user_to_remove->getUsername();
    
-      $statusPrefix = $user_to_remove->isXSEDEUser() ? 'XSEDE ' : '';
-      $displayUsername = $user_to_remove->isXSEDEUser() ? $user_to_remove->getXSEDEUsername() : $user_to_remove->getUsername();
+      $statusPrefix = $user_to_remove->isFederatedUser() ? 'Federated ' : '';
```
jpwhite4: This line can go completely.

db/migrations/7.1.0-7.5.0/moddb.sql
```diff
@@ -0,0 +1,2 @@
+-- Update the 'Federated' User Type color
+UPDATE UserTypes SET color = '#b914f6' WHERE type = 'Federated';
```
smgallo: Need a newline here.

html/controllers/user_admin/update_user.php
smgallo: Prior to implementing Globus auth, the XSEDE oAuth service returned the username as "<username>;<display>" (e.g., "mchupa;Michael Chupa") so there was extra code needed to decode it for viewing. This is no longer necessary.

html/internal_dashboard/controllers/controller.php
```diff
@@ -125,12 +125,10 @@
 
     case 'enum_user_types_and_roles':
 
-        $query = "SELECT id, type, color FROM moddb.UserTypes";
+        $query = "SELECT id, type, color FROM moddb";
```
smgallo: Why did you remove the table from this select statement?

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -210,7 +210,17 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
             root: 'user_types',
             autoLoad: true,
             baseParams: { operation: 'enum_user_types' },
-            fields: ['id', 'type']
+            fields: ['id', 'type'],
+            listeners: {
+                load: function (store, records) {
+                    for (var i = 0; i < records.length; i++) {
+                        var record = records[i];
+                        if (record.data.type === 'Federated') {
```
smgallo: Since we are already using `CCR.xdmod.FEDERATED_USER_TYPE` elsewhere why not use it here as well and check the id instead of the type string? If we need to make a change it's one less to make.

html/internal_dashboard/controllers/controller.php
ryanrath: Because I'm losing my mind

smgallo: `s/losing/lost/` ? :)

ryanrath: yup, pretty much

PR #396, opened 2018-01-30
switching to captcha v2

plessbd:
## Description
update to use new recaptcha
## Motivation and Context
Google is decommissioning the v1 api on marxh 31st, 2018 https://developers.google.com/recaptcha/docs/faq#what-happens-to-recaptcha-v1

## Tests performed
This is a captcha specifically meant to not allow automated testing.
I went though all the contact forms in my docker instance to verify that I was got valid responses.


html/index.php
```diff
@@ -291,10 +291,10 @@ function isReferrer($referrer)
             print "CCR.xdmod.logged_in = !CCR.xdmod.publicUser;\n";
             $useCaptcha = 'false';
             try {
-              $useCaptcha = xd_utilities\getConfiguration('mailer', 'captcha_private_key') !== '' ? 'true' : 'false';
+                $useCaptcha = xd_utilities\getConfiguration('mailer', 'captcha_public_key') !== '' ? 'true' : 'false';
```
smgallo: Here we are checking `captcha_public_key` but in the actual forms themselves we are using `if ($captcha_private_key !== '')`. Why the difference?

PR #397, opened 2018-01-30
update testPortalSettingsTemplate

plessbd:
Update testPortalSettingsTemplate to allow for additional comments without the test failing.


jtpalmer:
Why would you want comments in one file and not the other?  The purpose of the test is to make sure that the files match (including comments).

plessbd:
After discussion with jeff, the template is what should be getting updated.
We currently have a circular reference in that the build process requires configuration/portal_settings.ini to exist before it can build things like portal_settings.ini

PR #400, opened 2018-02-01
Update Centers::getCenters sql

ryanrath:
## Description
- A center is an organization that has a corresponding entry in resourcefact

## Motivation and Context
Not sure why this didn't make it in one of the earlier PR, but it's still needed

## Types of changes

## Checklist:

### DIFF
--- a/classes/Models/Services/Centers.php
+++ b/classes/Models/Services/Centers.php
@@ -27,8 +27,12 @@ class Centers
     public static function getCenters()
     {
         $db = DB::factory('database');
-
-        return $db->query("SELECT o.* FROM modw.organization o");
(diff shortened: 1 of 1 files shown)


classes/Models/Services/Centers.php
```diff
@@ -27,8 +27,12 @@ class Centers
     public static function getCenters()
     {
         $db = DB::factory('database');
-
-        return $db->query("SELECT o.* FROM modw.organization o");
+        return $db->query("
+    SELECT o.*
```
smgallo: This should be `SELECT DISTINCT o.*` or you'll get multiple rows due to multiple resources under the same SP.

PR #404, opened 2018-02-02
Update Filters in Metric Explorer in a group instead of when each individual row is selected

eiffel777:
## Description
Filters in the Metric Explorer are applied to a chart or data series to restrict the information that is shown. Filters are added via the “Add Filter” menu option and can be viewed via the “Filters” menu option or by selecting the Edit Dataset option in the Data Series menu. 

Filters are added and removed one at a time as you select or deselect the checkbox for each filter. This change will group changes to the filters into one request instead of a request each time a filter is selected or deselected by adding an Apply button to the panel that show the filters that already exist. 

The Apply, Check All and Uncheck All buttons will also be moved to the bottom bar of the dialog to create consistency with existing user interface elements. Screenshots of the new UI are below.

screenshot from 2018-02-02 14-58-10

screenshot from 2018-02-02 14-58-45

## Motivation and Context
When an individual filter is enabled or disabled, the chart is immediately updated to reflect the change. When enabling or disabling several filters in a short period of time, this behavior causes a request to be sent to the back-end for each filter that was modified and may cause undesirable behavior such as heavy database load.

## Tests performed
Have performed manually tests of new functionality on https://gmdean.ccr.xdmod.org and am writing appropriate automated tests

## Types of changes

## Checklist:


html/gui/js/AddDataPanel.js
```diff
@@ -136,7 +136,7 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
                     scope: this,
                     handler: function (b, e) {
                         XDMoD.TrackEvent('Metric Explorer', 'Data Series Definition -> Selected filter from menu', b.text);
-                        filterButtonHandler.call(b.scope, filtersGridPanel.toolbars[0].el, b.dimension, b.text, b.realms);
+                        filterButtonHandler.call(b.scope, this.items.items[0].items.items[11].items.items[0].el, b.dimension, b.text, b.realms);
```
smgallo: There are six `.items` here, which make it incredibly difficult to understand. Also, what does the index 11 reference? What happens if this array is modified and 11 is no longer the item you expect?

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -4362,11 +4367,14 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
                 }
             ],
             tbar: [
-                checkAllButton,
+                removeFilterItem
+            ],
+            bbar: [
+                applyFilterSelection,
```
smgallo: The Add Filter dialog has actual buttons across the bottom with Clear on the left and OK/Apply and Cancel on the right. Can we do the same thing here to keep the look and feel consistent? It looks like the Add Filter dialog is an Ext.Panel and adds the buttons in the `fbar` (see [FilterDimensionPanel.js](https://github.com/ubccr/xdmod/blob/xdmod7.1/html/gui/js/FilterDimensionPanel.js#L300)). Can we do the same thing here by placing the buttons in the chartFiltersMenu instead? We might need to wrap the filters grid in an Ext.Panel but we'd have to experiment to see if that works.

screenshot-2018-2-3 xdmod

smgallo: Check with @jpwhite4 so we keep the location of the check/uncheck all button consistent between this and the work he is doing on the Add Filter dialog.

jpwhite4:
diff --git a/html/gui/js/AddDataPanel.js b/html/gui/js/AddDataPanel.js
index dd5a68d..4803eb7 100644
--- a/html/gui/js/AddDataPanel.js
+++ b/html/gui/js/AddDataPanel.js
@@ -111,6 +111,8 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
         };
     },
     initComponent: function () {
+        var filterButtonHandler;
+
         if (!this.record && this.store) {
             this.record = CCR.xdmod.ui.AddDataPanel.initRecord(this.store, this.config, this.getSelectedFilters(), this.timeseries);
         }
@@ -157,7 +159,7 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
             }
         );
         filtersMenu.addItem(filterItems);
-        var filterButtonHandler = function (el, dim_id, dim_label, realms) {
+        filterButtonHandler = function (el, dim_id, dim_label, realms) {
             if (!dim_id || !dim_label) return;
             var filterDimensionPanel = new CCR.xdmod.ui.FilterDimensionPanel({
                 origin_module: 'Metric Explorer',

eiffel777:
Here are some screenshots of updated filter dialog box with new button layout. After talking with @smgallo I am going to add a cancel button real quick on Tuesday morning.

@jpwhite4 @smgallo @plessbd This is on https://gmdean.ccr.xdmod.org. If you guys have a few moments sometime can you take a look and let me know if I've missed anything.

**Global Filters**
screenshot from 2018-02-05 15-50-50

**Local Filters**
screenshot from 2018-02-05 15-51-22

smgallo:
I see what you mean about the window width but I think it will be OK. Can you shrink the width of the checkbox column similar to what is was in the original? Also it looks like we're missing the Applicable Categories column.

ryanrath:
Since having the checkbox checked doesn't necessarily mean that a filter is being applied to the current chart anymore. Do people think that it might be worthwhile to call out which filters are currently applied ( like changing the color / having a separate column to track which ones are currently applied / not )? Or maybe that there are filter changes that haven't been applied?

smgallo:
@ryanrath When a user first brings up the dialog the checked filters will be applied. If they check/uncheck them and want to get back to the original state they can click Cancel.

ryanrath:
\*nod\* sounds good.

smgallo:
The version currently on https://gmdean.ccr.xdmod.org looks good with the smaller checkbox column and cancel button. I did notice that if I open an existing filter, uncheck one and apply it, then re-open it and enable the previously un-checked filter then Cancel the one that was previously unchecked stays checked.

html/gui/js/AddDataPanel.js
```diff
@@ -291,16 +318,42 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
...
+            scope: this,
+            handler: function () {
+                this.filtersStore.each(function (record) {
+                    if (record.modified !== null) {
+                        if (record.get('checked') !== record.modified.checked) {
+                            var checked = !record.get('checked') || false;
+                            record.set('checked', checked);
+                            // Deletes the dirty and modified property from the record
+                            // to signify the record is now unmodified
+                            delete this.dirty;
```
jpwhite4: What is `this` in this context? The store or the record or the button?

plessbd:
please add  open_xdmod/modules/xdmod/integration_tests/secrets.json_TEMPLATE back in, we dont want to remove the template

eiffel777:
@plessbd Whoops, thanks for the catch. I added it back

plessbd:
now it doesnt have a new line (might never have had one) add it as well please :)

smgallo:
@eiffel777 Looks like there is a conflict that needs to be resolved now.

smgallo:
@jpwhite4, @plessbd Please review this one final time.

html/gui/js/AddDataPanel.js
```diff
@@ -253,8 +274,8 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
             var currentFilters = jQuery.extend({}, this.record.data.filters);
             this.filtersStore.loadData(currentFilters, false);
         }
-        var checkAllButton = new Ext.Button({
-            text: 'Check All',
+        var selectAllButton = new Ext.Button({
+            text: 'Select All',
             scope: this,
             handler: function (b, e) {
                 XDMoD.TrackEvent('Metric Explorer', 'Data Series Definition -> Clicked on Check All in Local Filters pane');
```
plessbd: We should update this tracking event to match the button text, not that we use the tracking all that much but just to keep things consistent

html/gui/js/AddDataPanel.js
```diff
@@ -263,8 +284,8 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
                 });
             }
         });
-        var uncheckAllButton = new Ext.Button({
-            text: 'Uncheck All',
+        var clearAllButton = new Ext.Button({
+            text: 'Clear All',
             scope: this,
             handler: function (b, e) {
                 XDMoD.TrackEvent('Metric Explorer', 'Data Series Definition -> Clicked on Uncheck All in Local Filters pane');
```
plessbd: We should update this tracking event to match the button text, not that we use the tracking all that much but just to keep things consistent

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -4198,50 +4197,44 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
...
         // ---------------------------------------------------------
 
-        var checkAllButton = new Ext.Button({
-            text: 'Check All',
+        var selectAllButton = new Ext.Button({
+            text: 'Select All',
             scope: this,
             handler: function( /*b, e*/ ) {
-                    XDMoD.TrackEvent('Metric Explorer', 'Clicked on Check All in Chart Filters pane');
+                XDMoD.TrackEvent('Metric Explorer', 'Clicked on Check All in Chart Filters pane');
```
plessbd: We should update this tracking event to match the button text, not that we use the tracking all that much but just to keep things consistent

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -4198,50 +4197,44 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
...
         // ---------------------------------------------------------
 
-        var uncheckAllButton = new Ext.Button({
-            text: 'Uncheck All',
+        var clearAllButton = new Ext.Button({
+            text: 'Clear All',
             scope: this,
             handler: function( /*b, e*/ ) {
-                    XDMoD.TrackEvent('Metric Explorer', 'Clicked on Uncheck All in Chart Filters pane');
+                XDMoD.TrackEvent('Metric Explorer', 'Clicked on Uncheck All in Chart Filters pane');
```
plessbd: We should update this tracking event to match the button text, not that we use the tracking all that much but just to keep things consistent

PR #412, opened 2018-02-08
set make timestamps look the same

plessbd:
i have no comments, I am to sad.[Your Code is bad](http://s2.quickmeme.com/img/29/292febc5ac05e551a651eacfa9ecf9eeeb1081e088ed0c3c1f7575e1eed90aaa.jpg)

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -1039,7 +1039,7 @@ SQL;
 
         $time_frags = explode('.', $init_time);
 
-        return date('m/d/Y, g:i:s A', $time_frags[0]);
+        return date('Y-m-d, H:i:s', $time_frags[0]);
 
     }//getLastLoginTimestamp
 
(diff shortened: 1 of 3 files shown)


classes/XDUser.php
```diff
@@ -1039,7 +1039,7 @@ public function getLastLoginTimestamp()
 
         $time_frags = explode('.', $init_time);
 
-        return date('m/d/Y, g:i:s A', $time_frags[0]);
+        return date('Y-m-d, H:i:s', $time_frags[0]);
```
jpwhite4: Given that we are changing the date format, why don't we use some sort of standard format such as an ISO 8601 compatible format rather than creating out own one?

plessbd: this is used to match what other things are already using

smgallo: Agree that ISO8601 is good, but we should do that in a separate PR that examines XDMoD as a whole.

PR #417, opened 2018-02-09
Acl config sync

ryanrath:
## Description
Ensure that `acl-config` adds support for both the removal and modification of configuration files that populate the tables that it manages as well as addition.  

## Motivation and Context
So that we can truly manage the state of the acl tables w/ config file modifications.

## Tests performed
Manual Tests: 

**Test 1:**
- Start an Open XDMoD docker container and connect
- Start the services provided by the docker container ( `~/bin/services start` )
- Add a new acl to the 'normaluser' user
  - Could be via directly adding a record in `moddb.user_acls` or via the internal dashboard
- Perform an source upgrade
- Verify that the newly added acl relation still exists
  - By either directly via the db or via the internal dashboard

**Test 2:**
- Start an Open XDMoD docker container and connect
- Start the container services ( `~/bin/services start` )
- Add a new acl to the system
  - Either modify roles.json and run `acl-config` or manually add a record to `moddb.acls`
- Perform a source upgrade -or- if roles.json was modified in the previous step then revert the file and run `acl-config` again 
- Verify that the newly added acl is not present
  - either via db query or internal dashboard

**NOTE: I also ran all of the unit, integration and component tests for each Test ( which passed )**
## Types of changes

## Checklist:


bin/acl-config
```diff
@@ -215,6 +216,72 @@ function main()
 
     $db = DB::factory('database');
 
+    // Ensure that we're starting from scratch while backing up the tables that
+    // we do not populate ( user_acls, user_acl_group_by_parameters ) which will
+    // be re-populated after we are done w/ the population of the tables
+    $tables = array(
```
smgallo: The array keys easily get lost in the SQL. Maybe a more readable way to do this is to set a few variables and then assign them into the array. For example:
```
$userAclCreateSql = <<< SQL
blah blah blah
SQL
$userAclPopulateSql = <<< SQL
blah blah blah
SQL

$tables = array(
    'user_acls' => array(
        'create' => $userAclCreateSql,
        'populate' => $userAclPopulateSql
    ),
    ...
);
```

bin/acl-config
```diff
@@ -253,8 +320,148 @@ function main()
...
+
+    foreach($tables as $table => $queries) {
+        $creationQuery = $queries['create'];
+        $query = <<<SQL
+CREATE TEMPORARY TABLE ${table}_bkup AS $creationQuery;
+SQL;
+        $log->debug("Backup query [$table]: $query");
+
+        if ($dryRun) {
+            $log->notice("Table[$table] would have been created:\n$query");
```
smgallo: You should display the query using `debug()` prior to execution and then if not in dryrun mode you can say "Table created".

bin/acl-config
```diff
@@ -253,8 +320,148 @@ function main()
...
+            continue;
+        }
+        // throws exception or returns row count
+        $rows = $db->execute($query);
+        $message = $rows > 0
+            ? "Table [$table] => records[$rows] populated"
```
smgallo: Why not just keep it simple and say `sprintf("Inserted %d rows into %s", $rows, $table)`"?

bin/acl-config
```diff
@@ -253,8 +320,148 @@ function main()
...
+    foreach($tables as $table) {
+        $query = <<<SQL
+DROP TABLE $table;
+SQL;
+        if ($dryRun) {
+            $log->info("Would have dropped: $table");
+            continue;
+        }
+
+        $log->debug("Dropping Table[$table]:\n\t$query");
```
smgallo: Similarly here, you can simply display the sql (it's self explanatory) above and then say "Table created" after executing the statement.

bin/acl-config
```diff
@@ -215,6 +216,72 @@ function main()
...
+        gb.group_by_id IS NOT NULL;
+SQL
+
+        )
+    );
+
+    // populated via this script.
+    backupTables($db, $tables);
+
+    wipeManagedTables($db);
```
smgallo: Why are you dropping the tables in `wipeManagedTables()` and then running the ETL to re-create them? And why not call this `dropManagedTables()` as it is more descriptive? You can just truncate them in the ETL by adding the `-o "truncate_destination_table=true"` flag on the command line which will apply that option to each action.

bin/acl-config
```diff
@@ -215,6 +216,72 @@ function main()
 
     $db = DB::factory('database');
 
+    // Ensure that we're starting from scratch while backing up the tables that
+    // we do not populate ( user_acls, user_acl_group_by_parameters ) which will
+    // be re-populated after we are done w/ the population of the tables
+    $tables = array(
+        'user_acls' => array(
+            'create' => 'SELECT ua.user_id, a.name as acl_name FROM user_acls ua JOIN acls a ON a.acl_id = ua.acl_id',
```
smgallo: This is not table creation but rather table backup? Why not include the entire backup SQL here so that your reference below to `user_acls_bkup` has better context?

ryanrath:
@smgallo This process doesn't / wasn't meant to handle the possible removal of an acl from a user as we don't specify which users are assigned which acls in the configuration files. 

re: 4 - Because I didn't think it would be a good thing to wipe which users have which acls for an entire installation? 

I can definitely make the clarification changes you suggested in your other comments.

smgallo:
@ryanrath Understood as per our conversation as long as ACLs that may have been removed are handled.

bin/acl-config
```diff
@@ -217,70 +217,32 @@ function main()
     $db = DB::factory('database');
 
     // Ensure that we're starting from scratch while backing up the tables that
-    // we do not populate ( user_acls, user_acl_group_by_parameters ) which will
-    // be re-populated after we are done w/ the population of the tables
-    $tables = array(
+    // we do not manage ( user_acls, user_acl_group_by_parameters ). These
+    // tables will be re-populated after we are done creating / re-populating
+    // the managed tables.
+    $userAclCreateBkup = retrieveETLSQLFile('create_user_acls_bkup');
```
smgallo: Are these 4 queries used elsewhere? If not, why not just leave the SQL definitions here in the code?

configuration/etl/etl_sql.d/acls/xdmod/create_uagbp_bkup.sql
```diff
@@ -0,0 +1,18 @@
+-- Name: create_uagbp_bkup.sql
```
smgallo: If these queries are not part of the ETL process they really shouldn't be in the ETL directory...

bin/acl-config
ryanrath: They are not. I can certainly do that, I was just taking `The array keys easily get lost in the SQL` one step further by removing the sql from the code entirely and putting them in sql files. If you would rather the sql be in the code then I can change it.

PR #424, opened 2018-02-12
Handle single-point timeseries plotting bug in our highcharts wrappers.

jsperhac:
## Description

*This pr is intended to address plotting of single point datasets, regardless of selected time aggregation. (The second commit does the same for non-timeseries plots so they display properly in e.g. Summary tab.)* 

This change fixes some faulty logic that determines whether plot point markers will be displayed in line and area charts. Previously, a blank plot resulted from plotting a dataset with a single timeseries point and a long Duration with daily aggregation. 

Example: Usage tab plot with single y data point and more than 30 time points appears to be empty (note that not all x-axis time points are labeled here):
single-point-before

Whereas: Fix to line plot with single y data point and more than 30 time points correctly displays the point:
single-point-after

The change counts up *y* points (plottable points) in timeseries datasets. If a single y point is present, it is plotted with a point marker. Otherwise, the previous logic is used. Namely, XDMoD displays point markers only if fewer than n x-axis points are present (NOTE: n=30 in Usage, n=21 in Metric Explorer. The n value is not changed in this fix.). 

Both the Usage adapter and the HighchartsTimeseries class were changed in this fix.

### Update: a second commit

I have added a second commit which also makes this fix in both timeseries and non-timeseries charts displayed in the Summary tab. 

Here's a Summary tab aggregate chart, prior to the fix. It should contain a single point:
uncorrected-single-agg-summary

With the second commit, the point is visible:
corrected-single-agg-summary

Here's another example from the Summary tab, showing timeseries charts.

Before fix, Summary tab chart does not display single point data series:
uncorrected-single-point-summary

After the fix, single point data series are displayed:
corrected-single-point-summary

## Motivation and Context

This fix addresses the case in which a single data point is being plotted in a timeseries chart for which there are more than 30 x-axis timepoints. For area or line plots, that single point was not displayed until mouseover of the chart area. 

This issue was found during tests for 7.1. See remarks here:
https://app.asana.com/0/14787510600562/550096118806901

### Second commit

The second commit I added also applies this logic to charts displayed in the Summary tab. Note that aggregate (non-timeseries) charts must also be considered!

## Tests performed

Change was tested by hand atop 7.1 code base. Note that the parameter whose calculation is changed in this fix is not used to do anything but turn plot point markers on and off.

Checked carefully: Sparse plots, such as Jobs summary plots with User quick filter enabled; toggling between timeseries and non-ts plots; single plot under multiple Durations. 

Comments are welcome as to whether existing tests need to be adapted for this!

## Types of changes

Note that this may cause a change to tests that involved plots that incorrectly showed no data...

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -696,6 +696,7 @@ class Usage extends Common
 
                 // For each data series...
                 $primaryDataSeriesRank = $usageOffset;
+
                 array_walk($meChart['series'], function (
                     &$meDataSeries,
                     $meDataSeriesIndex
@@ -743,9 +744,22 @@ class Usage extends Common
 
                     // If this is the primary data series and the chart is not a
                     // thumbnail, use line markers if and only if the number of
-                    // points is less than or equal to 30.
+                    // data points is less than or equal to 30,
+                    // or if there's a single y series data point.
                     if ($isPrimaryDataSeries && !$thumbnailRequested) {
-                        $meDataSeries['marker']['enabled'] = count($meDataSeries['data']) <= 30;
+                        // is there a single y data point?
+                        $y_values_count = 0;
+                        foreach ($meDataSeries['data'] as $value) {
+                            if ($value['y'] !== null ) {
+                                ++$y_values_count;
+                            }
+                            // we are only interested in the == 1 case
+                            if ($y_values_count > 1) {
+                                break;
+                            }
+                        }
+                        $meDataSeries['marker']['enabled'] = $y_values_count == 1 ||
+                                            count($meDataSeries['data']) <= 30;
                     }
 
                     // If this is a trend line data series...
--- a/classes/DataWarehouse/Visualization/HighChartTimeseries2.php
+++ b/classes/DataWarehouse/Visualization/HighChartTimeseries2.php
@@ -586,11 +586,11 @@ class HighChartTimeseries2 extends HighChart2
                                 break;
                             }
                         }
-                        // Display markers for scatter plots, or for non-thumbnail plots
-                        // with fewer than 21 points or with a single y value.
+                        // Display markers for scatter plots, non-thumbnail plots of data series
+                        // with fewer than 21 points, or for any data series with a single y value.
                         $showMarker = $data_description->display_type == 'scatter' ||
-                            ( ($y_values_count == 1 || $values_count < 21) &&
-                            $this->_width > \DataWarehouse\Visualization::$thumbnail_width);
+                            ($values_count < 21 && $this->_width > \DataWarehouse\Visualization::$thumbnail_width) ||
+                            $y_values_count == 1;
 
                         $isRemainder = $dataTruncated && ($yIndex === $numYAxisDataObjects - 1);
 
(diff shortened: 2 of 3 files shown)


classes/DataWarehouse/Access/Usage.php
```diff
@@ -740,9 +741,17 @@ public function getCharts(XDUser $user, $chartsKey = 'data') {
 
                     // If this is the primary data series and the chart is not a
                     // thumbnail, use line markers if and only if the number of
-                    // points is less than or equal to 30.
+                    // data points is less than or equal to 30,
+                    // or if there's a single y series data point.
                     if ($isPrimaryDataSeries && !$thumbnailRequested) {
-                        $meDataSeries['marker']['enabled'] = count($meDataSeries['data']) <= 30;
+                        $yValues = array();
```
jpwhite4: Seems unnecessary to build an array in order to count the number of entries in it. You could just increment a counter.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -740,9 +741,17 @@ public function getCharts(XDUser $user, $chartsKey = 'data') {
...
+                        foreach ($meDataSeries['data'] as $value) {
+                            if (is_array($value) && $value['y'] !== null ) {
+                                $yValues[] = $value['y'];
+                            }
+                        }
+                        $meDataSeries['marker']['enabled'] = (count($yValues) == 1 ||
```
jpwhite4: Why not just show the marker if yValuesCount < 30?

classes/DataWarehouse/Visualization/HighChartTimeseries2.php
```diff
@@ -571,7 +572,18 @@ public function configure(
                         }
 
                         $values = $yAxisDataObject->getValues();
+
+                        // Decide whether to show data point markers:
                         $values_count = count($values);
+                        // Count datapoints having actual, non-null y values:
+                        $y_values_count = count(array_filter($values, function ($value) {
```
jpwhite4: How come you use an array_filter here but an explicit loop in the change for Usage.php?

classes/DataWarehouse/Access/Usage.php
jsperhac: Your suggestion was my initial approach to this bug--but it changes the appearance of these sparse plots. I feel the change is for the better, but still, it's a change. Examples:

Using yValuesCount < 30 :
ten-points-after

Initial appearance:
ten-points-before

classes/DataWarehouse/Access/Usage.php
jsperhac: Good point, I'm not doing anything else with the array.

classes/DataWarehouse/Visualization/HighChartTimeseries2.php
jsperhac: In Usage.php the y values are nested more deeply and it was simpler to take apart as an explicit loop. Granted, it would be tidier to be consistent.

classes/DataWarehouse/Access/Usage.php
jpwhite4: Sound good.

PR #428, opened 2018-02-14
Updating styles to be more consistent

plessbd:
Make spacing more consistent
view branch name for further explanation...

Before

screen shot 2018-02-14 at 11 20 24 am

After
screen shot 2018-02-14 at 11 19 31 am


plessbd:
I am glad you got that I purposefully set the spacing of the title :)

PR #429, opened 2018-02-14
Attempt to provide drilldown filtering for group bys

ryanrath:
## Description
This is a rough draft of providing the ability to filter out specific group bys ( via configuration file ) in OpenXDMoD while keeping the behavior the same in XSEDE. 

## Motivation and Context
We provide the ability to filter group_bys included as query descripters from the Usage tab / Metric Explorer Tab via roles.json->roles->role->query_descripters->query_descripter->show/hide. Where as the ability to filter a group_by from drop down lists is only accessible from within the GroupBy*.php code `[get|set]AvailableOnDrilldown`. 

These changes allow us to provide a 'visible' property for group_by records in datawarehouse.json that, in OpenXDMoD, when false will keep the group_by from appearing in the drill down list. 

## Tests performed
Manual Tests: 

**Test 1:**
- Login to OpenXDMOD
- Click the 'Usage' Tab
- Ensure that 'Jobs by Service Provider' is not visible 
- Plot some data from Jobs
- Click a point
- Ensure that 'Service Provider' is not included in the 'Drilldown to:' menu. 

**Test 2:**
- Login to XSEDE XDMoD
- Click the 'Usage' Tab
- Ensure that 'Jobs by Service Provider' is visible 
- Plot some data from Jobs
- Click a point
- Ensure that 'Service Provider' is included in the 'Drilldown to:' menu. 

## Types of changes

## Checklist:


jpwhite4:
Please can you provide a bit more information about how this bug arose, since it was not present in the 7.0 release. The reason I ask is that this fix appears to create a new code path for filters which was not present before (even though it worked as expected before).

classes/DataWarehouse/Query/Jobs/GroupBy.php
```diff
@@ -1,47 +1,65 @@
...
+    }
 
-?>
\ No newline at end of file
+    public function getDrillTargets($statistic_name, $query_classname)
+    {
+        $registerd_group_bys = Aggregate::getRegisteredGroupBys();
+        $drill_target_group_bys = array();
+        $config = Config::factory();
+        $multipleServiceProviders = \xd_utilities\getConfiguration('features', 'multiple_service_providers') === 'on';
```
jpwhite4: Why does this code care whether or not there are multiple service providers?

ryanrath:
Going to close this one for now, the simpler solution is to just remove the 'provider' group by from datawarehouse.json. After running some tests I've confirmed that this doesn't affect filtering in OpenXDMoD w/ SUPREMM installed. If, in the future we need to revisit the ability to filter group_bys from drilldowns then we'll have this as a possible place to tlook.

ryanrath:
@jpwhite4 Yup, just tested the modifications we spoke about yesterday in docker w/ OpenXDMOD + SUPREMM and it looks like having the modification in `Query::getGroupBy` + the appropriate properties in [datawarehouse|roles].json ( or module equivalent ) yields filtering in both the population of the Usage tree and the population drilldowns ( and the Metric Explorer Group Bys ). I'll clean up my dev code and get it all committed.

classes/DataWarehouse/Query/Query.php
```diff
@@ -1503,7 +1503,23 @@ public static function getGroupBy($group_by_name)
...
+                    function ($value) use ($group_by_name) {
+                        return $value['name']=== $group_by_name;
+                    }
+                )
+            );
+            $visible = true;
+            if (array_key_exists('visible', $found)) {
+                $visible = $found['visible'];
+            }
+            $group_by_name_to_instance[$group_by_name]->setAvailableOnDrilldown($visible);
```
jpwhite4: I would only call the setAvailableOnDrillDown if the visible property exists. The reason for this is so that the behaviour can be described as: if the visible property exists then its value will override the default settings for the group by other wise the default settings will be used.

classes/DataWarehouse/Query/Jobs/GroupBy.php
```diff
@@ -1,47 +1,47 @@
...
-/* 
+/*
 * @author Amin Ghadersohi
 * @date 2011-Jan-07
 *
 * class for adding group bys to a jobs query
-* 
+*
 */
+use Xdmod\Config;
```
jpwhite4: Since this file no longer needs to have any functional code changes in it I recommend reverting it to exactly as it was before.

classes/DataWarehouse/Query/Jobs/GroupBy.php
```diff
@@ -1,47 +1,47 @@
...
-				$drill_target_group_bys[] = $group_by_name.'-'.$group_by_instance->getLabel();
-			}			
-		}
-		
-		sort($drill_target_group_bys);
-		return $drill_target_group_bys;
-	}
-
-	
+    public function __construct($name, array $additional_permitted_parameters = array(), $possible_values_query = NULL)
```
plessbd: unless I am missing something, this file only have tabs converted to spaces, not sure we need to do this.

classes/DataWarehouse/Query/Query.php
```diff
@@ -1503,7 +1503,22 @@ public static function getGroupBy($group_by_name)
 			static::registerGroupBys();
 			$classname = static::getGroupByClassname($group_by_name);
 			$group_by_name_to_instance[$group_by_name] = new $classname;
-		}
+
+			$realm = static::getRealm();
```
plessbd: this line should not have tabs, it should be spaces.

classes/DataWarehouse/Query/Jobs/GroupBy.php
ryanrath: We always need to do this

plessbd: if we were making functional changes to the file I would agree with the cleanup.  But we are not making any other changes to this file that is why I was saying I am not sure we need to do it (in this case)

PR #435, opened 2018-02-16
update php version

plessbd:
update composer to force version 5.4.16 instead of 5.3.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -401,7 +401,7 @@
     },
     "config": {
         "platform": {
-            "php": "5.3.3"
+            "php": "5.4.16"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -401,7 +401,7 @@
     },
     "config": {
         "platform": {
-            "php": "5.3.3"
+            "php": "5.4.16"
```
jtpalmer: Is there something in 5.4.16 that we need that isn't in earlier versions of 5.4?  Should the documentation be updated to specify 5.4.16 and not 5.4?

PR #446, opened 2018-03-01
Updated enum_user_types_and_roles to utilize acls

ryanrath:
## Description
Update Internal Dashboard -> Existing Users -> Role Filter ( having a role of <role> ) to utilize Acls. This also entailed updating the backend for the Internal Dashboard -> User Management -> Existing Users -> Prepare E-Mail button. Specifically, the Role Filter used to determine the list of users to send the email to. Also cleaned up this function `deriveUserEnumerationQuery` to remove unused parameters ( `context_filter` ) and change it from Dynamic SQL to a parameterized query. 

## Motivation and Context
The Role Filtering was not operating as intended due to roles having difference id's than acls ( perhaps obviously )

## Tests performed
**Manually Tested:**
- Log in to the Internal Dashboard
- Click the User Management -> Existing Users tab
- Select the 'having a role of: <Role>' drop down
- Select a number of different roles
- Ensure that the users displayed do indeed have the selected role
- Click the 'Prepare E-Mail' toolbar button
- Select the 'having a role of: <Role>' drop down
- Select a role
- Click the 'Send' button
- Ensure that the users identified by the email addresses displayed have the role previously selected.

**Integration Tests:**
`ControllerTest::testEnumUserTypesAndRoles`

## Types of changes

## Checklist:


html/internal_dashboard/controllers/controller.php
```diff
@@ -118,7 +118,7 @@
 
         $response['user_types'] = $results;
 
-        $query = "SELECT description, role_id FROM moddb.Roles WHERE abbrev != 'pub' ORDER BY description";
+        $query = "SELECT display description, acl_id role_id FROM moddb.acls WHERE name != 'pub' ORDER BY description";
```
jpwhite4: Please can use include the AS keyword so that it is clear that the columns are being aliased.

ryanrath: Not a problem, change incoming.

ryanrath:
Note: The shippable tests won't be able to pass until: https://github.com/ubccr/xdmod/pull/417 is merged as the id property for acls is different depending on whether or not the system is fresh installed or upgraded ( atm ). After 417 is merged it will be the same ( verified on OpenXDMoD docker checking install vs. upgrade ).

html/internal_dashboard/controllers/mailer.php
```diff
@@ -63,9 +63,9 @@
         $group_filter = \xd_security\assertParameterSet('group_filter');
         $role_filter = \xd_security\assertParameterSet('role_filter');
 
-        $query = \xd_dashboard\deriveUserEnumerationQuery($group_filter, $role_filter, '', true);
+        list($query, $params) = \xd_dashboard\deriveUserEnumerationQuery($group_filter, $role_filter, true);
```
jpwhite4: Since this function implementation is almost entirely different from the old one it does not make sense to keep the name the same. This function no longer derives the user enumeration instead it gets a list of unique email addresses. I recommend changing the function name to something more meaningful.

html/internal_dashboard/controllers/mailer.php
```diff
@@ -63,9 +63,9 @@
...
 
-        $results = $pdo->query($query);
+        $results = $pdo->query($query, $params);
```
jpwhite4: The PHP code below runs array_unique() and sort() but the SQL query uses DISTINCT and an ORDER BY.

The ORDER BY in the SQL will be overridden by the sort().

ryanrath:
@jpwhite4 I've updated the code per your review comments. If you could take another look when you have a sec that would rock.

open_xdmod/modules/xdmod/integration_tests/lib/Controllers/ControllerTest.php
```diff
@@ -275,7 +275,7 @@ public function listUsersGroupProvider()
     public function testEnumUserTypesAndRoles()
     {
         $expected = JSON::loadFile(
-            $this->getTestFiles()->getFile('controllers', 'enum_user_types_and_roles-xsede_to_federated')
+            $this->getTestFiles()->getFile('controllers', 'enum_user_types_and_roles-update_enum_user_types_and_roles')
```
plessbd: I think we should just name the files based on the version so that multiple updates to the same version only need to change the expected input / output of the file instead of the tests as well.  At least that is what Steve and I were thinking at one point.  Might want to have a discussion to discuss pros and cons.

`enum_user_types_and_roles-{major.min.patch}`
ie:
`enum_user_types_and_roles-8.0.0`

ryanrath: Sounds good, my only concern would be that if you have multiple branches adding / modifying tests for the same app version you might end up changing the expected output for other branches.

smgallo: Our plan is to move the artifacts into their respective repos in the near future.

PR #449, opened 2018-03-01
Use stronger password hashing algorithm

jpwhite4:
See asana issue https://app.asana.com/0/14740318781081/578673499549601

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -696,21 +696,34 @@ SQL;
 
         $pdo = DB::factory('database');
 
-        $userCheck = $pdo->query("SELECT id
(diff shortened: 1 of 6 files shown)


jpwhite4:
Note that the "upgrade" part of the CI tests are expected to fail until such a time as the docker image is updated to 7.5.

jpwhite4:
Note that the upgrade tests are now failing because the sql migration script is not being run.

plessbd:
migrations should be fixed with #453

plessbd:
see #454

plessbd:
Composer will need to be updated now that #420 has been merged

PR #450, opened 2018-03-01
Remove mandatory debug from ACL scripts and change -g option to -d for consistency

smgallo:
## Description

Existing acl scripts `acl-import` and `acl-xdmod-management` had debug hard-coded. This has been removed and parameters are now passed through to the underlying `acl-etl` script so debug or verbose mode can be turned on by the user.

Also change the debug parameter from `-g` to `-d` for consistency with other scripts.

## Motivation and Context

Consistency.

## Tests performed

Tested manually in docker.

## Types of changes

## Checklist:


bin/acl-import
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-acl-etl -pacls-import -g
+acl-etl -p acls-import "$@"
```
jpwhite4: It is very tempting to get rid of both of these one-line shell scripts altogether since they only call one command. (But not nin this pull request).

smgallo: Agreed on both counts.

PR #453, opened 2018-03-05
Make Upgrades work again

plessbd:
I broke all the things...
my code is bad and I feel a tiny little bit maybe kinda sorta bad

### DIFF
--- a/classes/OpenXdmod/Migration/Version750to800/DatabasesMigration.php
+++ b/classes/OpenXdmod/Migration/Version750to800/DatabasesMigration.php
@@ -1,9 +1,24 @@
 <?php
-namespace OpenXdmod\Migration\Version700To701;
+namespace OpenXdmod\Migration\Version750To800;
 
 /**
  * Migrate databases from version 7.5.0 to 8.0.0.
  */
-class DatabasesMigration extends \OpenXdmod\Migration\Version660To700\DatabasesMigration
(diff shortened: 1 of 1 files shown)


classes/OpenXdmod/Migration/Version750to800/DatabasesMigration.php
```diff
@@ -1,9 +1,24 @@
...
 /**
  * Migrate databases from version 7.5.0 to 8.0.0.
  */
-class DatabasesMigration extends \OpenXdmod\Migration\Version660To700\DatabasesMigration
+class DatabasesMigration extends \OpenXdmod\Migration\DatabasesMigration
 {
+    /**
+     * @see \OpenXdmod\Migration\Migration::__construct
+     **/
+    public function __construct($currentVersion, $newVersion)
```
smgallo: Why is this necessary? If no constructor is defined the next constructor up the chain should be called. Same thing for `execute()`. For example,
```php
class Migrate
{
    public function __construct($currentVersion, $newVersion)
    {
        echo sprintf('%s::%s(%s, %s)', __CLASS__, __FUNCTION__, $currentVersion, $newVersion) . PHP_EOL;
    }
}

class M2 extends Migrate
{
}

$m = new Migrate(1, 2);
$m2 = new M2(2, 3);
```

```
smgallo@dave:~/tmp$ php test_parent_constructor.php 
Migrate::__construct(1, 2)
Migrate::__construct(2, 3)
```

PR #464, opened 2018-03-13
Fix xdmod-ingest typo in 8.0 hierarchy.md

simonflood:
Fixed typo - Ingestor command is xdmod-ingestor not xdmod-ingest


simonflood:
Replaced by later request.

PR #477, opened 2018-03-15
Improve XSEDE ACL tests

smgallo:
## Description

Bring the ACL and other tests up to date so they work with both the Open XDMoD and XSEDE docker images. To do this the following steps were taken:
- Two different methods were being used to retrieve files from two different locations from [xdmod-test-artifacts](https://github.com/ubccr/xdmod-test-artifacts). Switch instances of `BaseTest::getFile()` to use `TestHarness\TestFiles` (from the integration tests) instead.
- When comparing user objects the entire JSON export of the object was compared, including database keys. Since keys may change while the values remain the same (e.g., if values were added in a different order or if additional values were added) we don't want to compare the keys. Filter only the fields that matter.
- Removed hardcoded Open XDMoD test values and used data from [xdmod-test-artifacts](https://github.com/ubccr/xdmod-test-artifacts) instead, allowing both Open XDMoD and XSEDE values to be included.
- Clean up unused code.
- Improve documentation on how to run tests and which versions to test.
- Improve `runtests.sh` to support running tests appropriate for the version we are testing. For example, do not test the Slurm shredder for XSEDE.

Several changes were made in ubccr/xdmod-test-artifacts#4 to go along with this PR including:
- Moving XSEDE input/output files to their proper location
- Moving hardcoded values from the test into data files
- Update expected output to match values from the XSEDE docker
- Removing files that are no longer needed. These files were never used in testing and existed to support the XSEDE docker

**_Note that the XSEDE 7.5 docker image requires modification for the tests to work properly. The required changes have been made to the create_xdmod_user.php script used to build the image and the forthcoming 8.0 docker image will include these changes._**

To run the tests against the xsede7.5 docker run the following commands:
```
docker run -h database --rm -it -p 3306:3306 tas-tools-int-01.ccr.xdmod.org/xdmod-centos7:xsede7.5 /bin/bash
```

Copy the updated create user script from the private repo and secrets file into the container:

```
$ docker cp ~/src/ccr-private-xdmod/docker/xdmod-xsede/bin/create_xdmod_users.php 019b1f25a9aa:/root/bin
$ docker cp ~/src/ccr-private-xdmod/docker/xdmod-xsede/assets/secrets.json  019b1f25a9aa:/root/
```

Delete then re-create the center director and center staff users:

```
# cd /root
# mysql -B -N -e "select id from moddb.Users where username in ('centerdirector', 'centerstaff');" | xargs -I\} bin/create_xdmod_users.php -d \}
Deleting user: centerdirector (7539)
Deleting user: centerstaff (7540)

# bin/create_xdmod_users.php -s secrets.json 
Creating user: centerdirector
Creating user: centerstaff
Creating user: principal
Error creating user 'principal': User principal already exists
Creating user: normaluser
Error creating user 'normaluser': User normaluser already exists
Creating user: admin
Error creating user 'admin': User admin already exists
```

## Motivation and Context

Bring ACL tests in line with XSEDE docker image.

## Tests performed

Tested with the open7.5-v4 and xsede7.5 docker images.

## Types of changes

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/component_tests/README.md
+++ b/open_xdmod/modules/xdmod/component_tests/README.md
@@ -4,7 +4,7 @@
 
 1. Set `TEST_ENV` based on the tests that you would like to perform.  The default is to run the Open
 XDMoD tests (e.g., `TEST_ENV=xdmod`).
-2. Execute the tests: `./runtests`
+2. Execute the tests: `./runtests.sh`
 
 ## Environment Variables
 
--- a/open_xdmod/modules/xdmod/component_tests/README.md
+++ b/open_xdmod/modules/xdmod/component_tests/README.md
@@ -1,0 +1,57 @@
+# XDMoD Component Tests
+
+## Running Tests
+
+1. Set `TEST_ENV` based on the tests that you would like to perform.  The default is to run the Open
+XDMoD tests (e.g., `TEST_ENV=xdmod`).
+2. Execute the tests: `./runtests`
+
+## Environment Variables
+
+The test environment is selected using the `TEST_ENV` environment variable. Supported values are
+- `xdmod` (this is the default if not set)
+- `xdmod-xsede`
+- `xdmod-supremm`
+
+The value of this variable determines the directory in the `xdmod-test-artifacts` repo that is used
+for reading test files and may also affect which tests are run. For example, XSEDE tests do not need
+to be run when testing Open XDMoD. This allows the same tests to be executed using different inputs
+and different expected outputs. The `xdmod-test-artifacts` repo must contain a directory matching
+the value of `TEST_ENV`.
+
+## Test Groups
+
+Depending on the XDMoD version being tested, not all tests are necessary and some tests may not run.
+For example, the Open XDMoD shredder tests will not work when run against the XSEDE docker image
+because the XSEDE version does not include the necessary `mod_shredder` database (XSEDE does not use
+this infrastructure but pulls from the XDCDB). We use PHPUnit [test
+groups](https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.group)
+to organize tests into logical groups that can be selected based on the version being tested. Groups
+may be added to a test class or individual test methods. Multiple groups may be added to the same
+target.
+
+Supported groups are:
+- `@group XDMoD-common` General Open XDMoD tests relevant to all modules. Note that the input and expected
+   output files may be controlled using the `TEST_ENV` environment variable.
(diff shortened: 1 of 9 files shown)


smgallo:
Note that the shippable tests will fail until ubccr/xdmod-test-artifacts#4 is merged.

open_xdmod/modules/xdmod/component_tests/README.md
```diff
@@ -0,0 +1,57 @@
+# XDMoD Component Tests
+
+## Running Tests
+
+1. Set `TEST_ENV` based on the tests that you would like to perform.  The default is to run the Open
+XDMoD tests (e.g., `TEST_ENV=xdmod`).
+2. Execute the tests: `./runtests`
```
plessbd: I believe you have to use `./runtests.sh`

Though I am fine if we get rid of the .sh portion on all of these and just make them executable.

open_xdmod/modules/xdmod/component_tests/lib/QueryTest.php
```diff
@@ -27,7 +27,7 @@ public function testGetDurrationResult($period, $start, $end, $groupby, $expecte
     }
 
     public function queryDataProvider(){
-        $expectedFileName = $this->getTestFile('aggregate_durations.json');
+        $expectedFileName = $this->getTestFiles()->getFile('acls', 'aggregate_durations');
```
plessbd: We could add a [realpath](http://php.net/manual/en/function.realpath.php) to this
to make it a little easier to read when there are errors
```
PHP Warning:  file_get_contents(/root/src/github.com/ubccr/xdmod/open_xdmod/modules/xdmod/component_tests/lib/..//./artifacts/xdmod-test-artifacts/xdmod/roles/output/get_identifer_absolute.json): failed to open stream: No such file or directory in /root/src/github.com/ubccr/xdmod/classes/CCR/Json.php on line 26
PHP Warning:  file_get_contents(/root/src/github.com/ubccr/xdmod/open_xdmod/modules/xdmod/component_tests/lib/..//./artifacts/xdmod-test-artifacts/xdmod/roles/input/get_active_center.json): failed to open stream: No such file or directory in /root/src/github.com/ubccr/xdmod/classes/CCR/Json.php on line 26
```

But it might be better to update https://github.com/ubccr/xdmod/blob/xdmod8.0/classes/CCR/Json.php#L26

smgallo: `realpath` returns `FALSE` if the file does not exist so that won't work in this case. However, I did write `\xd_utilities\resolve_path()` to address this.

PR #481, opened 2018-03-19
Cleanup some ETL json files

plessbd:
This is just white space cleanup to some of etl json so that going forward I can do easier compares of json files.


configuration/etl/etl_tables.d/xdmod/hosts.json
```diff
@@ -2,12 +2,13 @@
...
         "engine": "MyISAM",
-        "columns": [{
+        "columns": [
+            {
                 "name": "id",
                 "type": "int(11)",
                 "nullable": false,
                 "extra": "auto_increment"
-            },
+            }
```
jtpalmer: Missing comma here.

PR #485, opened 2018-03-19
Supress and fix wanrings

plessbd:
hosts and job_record contain duplicate entries and are expected to.  This supresses those warnings.
This also fixes other tables having warnings that were appearing about fields not having defaults.  Instead of allowing them on the table level, this just sets them on the ingest query.

### DIFF
--- a/configuration/etl/etl_tables.d/hpcdb-modw/request.json
+++ b/configuration/etl/etl_tables.d/hpcdb-modw/request.json
@@ -6,7 +6,11 @@
         "records": {
             "id": "rf.request_id",
             "primary_fos_id": "rf.primary_fos_id",
-            "account_id": "rf.account_id"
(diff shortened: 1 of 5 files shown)


configuration/etl/etl_tables.d/hpcdb-modw/request.json
```diff
@@ -6,7 +6,11 @@
         "records": {
             "id": "rf.request_id",
             "primary_fos_id": "rf.primary_fos_id",
-            "account_id": "rf.account_id"
+            "account_id": "rf.account_id",
+            "request_type_id": 0,
+            "grant_number": "''",
+            "start_date": "0000-00-00",
```
smgallo: These values need to be `'0000-00-00'` in order for MySQL to handle them properly.

PR #490, opened 2018-03-23
default etl config file for ease of use

plessbd:
sane defaults make life easier


classes/OpenXdmod/Setup/DatabaseSetup.php
```diff
@@ -134,7 +134,7 @@ public function handle()
          */
 
         $command = 'php ' . DATA_DIR . '/tools/etl/etl_overseer.php '
```
smgallo: Might as well just merge all of this into one string.

jtpalmer:
If you change all these files should you change `classes/OpenXdmod/Setup/AclEtl.php` and `classes/OpenXdmod/Setup/AclImport.php` as well?

PR #496, opened 2018-03-28
move things around to make life 'easier'

plessbd:
Update paths in etl to make life 'easier'

`etl_tables.d/{realm}/{target}/{table-definition.json}`
`etl_actions.d/{realm}/{target}/{table-definition.json}`

currently etl_actions.d is `etl_tables.d/actions/` until the etl code is updated with a new action path.


plessbd:
currently shippable seems to be failing on a timeout so here it is run in docker
```
Configuration:	/root/build/xdmod-supremm/tests/integration_tests/phpunit.xml.dist

........................I..

Time: 7.19 seconds, Memory: 5.25Mb

OK, but incomplete, skipped, or risky tests!


Configuration:	/root/build/xdmod/open_xdmod/modules/xdmod/component_tests/phpunit.xml.dist

Time: 240 ms, Memory: 7.75Mb

OK (17 tests, 22 assertions)


Configuration:	/root/build/xdmod/open_xdmod/modules/xdmod/component_tests/phpunit.xml.dist

Time: 7.36 seconds, Memory: 16.25Mb

OK (232 tests, 367 assertions)


Configuration:	/root/build/xdmod/open_xdmod/modules/xdmod/regression_tests/phpunit.xml.dist
Time: 1.87 minutes, Memory: 13.50Mb

OK (1144 tests, 3848 assertions)

**Repeated 4 time (for each user type)**

Configuration:	/root/build/xdmod/open_xdmod/modules/xdmod/regression_tests/phpunit.xml.dist

Time: 14.13 minutes, Memory: 25.75Mb

OK (3032 tests, 5720 assertions)

```

chakrabortyr:
There are places where you've missed renaming modw -> xdw. I think this change should be consistently applied if it's going to be made.

plessbd:
my git history said those never happened :)

plessbd:
After discussing with @smgallo removing the {source} from path, the use case of the federation will have the designation of `/blade` to show that it is coming from a different source in that specific use case as it generally wont happen.

PR #497, opened 2018-03-29
Add option to show columns that differ when comparing tables

smgallo:
## Description

When verifying ETL ingest or simply comparing the data in 2 tables, we often see cases where one or two columns differ in some rows of the source table as compared to the destination table. Previously, we were able to identify these missing rows but it was a manual process to determine whether or not the entire row was missing in the destination, differed in some significant way, or simply had a different modification time due to different ingest times.

We can now specify `--show-row-differences` and, based on the table primary key, determine if a row with the same key exists in the destination and if so, what columns are different. Excluded columns are not used for the comparison. For example:
```
php verify_table_data.php -v info -s modw -d modw_aws -t person -x order_id -x ts -x ts_ts -x status --show-row-differences
2018-03-29 09:58:48 [notice] Compare tables src=modw.person, dest=modw_aws.person
2018-03-29 09:58:48 [info] Exclude columns: order_id, ts, ts_ts, status
2018-03-29 09:58:48 [info] 17 columns
2018-03-29 09:58:48 [info] Row counts: modw.person = 57,223; modw_aws.person = 54,591
2018-03-29 09:58:49 [warning] Primary key does not exist in destination: (id = 59207)
2018-03-29 09:58:49 [warning] Rows differ for primary key: (id = 45461)
nsfstatuscode_id: '9' != '12'
      department: '' != 'Genome Center'
   email_address: 'email1@alum.utoronto.ca' != 'email2@ucdavis.edu'
```

## Motivation and Context

Got tired of doing this manually.

## Tests performed

Used to compare ingest from the sample AWS instance of the XDCDB to ingest from our local mirror.

## Types of changes

## Checklist:

### DIFF
--- a/tools/dev/verify_table_data.php
+++ b/tools/dev/verify_table_data.php
@@ -76,7 +76,7 @@ $options = array(
     'm:'  => 'map-column:',
     'n:'  => 'num-missing-rows:',
     'p:'  => 'pct-error-column:',
-    '4'   => 'show-row-differences',
+    '5'   => 'show-row-differences',
     's:'  => 'source-schema:',
     't:'  => 'table:',
     '4:'  => 'truncate-column:',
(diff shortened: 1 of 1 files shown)


tools/dev/verify_table_data.php
```diff
@@ -74,6 +76,7 @@
     'm:'  => 'map-column:',
     'n:'  => 'num-missing-rows:',
     'p:'  => 'pct-error-column:',
+    '4'   => 'show-row-differences',
```
plessbd: index 4 is used twice...

PR #499, opened 2018-04-05
Separate table def files from action def files

smgallo:
**Note: Another PR will follow next week to update ubccr/xdmod-xsede**

## Description

This PR separates ETLv2 table definitions from action definition files and places them into separate locations.
- etl_tables.d is the location for reusable table definitions
- etl_action_defs.d is the location for action definitions (this are the files referenced in the action configurations in etl.d)

The following changes were made:
- The code was changed to look for the action definition files using the new path (7633c67)
- Relative references to the `etl_tables.d` directory in definition files have been replaced with the `${table_definition_dir}` variable (99bde76, 65d043e)
- Any action definitions that defined a table within the file have now had the table definition moved to its own file and included using `$ref`.
- Action definitions have now been moved into their own directory (91753f3)
- Pipeline definition files (e.g., those defined in `etl.d`) have been updated to reflect the new paths as needed (56f07e2)
- Actions that ingest static JSON data from a file previously used only the table definition. These actions now needed an action definition so those files were created and simply include the table definition using `$ref` (3faadbe)
- `aRdbmsDestinationAction` no longer requires that a definition file is specified in the action options. Since the `TableManager` action only requires table definitions we cannot have this requirement at this level in the class hierarchy so the it has been pushed into the `aIngestor`, `aAggregator`, and `UpdateIngestor` classes. (00059cb)
- The Configuration transformer infrastructure has been updated so extensible configuration options can be passed to the transformer, allowing arbitrary variables to be passed to the transformer that handles JSON references. This deprecates the need for `JsonReferenceWithMacroTransformer`. (0a4311c, 1416946)
- Minor logging improvements (29dfacc)
- Updated tests for new file structure (d549e5c)

## Motivation and Context

With the advent of RFC-6901 JSON references, we can now define a table once and re-use the definition in multiple instances rather than defining the same table in multiple action definition files. All table definitions are now stored in one directory with action definition files stored in another directory.

## Tests performed

Ran the following in dry-run mode to ensure that all configuration files are properly processed.  Also ran the following without dryrun for those pipelines that the 8.0 docker supports.
```
php etl_overseer.php -t -v debug -p acls-import| fgrep '[error]'
php etl_overseer.php -t -v debug -p acls-xdmod-management | fgrep '[error]'
php etl_overseer.php -t -v debug -p hpcdb-xdw.ingest | fgrep '[error]'
php etl_overseer.php -t -v debug -p jobs-cloud-common | fgrep '[error]'
php etl_overseer.php -t -v debug -p jobs-cloud-eucalyptus -r IU-TACC-JETSTREAM | fgrep '[error]'
php etl_overseer.php -t -v debug -p jobs-common | fgrep '[error]'
php etl_overseer.php -t -v debug -p jobs-xdw.aggregate | fgrep '[error]'
php etl_overseer.php -t -v debug -p jobs-xdw.bootstrap | fgrep '[error]'
```

Also ran component and unit tests.

## Types of changes

## Checklist:


smgallo:
I've tagged @plessbd @jtpalmer @jpwhite4 @eiffel777 @chakrabortyr @ryanrath here for review as these changes could affect their work.

smgallo:
@ryanrath The UpdateIngestor doesn't support altering of tables so it doesn't extend aAction. Although, it is old and only used in 1 use case (resource allocations historical data) at the moment. Now that we use ON DUPLICATE KEY UPDATE rather than REPLACE  I wonder if it is needed at all any more.

PR #503, opened 2018-04-11
Updating filter column

ryanrath:
## Description
The `organization_id` column has been renamed to `resource_organization_id`.
This was causing a problem when logging into the site as a center director ( sql
syntax error as the column was not found ).

## Motivation and Context
Users should be able to login regardless of their Acl assignments

## Tests performed
Manual Test:
- Login as a user who's highest privileged acl is center director to an unpatched XDMoD 8.0 site. 
- Notice that an exception is presented and the site is unusable. 

Apply this PR

- Login as a user who's highest privileged acl is center director
- Notice that an exception is **not** thrown and the site is usable again

**NOTE**: I didn't add any new tests as just running the automated tests on an XSEDE instance would find this.

## Types of changes

## Checklist:


jpwhite4:
This needs to be changed in the SUPReMM realm too

PR #505, opened 2018-04-11
Add framework files for aggregation in ETLv2.

jpwhite4:
## Description
Update the etl framework to generate ETLv2 configuration files. 
Add an ETLv2 Ingestor class that can `explode()` column values to multiple rows.

## Motivation and Context
ETLv2 aggregation code is significantly faster to run that the old aggregation code. This change allows the various realms that use the javascript-base etl framework to use this optimized code. 

## Tests performed
Tests performed in https://github.com/ubccr/xdmod-supremm/pull/98


classes/ETL/Ingestor/ExplodeTransformIngestor.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * @see ETL\Ingestor\pdoIngestor::__construct()
+     */
+    public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log $logger = null)
+    {
+        parent::__construct($options, $etlConfig, $logger);
+
+        $this->verifyRequiredConfigKeys(array('explode_column'), $options);
+
+        foreach($options->explode_column as $key => $value) {
```
smgallo: Would this be better expressed as ?
```
$this->srcKey = key($options->explode_column);
$this->destKey = current($options->explode_column);
```

etl/js/lib/etlv2.js
```diff
@@ -0,0 +1,328 @@
...
+    for (m in map) {
+        if (map.hasOwnProperty(m)) {
+            output = output.replace(new RegExp(m, 'g'), map[m]);
+        }
+    }
+
+    return output;
+};
+
+var mkdirandwrite = function (dirname, filename, data) {
```
smgallo: Should be camelCase

classes/ETL/Ingestor/ExplodeTransformIngestor.php
jpwhite4: Can' use the `key()` and `current()` directly on `$options->explode_column` as it is a stdClass object rather than an array.

PR #509, opened 2018-04-13
Update Center Promotion

ryanrath:
## Description
The Center Promotion functionality provided by XDMoD has been not working as intended since it's inception. Originally it was intended that this functionality allow Center Directors to promote users associated with their center to have more access to that Centers data. Thus alleviating the need to contact admins to have this access granted. The way in which this was implemented allowed a user with the Center Director Role to promote users who had been assigned the Center Staff Role for their Center to Center Director. They were also able to demote users who they had previously promoted to Center Director back down to Center Staff. This requires that users be setup w/ the Center Staff Role before they are eligible for promotion by a Center Director which is only able to be done by Admins. Which defeats the intended purpose of saving both Center Directors and Admins time. 

The implementation presented in this Pull Request works as follows. Center Directors are still able to promote / demote users but they instead will be promoting them to and demoting them via the addition / removal of the Center Staff Acl. This necessitated the redefining of which users were eligible for promotion from those who were assigned the Center Staff Acl for the Directors center. To those Users who are considered "associated" with the Director's center.
A user is considered "Associated" if one of the following is true:
- The `organization_id` recorded in their User account is the same as the Director
- The Person assigned to their User account is associated with the same Center as the Director
- The user has a Center Staff acl for the same Center as the Director
  - this was previously the only way in which a user could be considered "associated"

## Motivation and Context
The feature was not working as intended.

## Tests performed
Manual tests were performed exercising the Promotion UI
Integration Tests were also added to exercise the endpoints 

## Types of changes

## Checklist:


classes/Models/Services/Users.php
```diff
@@ -128,4 +132,245 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+      o.id organization_id
+    FROM moddb.Users u
+      JOIN modw.person p ON p.id = u.person_id
+      JOIN modw.organization o ON o.id = p.organization_id
+      JOIN modw.resourcefact rf ON o.id = rf.organization_id
+  ) uo
+    ON uo.user_id = u.id
+  -- This left join retrieves the correct center value for the specified :user_id
+  -- i.e. the user running this query.
+  LEFT JOIN
```
smgallo: Don't need a `LEFT` join here.

classes/Models/Services/Users.php
```diff
@@ -128,4 +132,245 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+    public static function getUsersForPromotion($userId)
+    {
+        $db = DB::factory(self::DB_SECTION_NAME);
+        $query = <<<SQL
+SELECT DISTINCT
+  u.id,
+  CONCAT(u.last_name, ', ', u.first_name, ' [', o.abbrev, ']') AS name
+FROM Users u
+  -- This left join retrieves the correct center associated with this user.
+  LEFT JOIN
```
smgallo: Don't need a `LEFT` join here.

classes/Models/Services/Users.php
```diff
@@ -128,4 +132,245 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+     * center that are eligible for "promotion". Promotion is defined in this
+     * context as granting the "center staff" acl w/ associated parameter record
+     * for the promoters center.
+     *
+     * @param integer $userId the id of the user to be used when determining which
+     *                users are eligible for promotion.
+     * @return array populated if users were found, else an empty array.
+     * @throws Exception if unable to query the database.
+     */
+    public static function getUsersForPromotion($userId)
```
smgallo: What happens if the specified user is a member of (or center director of) multiple centers?

classes/Models/Services/Users.php
```diff
@@ -128,4 +132,245 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+    WHERE u.id = :user_id
+    UNION
+    SELECT DISTINCT o.id organization_id
+    FROM moddb.Users u
+      JOIN modw.person p ON p.id = u.person_id
+      JOIN modw.organization o ON o.id = p.organization_id
+      JOIN modw.resourcefact rf ON o.id = rf.organization_id
+    WHERE u.id = :user_id
+  ) co ON uo.organization_id = co.organization_id
+  -- This left join retrieves all users that have a 'cd' record and the associated center id (value)
```
smgallo: Any reason we can't just find the list of users and organizations where the user is not a center director here instead of left joining and then looking for `NULL` rows?

classes/Models/Services/Users.php
ryanrath: It would have returned users for all of the centers. I've updated the query to only retrieve the first center returned based on precedence.

classes/Models/Services/Users.php
ryanrath: Once we move to defining the organization / center for a user via User.organization_id or User.person_id.organization_id then we'll be able to simplify this logic. The problem is that currently we technically allow users to be a cs/cd for any number of centers. So not only do we have to check for the presence of a record for not 'cd' but we have to check for the presence of 'cd' as well.

classes/Models/Services/Users.php
```diff
@@ -128,4 +132,265 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+  --   - a Users.organization_id that is a valid center
+  --   - a modw.person.organization_id that is a valid center and is related via their
+  --     Users.person_id value
+  JOIN
+  (
+      -- This retrieves all of the centers for a particular user
+      SELECT DISTINCT
+        uagbp.user_id,
+        uagbp.value organization_id
+      FROM user_acl_group_by_parameters uagbp
```
smgallo: The `moddb.` got lost here.

classes/Models/Services/Users.php
```diff
@@ -128,4 +132,265 @@ public static function getUsers($userTypeId, $aclId, $searchFragment, $requireEm
...
+        $query = <<<SQL
+SELECT
+  u.id,
+  CONCAT(u.last_name, ', ', u.first_name, ' [', uo.abbrev,']') AS name
+FROM Users u
+  LEFT JOIN
+  (
+    SELECT DISTINCT uagbp.value organization_id,
+    o.abbrev
+    FROM user_acl_group_by_parameters uagbp
```
smgallo: Here too.

PR #512, opened 2018-04-17
convert some of the sql to etlv2

plessbd:
Convert `moddb` from using sql to using the etlv2 structure
~The exception to this is `modw.resourcefacts` it is currently required to exist for the `etl_overseer` to be able to run.~

This also updates the data for the `moddb` tables to json files to match other places wehere we put data in.


configuration/etl/etl_action_defs.d/jobs/xdw/schema-version-history.json
```diff
@@ -0,0 +1,8 @@
+{
+    "#": "This action loads data from a structured file and only needs the table definition",
+    "table_definition": [
+        {
+            "$ref": "${table_definition_dir}/jobs/xdw/schema-version-history.json#/table_definition"
```
smgallo: We'll have to re-examine our directory structure in the future. I think that we should head towards `<module>/<realm>/<schema>` for example `xdmod/jobs/xdw` that way we can be sure there will be no conflicts with the XSEDE bits.

configuration/etl/etl_data.d/jobs/xdw/error-descriptions.json
```diff
@@ -1,58 +1,17 @@
 [
-{
```
smgallo: You could have left this one if you wanted, `StructuredFileIngestor` should handle it. :)

configuration/etl/etl_tables.d/jobs/xdw/days.json
```diff
@@ -0,0 +1,162 @@
...
+        "nullable": false,
+        "default": null,
+        "extra": null,
+        "comment": "The id of the day record.",
+        "hints": null,
+        "rename": null,
+        "name": "id"
+      },
+      {
+        "type": "smallint(5) unsigned",
```
smgallo: FYI these definitions are different from what we have in XSEDE for the days, months, etc. tables. I don't think that will cause an issue, but we should be aware.

configuration/etl/etl_data.d/jobs/xdw/error-descriptions.json
plessbd: just wanted to be consistent with the rest of the files

configuration/etl/etl_tables.d/jobs/xdw/days.json
plessbd: interesting...  I pulled them from an openxdmod install so yeah...

plessbd:
This is going to fail until @smgallo updates etl process to auto backtick the column names

configuration/etl/etl_tables.d/xdb/chart-pool.json
```diff
@@ -0,0 +1,58 @@
...
+            {
+                "name": "insertion_rank",
+                "type": "int(11)",
+                "nullable": false,
+                "extra": "auto_increment"
+            },
+            {
+                "name": "chart_title",
+                "type": "text"
+            },
```
smgallo: While the default is "nullable": true (since mysql assumes nullable) it is nice to have it explicitly stated.

configuration/etl/etl_tables.d/xdb/report-template-acl.json
```diff
@@ -0,0 +1,18 @@
...
+    "table_definition": {
+        "name": "ReportTemplateACL",
+        "engine": "MyISAM",
+        "columns": [
+            {
+                "name": "template_id",
+                "type": "int(11)"
+            },
+            {
+                "name": "role_id",
```
smgallo: The `role_id` in the roles table is `int(4)`.

configuration/etl/etl_tables.d/xdb/users.json
```diff
@@ -0,0 +1,130 @@
...
+                "time": "BEFORE",
+                "event": "INSERT",
+                "table": "Users",
+                "body": "\n\n\tSET NEW.time_created = NOW();\n\n\tSET NEW.time_last_updated = NOW();\n\n",
+                "definer": "xdmod@localhost",
+                "schema": "moddb",
+                "name": "users_insert_timestamp"
+            },
+            {
+                "time": "BEFORE",
```
smgallo: Any idea why we are setting `time_last_updated` here and not using the timestamp column to track it?

plessbd: nope...

configuration/etl/etl_tables.d/xdb/report-template-acl.json
plessbd: cleanup comes next 👍

configuration/etl/etl_tables.d/xdb/chart-pool.json
plessbd: ok, just for you 😈

PR #518, opened 2018-04-18
fix column rename

plessbd:
`local_charge` changed to `local_charge_su`


smgallo:
Note that this is a compatibility update and we are not currently using SUs in Open XDMoD.

PR #520, opened 2018-04-18
Log a warning on drop column

smgallo:
## Description

When a table column is to be dropped by ETLv2, log a warning so the pipeline can be run in dryrun mode and verified prior to potentially losing data.

## Tests performed

Manually tested against XSEDE docker.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/DbModel/Column.php
+++ b/classes/ETL/DbModel/Column.php
@@ -34,12 +34,10 @@ class Column extends NamedEntity implements iEntity
         'default'  => null,
          // Column extra (see http://dev.mysql.com/doc/refman/5.7/en/create-table.html)
         'extra'    => null,
-        // The trigger definer for ACL purposes
+        // The column comment
         'comment'   => null,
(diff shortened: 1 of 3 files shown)


classes/ETL/DbModel/Column.php
```diff
@@ -34,12 +34,10 @@ class Column extends NamedEntity implements iEntity
...
         'extra'    => null,
-        // The trigger definer for ACL purposes
+        // The column comment
         'comment'   => null,
-        // Column hints object
-        'hints'     => null,
-        // The trigger definer for ACL purposes
-        'rename'    => null
+        // Column hints object used to control behavior such as renaming columns
+        'hints'     => null
```
plessbd: is there a list of supported hints anywhere?

smgallo: Only the `rename_from` at this point. Documentation is sorely needed but comes at the cost of the features we need. :-|

PR #521, opened 2018-04-18
Set organization for new SSO Users

ryanrath:
## Description
Updating the XDSamlAuthentication::getXdmodaccount to utilize the `organization_id` and `organization` saml attributes to determine a users organization. 

## Motivation and Context
With the update to Center User Promotion it is increasingly important that we know which organization a user is associated with ( if any ).  Previously we would only know if they were successfully matched with a person ( and that person is associated with an organization ).

## Tests performed
Manual Tests were performed ( OpenXDMoD docker w/ saml idp setup )

- Attempt to login as a user w/ an organization_id, organization and system username that will not be found in the db.
- After successfully login check the db and ensure that the User record created shows no associated person or organization.

- Attempt to login as a user w/ a valid organization_id, invalid organization and valid system username
- After successfully logging in check the db to ensure that the User record created shows an associated person and an associated organization.

- Attempt to login as a user w/ an invalid organization_id, valid organization and valid system username
- After successfully logging in check the db to ensure that the User record created shows an associated person and an associated organization.

## Types of changes

## Checklist:

### DIFF
--- a/classes/Authentication/SAML/XDSamlAuthentication.php
+++ b/classes/Authentication/SAML/XDSamlAuthentication.php
@@ -2,8 +2,10 @@
 
 namespace Authentication\SAML;
 
+use CCR\MailWrapper;
 use \Exception;
 use CCR\Log;
+use Models\Services\Organizations;
 
 class XDSamlAuthentication
 {
@@ -34,7 +36,33 @@ class XDSamlAuthentication
      * @var boolean
      */
     protected $_allowLocalAccessViaSSO = true;
+
+    /**
+     * The name of the organization to be used when `force_default_organization` is true.
+     *
+     * @var string
+     */
+    protected $_defaultOrganizationName;
+
+    /**
+     * Always use the organization identified by `default_organization_name` when assigning an
+     * organization to a new SSO user.
+     *
+     * @var bool
+     */
+    protected $_forceDefaultOrganization;
+
+    /**
+     * Controls when the admins are notified when an organization cannot be identified for a new SSO
+     * user.
+     *
+     * @var bool
+     */
+    protected $_emailAdminForUnknownUserOrganization;
+
     private $logger = null;
+
+    private $emailLogger = null;
 
     public function __construct()
     {
@@ -45,6 +73,18 @@ class XDSamlAuthentication
                'db' => true,
                'mail' => false,
                'console' => false
+            )
+        );
+        $this->emailLogger = Log::factory(
+            'XDSamlAuthentication-email',
+            array(
+                'file' => false,
+                'db' => false,
+                'console' => false,
+                'mail' => true,
+                'emailTo' => \xd_utilities\getConfiguration('general', 'tech_support_recipient'),
+                'emailFrom' => \xd_utilities\getConfiguration('mailer', 'sender_email'),
+                'emailSubject' => 'XDMoD SSO: Additional Actions Necessary'
             )
         );
         $this->_sources = \SimpleSAML_Auth_Source::getSources();
@@ -64,6 +104,10 @@ class XDSamlAuthentication
                 $this->_as = new \SimpleSAML_Auth_Simple($this->_sources[0]);
             }
         }
+
+        $this->_defaultOrganizationName = \xd_utilities\getConfiguration('sso', 'default_organization_name');
+        $this->_forceDefaultOrganization = \xd_utilities\getConfiguration('sso', 'force_default_organization') === 'on';
+        $this->_emailAdminForUnknownUserOrganization = \xd_utilities\getConfiguration('sso', 'email_admin_sso_unknown_org') === 'on';
     }
 
     /**
@@ -81,6 +125,7 @@ class XDSamlAuthentication
      * Attempts to find a valid XDMoD user associated with the attributes we receive from SAML
      *
      * @return mixed a valid XDMoD user if we have one, false otherwise
+     * @throws Exception
      */
     public function getXdmodAccount()
     {
@@ -110,6 +155,16 @@ class XDSamlAuthentication
             }
             $emailAddress = !empty($samlAttrs['email_address'][0]) ? $samlAttrs['email_address'][0] : NO_EMAIL_ADDRESS_SET;
             $personId = \DataWarehouse::getPersonIdByUsername($thisSystemUserName);
+
+            $samlOrganization = $samlAttrs['organization'];
+            if ($this->_forceDefaultOrganization) {
+                $userOrganization = Organizations::getIdByName($this->_defaultOrganizationName);
+            } elseif (!empty($samlOrganization)) {
+                $userOrganization = Organizations::getIdByName($samlOrganization[0]);
+            } else {
+                $userOrganization = -1;
+            }
+
             if (!isset($samlAttrs["first_name"])) {
                 $samlAttrs["first_name"] = array("UNKNOWN");
             }
@@ -129,7 +184,7 @@ class XDSamlAuthentication
                     $samlAttrs["last_name"][0],
                     array(ROLE_ID_USER),
                     ROLE_ID_USER,
-                    null,
+                    $userOrganization,
                     $personId
                 );
             } catch (Exception $e) {
@@ -142,7 +197,7 @@ class XDSamlAuthentication
                 $this->logger->err('User creation failed: ' . $e->getMessage());
                 return false;
             }
-            self::notifyAdminOfNewUser($newUser, $samlAttrs, ($personId != UNKNOWN_USER_TYPE));
+            self::notifyAdminOfUserWithUnknownOrganization($newUser, $samlAttrs, ($personId != UNKNOWN_USER_TYPE));
             return $newUser;
         }
         return false;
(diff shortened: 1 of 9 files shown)


classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -110,6 +127,31 @@ public function getXdmodAccount()
...
             $personId = \DataWarehouse::getPersonIdByUsername($thisSystemUserName);
+
+            if (!isset($samlAttrs['organization_id'])) {
+                $userOrganization = null;
+            } else {
+                // Attempt to look up their organization by the most reliable method we have
+                // currently.
+                $userOrganization = Organizations::getOrganizationIdByName($samlAttrs['organization_id'][0]);
+                if (!empty($userOrganization)) {
+                    $userOrganization = $userOrganization[0]['id'];
```
jpwhite4: The logic to decode the (database specific) data format belongs in the getOrganizationIdByName() function not in this code.

The getOrganizationIdByName() and getOrganizationIdByLongName should return an Id or null if no match rather than a nested array of stuff.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -47,6 +52,18 @@ public function __construct()
...
         );
+        $this->emailLogger = Log::factory(
+            'XDSamlAuthentication-email',
+            array(
+                'file' => false,
+                'db' => false,
+                'console' => false,
+                'mail' => true,
+                'emailTo' => \xd_utilities\getConfiguration('general', 'tech_support_recipient'),
+                'emailFrom' => 'sso-user-creation@ccr.xdmod.org',
```
jpwhite4: I don't think we should hardcode the ccr.xdmod.org domain name here. It is fine for our instance but does not make sense for the open source version

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -110,6 +127,21 @@ public function getXdmodAccount()
...
+                $userOrganization = null;
+            } else {
+                // Attempt to look up their organization by the most reliable method we have
+                // currently.
+                $userOrganization = Organizations::getOrganizationIdByName($samlAttrs['organization_id'][0]);
+
+                // If we didn't find their organization via organization_id then fall back to trying
+                // organization.
+                if ($userOrganization === null && isset($samlAttrs['organization'])) {
+                    $userOrganization = Organizations::getOrganizationIdByLongName($samlAttrs['organization'][0]);
```
jpwhite4: Is it possible for `$samlAttrs['organization_id']` to not be set but `$samlAttrs['organization']` to be set?

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -199,7 +231,27 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+        $userOrganization = $user->getOrganizationId();
+
+        // If userOrganization is null -or- if the value < 1 ( 0 )
+        // -and-
+        // we have an organization_id provided by the samlAttributes
+        // -then-
+        // this is an error case and we need to notify the admins that there will
+        // be additional setup required.
+        $unableToFindOrganization =
+            (!isset($userOrganization) || $userOrganization < 1) &&
```
jpwhite4: What is the significance of `$userOrganization < 1`

classes/Authentication/SAML/XDSamlAuthentication.php
ryanrath: Good question, I'll check with @chakrabortyr && @plessbd

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -110,6 +127,21 @@ public function getXdmodAccount()
...
+                // currently.
+                $userOrganization = Organizations::getIdByName($samlAttrs['organization_id'][0]);
+            }
+
+            // If we didn't find their organization via organization_id then fall back to trying
+            // organization.
+            if ($userOrganization === null && isset($samlAttrs['organization'])) {
+                $userOrganization = Organizations::getOrganizationIdByLongName($samlAttrs['organization'][0]);
+            }
+
```
jpwhite4: This does not appear to handle the case where both organization_id and organization are set but organization_id is not a known organization in the database and organization is in the database.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -211,5 +263,36 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
         } else {
             $this->logger->notice("New " . ($linked ? "linked": "unlinked") . " Single Sign On user created" . $body);
         }
+
+        if ($unableToFindOrganization) {
```
jpwhite4: Isn't this going to spam the admins of all Open XDMoD software instances whenever a user logs in?

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -211,5 +263,36 @@ private function notifyAdminOfNewUser($user, $samlAttributes, $linked, $error =
...
+            $this->emailLogger->notice("Additional SSO User Action Required". $body);
+
+            $emailAddress = $user->getEmailAddress();
+            if (!isset($emailAddress)) {
```
chakrabortyr: Do we really need three !isset checks on the same variable? You could condense this to something like:
```

$userEmail = $user->getEmailAddress();
$emailAddress = $userEmail ? $userEmail : $samlAttributes['email_address'][0];

if (!isset($emailAddress)) {
 // error here
}

```

To elaborate, the only SAML attribute we care about with respect to emails is "email_address". The "email" attribute is provided to us by another IDP (Globus in this case), we then map it over to "email_address". As the former isn't even guaranteed to exist, we probably don't need to check for it.

classes/Authentication/SAML/XDSamlAuthentication.php
ryanrath: As written, it definitely does not. I've updated the default values / values checked to be -1 ( vs null ). Here's a test script I tossed together to make sure things are copacetic 

```php
<?php

function getIdByName($name)
{
    return $name === 'Screwdriver' ? 1 : -1;
}

function getIdByLongName($longName)
{
    return $longName === 'screw - Screwdriver' ? 1 : -1;
}


$samlAttrs = array(

    # Not in OpenXDMoD DB
    'organization_id' => array('University at Buffalo'),

    # In OpenXDMoD DB
    'organization'    => array('screw - Screwdriver')
);

/*
 * These next two blocks of logic are right from the code.
 * I have updated the default value / 
 */
if (!isset($samlAttrs['organization_id'])) {

   // *** This was updated *** 
    $userOrganization = -1;
} else {
    // Attempt to look up their organization by the most reliable method we have
    // currently.
    $userOrganization = getIdByName($samlAttrs['organization_id'][0]);
}

// If we didn't find their organization via organization_id then fall back to trying
// organization.
// *** This was updated ***
if ($userOrganization === -1 && isset($samlAttrs['organization'])) {
    $userOrganization = getIdByLongName($samlAttrs['organization'][0]);
}

echo sprintf("%s\n", $userOrganization);
```

I would expect to receive `1` and that seems to be what I receive. Let me know if I've missed something.

Thanks!

classes/Authentication/SAML/XDSamlAuthentication.php
ryanrath: Certainly, particularly if we don't care about `email` ( Didn't know that at the time of writing). Hopefully you can understand why I went with this particular layout as opposed to something like: 
```php
$emailAddress = $user->getEmailAddress()
                ? $user->getEmailAddress()
                : $samlAttributes['email_address'][0]
                  ? $samlAttributes['email_address'][0]
                  : $samlAttributes['email'][0]
                    ? $samlAttributes['email'][0]
                    : null;
if (!isset($emailAddress)) {
 // error here
}

```

classes/Authentication/SAML/XDSamlAuthentication.php
ryanrath: @jpwhite4 Just talked with @smgallo . Would adding a configuration option so that admins can opt-out of email notification sound good?

ryanrath: @jpwhite4 I've added the configuration option ( on by default ) in the latest commit.

ryanrath:
@chakrabortyr We had a quick pow wow about things this morning. I've just committed the updates that came out of it. We're still going to need to update our auth-procs to make sure that the 'organization'  saml attribute is present.

ryanrath:
I have also updated our ccr private repo w/ the new SSO properties so any deployments with those will not break as soon as this is merged.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -34,22 +37,94 @@ class XDSamlAuthentication
...
-               'db' => true,
-               'mail' => false,
-               'console' => false
+                'file' => false,
+                'db' => true,
+                'mail' => false,
+                'console' => false
+            )
+        );
+        $this->emailLogger = Log::factory(
```
plessbd: Per discussion emailLogger v MailWrapper

Dockerfile
```diff
@@ -1 +1 @@
-FROM       tas-tools-ext-01.ccr.xdmod.org/xdmod-centos7:open7.5.1-v4
```
jpwhite4: The problem with having a script that writes to a randomly generated file is that it will be impossible to write a test that checks that the email was generated correctly (because you have no way of determining the file that contains the contents of the email). Why not have a file that is name deterministically - perhaps using the sender and subject line of the email in the filename.

ryanrath: Hrmm, I was kinda hoping to not have to parse whatever is passed in. But it looks like the only two fields that are 'required' in an email are `From:` and `Date:` ( per RFC 5322 sections: 3.6 [ 3.6.1 & 3.6.2 ] https://tools.ietf.org/html/rfc5322#section-3.6 ). I can certainly see about parsing those values out for use in the file name. 

If we're not concerned with whether or not we'll have a value supplied for one of the fields used in the file name then from / sender + subject could work. Do you have a preference on default value if subject isn't found?

ryanrath: Actually, we could start with subject and if one is not provided then fallback to `Date:`

jpwhite4: I think that subject will always be present in our use case (there are no code paths in XDMoD where the subject is not set). Anyway if you want to keep it simple you could just use a deterministic filename such as /tmp/mailoutput and append to it. Remember the idea is to allow us to write tests that check that the mail sending works properly.

ryanrath: I am certainly all about having the ability to test that the emails we send out are sent as we expect them to be, but that wasn't the primary reason of this change. The primary reason is that  without it ( or updating our docker image so that sendmail doesn't error out when trying to use it ) + sending emails during SSO login == broken SSO log in which in turn means that the SSO tests never finish and the shippable builds can't pass.

ryanrath: Alrighty, just had a quick conversation with @smgallo about this. How does this sound: 
  - Emails are saved to `/tmp/sent_mail/<random_#>.eml`
  - Meta data about the email is appended to `/tmp/sent_mail/mail.idx` ( name's not important, just that it's at a constant location ). 
    - Tab delimited: `<from value>    <subject value>    <path/to/email/file.eml>`

This way we don't have to worry about what characters are in the from / subject values ( like `/` that need to be escaped ). There's a consistent location to retrieve information about which emails were written where ( or not ) and we can get at the full contents of the email if needed.

jpwhite4: Sounds a bit over complicated to me - I would have just done:

```bash
#!/bin/bash
cat > /tmp/mailoutput
```

ryanrath: Do you only care about the last email sent?

ryanrath: I mean, that certainly makes things easier. Just want to make sure that's the case before I change things.

jpwhite4: ok then:
```bash
#!/bin/bash
cat >> /tmp/mailoutput
```

plessbd: or
```
yum install postfix -y
yum remove ssmtp -y

sed -ie 's/inet_interfaces = localhost/#inet_interfaces = localhost/' /etc/postfix/main.cf
sed -ie 's/smtp      inet  n       -       n       -       -       smtpd/#smtp      inet  n       -       n       -       -       smtpd/' /etc/postfix/master.cf
sed -ie 's/smtp      unix  -       -       n       -       -       smtp/smtp      unix  -       -       n       -       -       local/' /etc/postfix/master.cf
sed -ie 's/relay     unix  -       -       n       -       -       smtp/relay     unix  -       -       n       -       -       local/' /etc/postfix/master.cf

echo '/.*/ root' >> /etc/postfix/virtual
echo 'virtual_alias_maps = regexp:/etc/postfix/virtual' >> /etc/postfix/main.cf
newaliases
```

then edit ~/bin/services to start /usr/sbin/postfix and all mail will get delivered to /var/mail/root

plessbd: As added bonus `/var/mail/<user>` has a standard format: https://en.wikipedia.org/wiki/Mbox

Which we should then be able to use "off the shelf" tools to parse
https://pear.php.net/manual/en/package.mail.mail-mbox.php

ryanrath: Alrighty, I've removed the commits that updated the tests / shippable Dockerfile as they were updated in https://github.com/ubccr/xdmod/pull/647.  These changes now mean that the tests pass / shippable build completes. We should be good to go now. 

@smgallo @jpwhite4 @plessbd

ryanrath:
Updated Documentation included in https://github.com/ubccr/xdmod/pull/648

Dockerfile
```diff
@@ -1 +1 @@
...
+FROM       tas-tools-ext-01.ccr.xdmod.org/xdmod-centos7:open7.5.1-v5
```
plessbd: This shouldn't need to be done, since "master" has been updated this should disappear after a rebase / merge

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -64,6 +126,10 @@ public function __construct()
                 $this->_as = new \SimpleSAML_Auth_Simple($this->_sources[0]);
             }
         }
+
+        $this->_defaultOrganizationName = ORGANIZATION_NAME;
```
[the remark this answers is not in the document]
ryanrath: It's not strictly needed. I'm not sure I like it from a code readability standpoint but it would obviously be less code overall.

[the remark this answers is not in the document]
smgallo: If these are only used in `getOrganizationId()` then they should be local to that method.

ryanrath:
@smgallo @plessbd @jpwhite4 , updates per our conversations today are up.

PR #531, opened 2018-05-04
Add foreign key constraints to ETL

jtpalmer:
## Description

Adds foreign key constraint support to ETL.

## Motivation and Context

Tables in `mod_hpcdb` have foreign key constraints.  In order to improve job ingestion these tables should be managed by the ETL process.

## Tests performed

Updated ETL tests.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/DbModel/ForeignKeyConstraint.php
+++ b/classes/ETL/DbModel/ForeignKeyConstraint.php
@@ -57,7 +57,7 @@ class ForeignKeyConstraint extends NamedEntity implements iEntity
     public function initialize(stdClass $config)
     {
         if (!isset($config->name)) {
-            $config->name = $this->generateConstraintName($config->columns);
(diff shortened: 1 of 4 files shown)


classes/ETL/DbModel/ForeignKeyConstraint.php
```diff
@@ -115,22 +151,26 @@ protected function filterAndVerifyValue($property, $value)
...
     {
         $str = implode('_', $columns);
         $name = ( strlen($str) <= 32 ? $str : md5($str) );
+
         return 'constraint_' . $name;
     }
 
     /**
-     * Constraints are considered equal if all properties are the same.
+     * Constraints are considered equal if all non-null properties are the same.
```
smgallo: Something that we'll need to deal with is keywords in different case because as it stands now (for all of the `DbModel` classes) "FROM" and "from" are considered different. Don't worry about it here, I'm going to think about how to handle it globally and change them all at once.

PR #533, opened 2018-05-04
Remove javscript from php chart code

jpwhite4:
The javascript functions needed for the Highcharts charts are now defined in javascript rather than in php. The API calls that return jchart data now return properly encoded json rather javascript.

The regression tests have been updated: there was a bug in the previous javascript encoding code that did not handle data with single quote characters. The buggy code is removed by this change.

Note that where possible I have tried to keep the existing javascript code in place unmodified (to reduce the chance of inadvertent regressions).


jpwhite4:
Note that the unit test failure is caused because the test artefacts were changed to support #532 and not due to problems with this code change.

PR #535, opened 2018-05-07
Add submission venue for events

eiffel777:
Added submission_venue_id column to event table and for open stack ingestion set the submission_venue_id to be 3 which matches the row in the submission_venue table for data ingested from the OpenStack API.

## Tests performed
Tested with docker.

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_action_defs.d/cloud_openstack/event.json
+++ b/configuration/etl/etl_action_defs.d/cloud_openstack/event.json
@@ -18,7 +18,8 @@
             "event_type_id": "staging.event_type_id",
             "record_type_id": "staging.record_type_id",
             "host_id": "staging.host_id",
-            "image_id": "staging.image_id"
+            "image_id": "staging.image_id",
+            "submission_venue_id": 3
         },
 
         "joins": [
@@ -45,7 +46,8 @@
             "event_time_utc": "event_time_utc",
             "event_type_id": "event_type_id",
             "record_type_id": "record_type_id",
-            "host_id": "host_id"
+            "host_id": "host_id",
+            "submission_venue_id": "submission_venue_id"
         }
     }
 }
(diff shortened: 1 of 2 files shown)


configuration/etl/etl_action_defs.d/cloud_openstack/event.json
```diff
@@ -18,7 +18,8 @@
             "event_type_id": "staging.event_type_id",
             "record_type_id": "staging.record_type_id",
             "host_id": "staging.host_id",
-            "image_id": "staging.image_id"
+            "image_id": "staging.image_id",
+            "submission_venue_id": 3
```
jpwhite4: I prefer to avoid hard-coding database-table-specific primary key values. Is it possible to look this value up in the submission_venue table rather than hardcoding it.

jpwhite4:
Should this code be put in the xsede module rather than the core? The sbumission_table does not exist in the open source release.

configuration/etl/etl_action_defs.d/cloud_openstack/event.json
smgallo: We can join and use the venue name, although that's still hard-coding just a string instead of an integer.

smgallo:
It is in the open source release. See `configuration/etl/etl_tables.d/jobs/submission_venue.json` and `configuration/etl/etl_data.d/jobs/submission_venue.json`.

configuration/etl/etl_action_defs.d/cloud_openstack/event.json
eiffel777: The id for a submission_venue_id is hardcoded in other places like configuration/etl/etl_action_defs.d/jobs/xdw/job-records-job-tasks.json and places in the xsede module so I thought I would keep the consistency.

PR #539, opened 2018-05-08
Stop using proc_open for calling etl, instead invoke directly

plessbd:
When using proc_open and ignoring things you can get into a situation where the process gets blocked.

In this case call the etlv2 directly instead and avoid the issue all together.

### DIFF
--- a/classes/OpenXdmod/Setup/DatabaseSetup.php
+++ b/classes/OpenXdmod/Setup/DatabaseSetup.php
@@ -8,6 +8,7 @@ namespace OpenXdmod\Setup;
 use TimePeriodGenerator;
 use CCR\DB\MySQLHelper;
 use CCR\DB;
+use CCR\Log;
 
 /**
  * Database setup.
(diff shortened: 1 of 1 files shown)


classes/OpenXdmod/Setup/DatabaseSetup.php
```diff
@@ -132,58 +140,17 @@ public function handle()
...
-            $this->console->displayMessage("\t" . $command);
-            $this->console->displayBlankLine();
-            $this->console->displayMessage('Details:');
-            $this->console->displayMessage(print_r(error_get_last(), true));
-            $this->console->displayBlankLine();
 
-            $this->console->prompt('Press ENTER to continue.');
+        $etlConfig = new \ETL\Configuration\EtlConfiguration(CONFIG_DIR . '/etl/etl.json', null, $logger, array());
+        $etlConfig->initialize();
+        \ETL\Utilities::setEtlConfig($etlConfig);
```
smgallo: I really need to find a good way to eliminate this requirement. Wonder if I can do it in the overseer itself...

PR #552, opened 2018-05-18
Globalize Query Descripters

ryanrath:
Migrating the `aRole::hasDataAccess` and `aRole::getQueryDescripters` functions out of `aRole.php` and into a service class. These new functions utilize acl information from the database & a few pieces of information from config files to produce the same output as the old functions. Each call site for the old functions was updated to utilize the new functions.

Besides being part of the Acl migration, these changes allow non-feature, non-hierarchical acls to provide users access to new / different realm / group by  information.

## Motivation and Context

A user was having problems granting access to a new realm via a new acl ( as opposed to updating the existing acls w/ access ). Upon investigation it was found that the source of the problem stemmed from the code generating the UI output. It was only taking the users most privileged acl into consideration instead of all of a users acls. 

## Tests performed

- New Component Tests were added for each function. These tests directly compare the output of the `aRole` functions w/ the `Acls` functions.
- All Unit tests were run
- All Component tests were run
- All Integration tests were run
- All ( REG_TEST_ALL="1" ) regression tests were run for all users.
- All Automated UI tests were run

## Types of changes

## Checklist:


classes/Models/Services/Acls.php
```diff
@@ -906,4 +905,254 @@ public static function getAclsByTypeName($aclTypeName)
...
+                    'tg_usage',
+                    $row['realm'],
+                    $row['group_by']
+                );
+
+                if (isset($row['`show`'])) {
+                    $descripter->setShowMenu($row['show']);
+                }
+
+                if (isset($row['`disable`'])) {
```
jpwhite4: Are these backticks needed here and on line 978?

ryanrath: Yup, without the back ticks the component tests fail. Now, I'm not sure why it they aren't needed when retrieving the results. But they definitely fail without being defined ( I also tried adding them to the retrieval code and the tests still pass ).

ryanrath: I'm going to do some digging real quick into what's actually being returned. I'll update with what I find.

html/controllers/user_interface/get_menus.php
```diff
@@ -89,46 +89,55 @@
...
-                            'chartSettings'        => $query_descripter->getChartSettings(true),
-                            'node_type'            => 'group_by',
-                            'iconCls'              => 'menu',
-                            'description'          => $query_descripter->getGroupByLabel(),
-                            'leaf'                 => false,
+                            'chartSettings' => $query_descripter->getChartSettings(true),
+                            'node_type' => 'group_by',
+                            'iconCls' => 'menu',
+                            'description' => $query_descripter->getGroupByLabel(),
+                            'leaf' => false,
```
smgallo: Remove trailing comma here.

jpwhite4:
This pull request is pointing at the wrong branch. Needs to be xdmod8.0

classes/Models/Services/Acls.php
```diff
@@ -906,4 +905,252 @@ public static function getAclsByTypeName($aclTypeName)
...
+        // 'enabled' / 'hidden' from each acls query_descripter.
+        if (count($rows) > 0) {
+            foreach ($rows as $row) {
+                $descripter = new QueryDescripter(
+                    'tg_usage',
+                    $row['realm'],
+                    $row['group_by']
+                );
+
+                if (isset($row['not_enabled'])) {
```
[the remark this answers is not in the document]
ryanrath: Yup, pushed in the latest commit.

classes/Models/Services/Acls.php
```diff
@@ -906,4 +905,248 @@ public static function getAclsByTypeName($aclTypeName)
...
+            $params[':group_by_name'] = $groupByName;
+        }
+
+        if (isset($statisticName)) {
+            $query .= " AND s.name = :statistic_name\n";
+            $params[':statistic_name'] = $statisticName;
+        }
+
+        // Need to add the group by so that we only get one row per realm / group_by
+        $query .= "GROUP BY r.display, gb.name";
```
jpwhite4: You are grouping by display and name columns but the no aggregation function on the not_enabled column. This is not legal SQL and is a non-standard extension in MySQL.

ryanrath: Updated in latest commit.

ryanrath: I'm also envisioning the SQL police attempting to arrest me for illegal SQL use.

PR #559, opened 2018-05-25
Adding tests artifacts

ryanrath:
## Description
Just adding some test artifacts that were added in between the moving of the
artifacts into the main repo.

## Motivation and Context
Having test artifacts

## Types of changes

## Checklist:


plessbd:
is there a way to easily generate the input rather than storing the files.

Joe did some neat stuff (https://github.com/ubccr/xdmod/blob/xdmod8.0/open_xdmod/modules/xdmod/regression_tests/lib/TestHarness/Utilities.php & https://github.com/ubccr/xdmod/blob/0e4672b0606772dfdb0705fe933cbb5c1f259279/open_xdmod/modules/xdmod/regression_tests/lib/Controllers/UsageChartsTest.php#L143) to build all the combinations given the values. Seems a little cleaner than having all the input files.  Understandable if not able to be done easily just a "nice to have"

plessbd:
Github is saying a bunch of the "json" files are actually binary: acls/output/get_query_descripters-jobs_none_total_node_hours.json should we change then from .json to something else in this case?

ryanrath:
Not sure if I get what you mean by generate all the input files? It looks like there's only 4 input files. A metric ton of output files, but very few input.

jpwhite4:
What tests use these artefacts?

ryanrath:
These test artifacts are being introduced in PR 560: https://github.com/ubccr/xdmod/pull/560 instead. This way they can be added along with their respective tests.

PR #561, opened 2018-05-31
Updating sso login form action

ryanrath:
sipmlesaml-idp updated the action that triggers a sign on so the test needed to
be updated.

## Description
Updated the expected sign in form action for the SSO tests.

## Motivation and Context
This change was necessary because of this change in the saml-idp repo (https://github.com/mcguinness/saml-idp/commit/03d6ca053dd6b1278c75ce7e1e762ff7912fe8a1) 

## Tests performed
The saml tests were re-run after the modification was made.

## Types of changes

## Checklist:


plessbd:
IN an ideal world we would update the stuff that brings in saml-idp to be set at a specific version so this wont happen again, but in the need for this to allow tests to pass I approves

plessbd:
added commit to force to a specific version of saml-idp

PR #562, opened 2018-05-31
Fix calendar quarter calculation

jtpalmer:
## Description

Bug fix.

## Motivation and Context

The start date of calendar quarters was being calculated incorrectly if the current day of the month is not in the previous month (e.g. 31 May since April only has 30 days).

## Tests performed

No new tests.

## Types of changes

## Checklist:


plessbd:
In no place should we have indentation of 7 spaces, so I ignore travis.  As long as this passes to the point of the report generator and fails on the saml I think this is good to go.

PR #572, opened 2018-06-06
Replace Role Parameters Usage

ryanrath:
## Description
Removed usage of `aRole::getParameters` and replaced it with a service function `Parameters::getParameters`. 

## Motivation and Context
Moving towards removal of Roles.

## Tests performed
All automated tests were performed / passed as well as manual testing

## Types of changes

## Checklist:

**NOTE: Sorry in advance for the set of double commits. This shouldn't be a problem once everything is squished / merged.**


classes/Models/Services/Parameters.php
```diff
@@ -0,0 +1,62 @@
...
+                    $organizationId = $user->getOrganizationID();
+                    if ($organizationId === null ||
+                        in_array((int)$organizationId, array(UNKNOWN_ORGANIZATION_ID, 0))
+                    ) {
+                        $organizationId = Organizations::getOrganizationForUser($user->getUserID());
+                    }
+                    $parameters['provider'] = (string)$organizationId;
+                    break;
+                case 'person':
+                    $parameters['person'] = (string)$user->getPersonID();
```
jpwhite4: What is the reason for the explicit string cast here?

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/UserHelper.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace TestHarness;
+
+use XDUser;
+
+class UserHelper
```
jpwhite4: Is this class really needed? It is only called from one place with one set of defined parameters.

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/UserHelper.php
```diff
@@ -0,0 +1,45 @@
...
+        $firstName,
+        $middleName,
+        $lastName,
+        array $acls = null,
+        $primaryRole = null,
+        $email = null,
+        $organizationId = null,
+        $personId = null
+    ) {
+        $acls = isset($acls) ? $acls : array(ROLE_ID_USER);
```
jpwhite4: This code is never called.

jpwhite4:
Are there any corresponding changes needed for the AppKernels, SUPReMM and XSEDE modules?

classes/Models/Services/Parameters.php
ryanrath: So that the output of this function matches the output of `aRole::getParameters`. `aRole::getParameters` returns records from `UserRoleParameters` and specifically the `param_value` column which is defined as `varchar(1024)`.

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/UserHelper.php
ryanrath: Just updated this in the latest commit. I had forgotten to replace the other code that could utilize this helper function.

classes/Models/Services/Parameters.php
smgallo: Unless you're doing a `===` comparison between the two values and one is an integer while the other is a string this should not be needed, should it?

open_xdmod/modules/xdmod/integration_tests/lib/TestHarness/UserHelper.php
ryanrath: Updated in latest commit, the code / tests in XDUserTest do sometimes call this function without an $acls parameter.

classes/Models/Services/Parameters.php
ryanrath: True, which is what we're doing in the tests. I wanted to make sure that the new output exactly matched the old.

ryanrath:
@jpwhite4 Not that I've found ( executed at root of XDMoD working copy with all modules checked out at xdmo8.0): 
```
grep -iRn "\->getParameters" . --exclude-dir=vendor --exclude-dir=docs
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:63:                "\n[%s] Failed asserting that role->getParameters === Parameters::getParameters.\nExpected: %s\nActual: %s",
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:85:     * Test that the output of Parameters::getParameters === role->getParameters where:
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:204:                    "Expected that the output from role->getParameters would match the output from Parameters::getParameters"
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:415:                    "\t\$role->getParameters:              %s",
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:418:                    "\tExpected \$role->getParameters:     %s",
./open_xdmod/modules/xdmod/component_tests/lib/Roles/RoleParametersTest.php:421:                    "\t\$role->getParameters equals:        %s",
./classes/DataWarehouse/Query/Query.php:1127:                $rolewheres = $this->getParameters($group_by_instance->pullQueryParameters($param));

```

ryanrath:
@jpwhite4 Scratch that, I did find some updates that need to occur. I'm putting the branches / PR's together right now. They are: 
- SUPREMM: `classes/DataWarehouse/Query/SUPREMM/JobMetadata:310` a call to `setMultipleRoleParameters` needed to have an `XDUser` instance added as a parameter.
- XSEDE: `classes/DataWarehouse/Visualization/HighChartCustomQuery` the constructor needed to have an XDUser parameter added to the end / piped to the parent constructor.
- AppKernels: `classes/DataWarehouse/Visualization/HighChartAppKernel` the constructor needed to have an XDUser parameter added to the end / piped to the parent constructor.

ryanrath:
@jpwhite4 I have all of the accompanying PR's up now.

classes/DataWarehouse/Access/MetricExplorer.php
```diff
@@ -365,10 +367,12 @@ private function getDataSeries()
             foreach ($v as $x) {
                 $y = (object)$x;
 
+                // Ensure that each filter->data entry is an object
                 for ($i = 0, $b = count($y->filters['data']); $i < $b; $i++) {
                     $y->filters['data'][$i] = (object)$y->filters['data'][$i];
                 }
 
+                // Ensure that $y->filters is an object
```
jpwhite4: How come this comment was added? Is the casting of y->filters needed for the roles changes? Also this comment provides no useful information - it is quite obvious that the y->filters is being cast to an object in the code. Ideally the comment would explain **why** this is being done.

ryanrath:
@jpwhite4 Comments removed in the latest commit. I'll also make sure to resolve the conflicts

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,71 @@
...
+SELECT src.organization_id
+    FROM (
+      SELECT user_org.*
+      FROM (
+        SELECT DISTINCT
+          uagbp.value organization_id,
+          1
+        FROM moddb.user_acl_group_by_parameters uagbp
+        JOIN modw.organization o ON uagbp.value = o.id
+        JOIN modw.resourcefact rf ON o.id = rf.organization_id
```
jpwhite4: Why is resourcefact joined here?

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,71 @@
...
+          ON uagbp.group_by_id = gb.group_by_id AND gb.name = 'provider'
+        JOIN moddb.acls a ON uagbp.acl_id = a.acl_id AND a.name = 'cd'
+        WHERE uagbp.user_id = :user_id
+        UNION
+        SELECT DISTINCT
+          u.organization_id,
+          2
+        FROM moddb.Users u
+        JOIN modw.organization o ON o.id = u.organization_id
+        JOIN modw.resourcefact rf ON o.id = rf.organization_id
```
jpwhite4: Why is resourcefact joined here?

classes/Models/Services/Parameters.php
```diff
@@ -0,0 +1,62 @@
...
+            $aclConfig = $dimensions = $roles[$aclName];
+            $dimensions = isset($aclConfig['dimensions']) ? $aclConfig['dimensions'] : array();
+        } catch (\Exception $e) {
+            throw new \Exception("Unable to retrieve dimension information about $aclName");
+        }
+
+        foreach ($dimensions as $dimension) {
+            switch ($dimension) {
+                case 'provider':
+                    $organizationId = $user->getOrganizationID();
```
jpwhite4: Why do we have a getOrganzationID() function for the user class (that presumably gets the organization id for a user) and a saperate function Organizations::getOrganizationForUser that gets the organization id for a user? Under what circumstances does user->getOrganizationID() return null?

ryanrath: The function: `XDUser::getOrganizationID` returns the value of `moddb.Users.organization_id`. It returns null when a User is created and an 'Organization' is not set in the Create User UI ( or supplied to the XDUser constructor ). Previously the `moddb.Users.organization_id` column was only populated when the user was assigned the Campus Champion acl but we decided to open it up for all users as a more efficient alternative to tracking Organization association via the `UserRoleParamters` / `user_acl_group_by_parameters` tables. This is mainly due to redefining that a user may only be associated with a single organization at a time.

classes/Models/Services/Organizations.php
ryanrath: So the reason that `resourcefact` is joined in is that originally the value returned needed to be a "Center". Which was defined as ( as I remember it ): an Organization that had a corresponding entry in `resourcefact`. Granted, looking back on this, some combination of: renaming and or moving the function should probably occur.

classes/Models/Services/Parameters.php
```diff
@@ -0,0 +1,62 @@
...
+                    ) {
+                        $organizationId = Organizations::getOrganizationForUser($user->getUserID());
+                    }
+                    $parameters['provider'] = (string)$organizationId;
+                    break;
+                case 'person':
+                    $parameters['person'] = (string)$user->getPersonID();
+                    break;
+                case 'pi':
+                    $parameters['pi'] = Users::isPrincipalInvestigator($user) ? (string)$user->getPersonID() : -1;
```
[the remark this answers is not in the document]
ryanrath: @jpwhite4 After looking through previous versions of the code the only possibly reason I can think that -1 was being used as a default was to fulfill the ternary. Which, doesn't really need to be there because we shouldn't be doing authorization at this level. I'll be pushing an updated version in just a moment ( it will also include the simplification of the 'provider' case that we talked about ).

ryanrath:
per a conversation with @jpwhite4, I'm going to close this PR as I've incorporated the changes from this PR into https://github.com/ubccr/xdmod/pull/608 . That way we only have one PR to test / deal with.

PR #589, opened 2018-06-15
Preserve order of table columns as defined in JSON config file

smgallo:
## Description

When modifying an existing table, the current ETL adds newly defined columns to the end of the table. This modification preserves the ordering of the columns as specified in the JSON configuration file by using `FIRST` and `AFTER` clauses in the `ALTER TABLE` statement.

Also included as a part of this functionality is the ability to examine an existing table and re-order the rows to match the column array order of the JSON table definition (suggested by @jpwhite4).

## Motivation and Context

Cleaner table definitions to compare to JSON configuration.

## Tests performed

Ran an XSEDE ingestion against docker.

Added tests to cover the following use cases, starting with a newly created table:
- Group 1
    1. Add a new column at the start of the list
    2. Add a new column in the middle of the list
    3. Delete a column
    4. Add a new column after the deleted column
    5. Rename a column
    6. Add a new column after the renamed column
    7. Swap the order of awarded and recommended
- Group 2
    1. Reorder the first column in the table
    2. Reorder the last 2 columns in the table
    3. Reorder columns in the middle of the table
    4. Alter the definition of one of the re-ordered columns
- Group 3
    1. Rename and reorder a column at the same time

## Types of changes

## Checklist:


open_xdmod/modules/xdmod/component_tests/lib/ETL/DbModel/DbModelTest.php
```diff
@@ -0,0 +1,148 @@
...
+     */
+
+    public static function setupBeforeClass()
+    {
+        self::$testArtifactInputPath = realpath(BASE_DIR . '/tests/artifacts/xdmod-test-artifacts/xdmod/etlv2/configuration/input/');
+        self::$testArtifactOutputPath = realpath(BASE_DIR . '/tests/artifacts/xdmod-test-artifacts/xdmod/etlv2/configuration/output/');
+
+        // The modify_table.json file defines actions to create and modify the table as well was the test schema.
+        $configFile = self::$testArtifactInputPath . "/xdmod_etl_config_8.0.0.json";
+        self::$etlConfig = new EtlConfiguration($configFile, self::$testArtifactInputPath);
```
jtpalmer: I think you need the default module name here.  i.e.:

```
self::$etlConfig = new EtlConfiguration(
    $configFile,
    self::$testArtifactInputPath,
    $this->logger,
    array('default_module_name' => 'xdmod')
);
```

smgallo: @jtpalmer actually there's a bug in my code. It should work fine if you don't specify a default module (e.g., it will construct action names as "module.action") but I missed a check if there is no module name and removed the pipeline defaults. A fix is coming in with this PR.

PR #584, opened 2018-06-12
Guaranteed Order for Reconstructed Cloud Events

chakrabortyr:
## Description
Fixes a bug in which reconstructed events would appear out of order.

## Motivation and Context
The State Reconstructor relies on a guaranteed order to properly timestamp reconstructed cloud events. The lack of a guaranteed order will impact reporting.

## Tests performed
Yes.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project as found in the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #590, opened 2018-06-20
Enable User Defined Default Times

chakrabortyr:
Please do not merge prior to merging [#584](#584)

## Description

`etl_overseer -r RESOURCE -d end_time 'now - 7 days' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -d end_time '2018-02-26 + 1 second' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -d end_time '+ 14 days' -p cloud-state-pipeline`

## Motivation and Context
See above.

## Tests performed
Yes.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
+++ b/classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
@@ -15,10 +15,9 @@ use Log;
 class StateReconstructorTransformIngestor extends pdoIngestor implements iAction
 {
     private $stop_event_ids;
-
     private $start_event_ids;
-
     private $instance_state;
+    private $end_time;
 
     /**
      * @see ETL\Ingestor\pdoIngestor::__construct()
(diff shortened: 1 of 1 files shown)


classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -27,20 +26,23 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
...
+        $this->stop_event_ids = array(4, 6, 17, 19);
         $this->start_event_ids = array(2, 8, 9, 10, 11, 16);
         $this->all_event_ids = array_merge($this->start_event_ids, $this->stop_event_ids);
+        $this->end_time = $etlConfig->getVariableStore()->end_time;
 
         $this->resetInstance();
     }
 
     private function initInstance($srcRecord) {
+        $default_end_time = isset($this->end_time) ? strtotime($this->end_time) : $srcRecord['event_time_utc'];
```
jpwhite4: Recommend that you call the date(strtotime()) functions once in the __construct() function rather than here.

classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -27,20 +26,23 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
     {
         parent::__construct($options, $etlConfig, $logger);
 
-        $this->stop_event_ids = array(4, 6);
+        $this->stop_event_ids = array(4, 6, 17, 19);
```
plessbd: What do these numbers mean?  Is the only way for me to find their definition by looking through some other code or the database to find what they are?
Would be nice to use enums or constants rather than just numbers.

jpwhite4: A very pertinent question. In fact one could ask what does any of this mean. Perhaps a starting point would be this [introductory text](https://books.google.com/books/about/What_Does_It_All_Mean.html?id=Nz_XMrjBQEUC&printsec=frontcover&source=kp_read_button#v=onepage&q&f=false)

chakrabortyr:
I'm beginning to think we need some sort of childproofing on this repository.

plessbd:
Why so you cant be allowed to break it? :)

PR #592, opened 2018-06-21
Rename NewRest -> Rest

ryanrath:
## Description
Just renaming the NewRest namespace to Rest. Some additional cleanup was performed on some of the files that were moved so that the style checks would pass. 

NOTE: There will be PR's in both appkernels and xsede that will need to be merged when this goes in as well.

## Motivation and Context
No reason to have `NewRest` anymore as we've already migrated everything in the old `Rest`. 

## Tests performed
All unit, component, integration and regression tests were run and passed.

## Types of changes

- [X] Code Cleanup

## Checklist:

### DIFF
--- a/classes/Rest/Controllers/AuthenticationControllerProvider.php
+++ b/classes/Rest/Controllers/AuthenticationControllerProvider.php
@@ -9,7 +9,7 @@ use Rest\Utilities\Authentication;
 
 /**
  * Class AuthenticationControllerProvider
- * alskfjsalkj
+ *
  * This class is responsible for maintaining the authentication routes for the
  * REST stack.
  *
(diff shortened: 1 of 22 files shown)


classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -1,15 +1,15 @@
...
 use Silex\Application;
 use Symfony\Component\HttpFoundation\Request;
 
-use NewRest\Utilities\Authentication;
+use Rest\Utilities\Authentication;
 
 /**
  * Class AuthenticationControllerProvider
- *
+ * alskfjsalkj
```
jtpalmer: This looks like it shouldn't have been added.

ryanrath: lol this most definitely should not have been added. We were testing the style change portion of the travis script. Thanks! I'll get this removed.

ryanrath: Hi @jtpalmer That should be all set now. I also added you to the related PR's. Thanks!

PR #595, opened 2018-06-22
Multiple Test Updates

plessbd:
This updates the docker to a new build (7.5.1-v4)
This build contains:
* newer build of xdmod 7.5.1
* preinstalled php-mbstring (faster builds)
* updated chrome v67
* automated test dependency updates (multiple)

An alternative to the option changes in shippable.xml is `'--disable-dev-shm-usage'` in wdio.conf.js but according to https://github.com/SeleniumHQ/docker-selenium/issues/636#issuecomment-389109831
```
it is really needed to share the shm with the container, otherwise Chrome/Firefox won't run properly.
```


smgallo:
If we want to do faster builds, we can create our own docker with all of the dependencies installed including composer and phantomjs. See https://tas-tools-int-01.ccr.xdmod.org/git/UBCCR/ccr-private-xdmod/src/master/docker/xdmod/Dockerfile.centos7:xdmod-base and https://tas-tools-int-01.ccr.xdmod.org/git/UBCCR/ccr-private-xdmod/src/master/docker/xdmod/Dockerfile.xdmod-centos7:open7.5.1-replication

plessbd:
@smgallo get back to vacation
@smgallo this one was specifically just added php-mbstring.
I should have said specifically shippable builds, not docker builds.

It was also a test so that we could use current versions of chrome instead of having to have a specific one possibly eliminating the need to actually need to have our own docker images at all for shippable.  @jpwhite4 also found in that if we do something like https://github.com/ubccr/supremm/blob/master/shippable.yml#L4 that it seems shippable caches the docker build and that runs faster rather than it re-downloading the image every time.  So we are going to look into that as well.

Small steps ;)

open_xdmod/modules/xdmod/automated_tests/test/specs/xdmod/metricExplorer.page.js
```diff
@@ -4,6 +4,7 @@ class MetricExplorer {
         this.originalTitle = '(untitled query 2)';
         this.newTitle = '<em>"& (untitled query) 2 &"</em>';
         this.possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+        this.filterPI = 'Alpine';
```
jpwhite4: This belongs in the test file rather than the page class.

PR #602, opened 2018-06-29
Specify Custom End Dates

chakrabortyr:
This is basically the same pull request that Ben did weird stuff to #590. It also addresses some feedback from #584, which needs to be merged prior to merging this one.

## Description

`etl_overseer -r RESOURCE -d end_time 'now - 7 days' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -d end_time '2018-02-26 + 1 second' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -d end_time '+ 14 days' -p cloud-state-pipeline`

## Motivation and Context
See above.

## Tests performed
Yes.

## Types of changes

## Checklist:


classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -12,13 +22,24 @@
 
 use Log;
 
-class StateReconstructorTransformIngestor extends pdoIngestor implements iAction
+abstract class EventState
```
smgallo: This is confusing. "EventState" is ambiguous - what event? Why create a class containing constants that is used in 1 other class? These should be embeded directly in `StateReconstructorTransformIngestor`. Also, why not name it `CloudStateReconstructorTransformIngestor` as that is more descriptive?

classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -27,42 +48,50 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
...
+        $default_end_time = isset($this->_end_time) ? strtotime($this->_end_time) : $srcRecord['event_time_utc'];
+
+        $this->_instance_state = array(
             'instance_id' => $srcRecord['instance_id'],
             'start_time' => $srcRecord['event_time_utc'],
             'start_event_id' => $srcRecord['event_type_id'],
-            'end_time' => $srcRecord['event_time_utc'],
-            'end_event_id' => $srcRecord['event_type_id']
+            '_end_time' => date('Y-m-d H:i:s', $default_end_time),
+            'end_event_id' => 4
```
smgallo: What does `4` represet? Is this the `STOP` event?

classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -27,42 +48,50 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
...
+    {
+        $default_end_time = isset($this->_end_time) ? strtotime($this->_end_time) : $srcRecord['event_time_utc'];
+
+        $this->_instance_state = array(
             'instance_id' => $srcRecord['instance_id'],
             'start_time' => $srcRecord['event_time_utc'],
             'start_event_id' => $srcRecord['event_type_id'],
-            'end_time' => $srcRecord['event_time_utc'],
-            'end_event_id' => $srcRecord['event_type_id']
+            '_end_time' => date('Y-m-d H:i:s', $default_end_time),
```
smgallo: Why is there an underscore prefix here?

classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
chakrabortyr: Yes, I forgot to change that.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,132 @@
...
+    private $_end_time;
+
+    /**
+     * @see ETL\Ingestor\pdoIngestor::__construct()
+     */
+    public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log $logger = null)
+    {
+        parent::__construct($options, $etlConfig, $logger);
+
+        $this->_stop_event_ids = array(STOP, TERMINATE, SUSPEND, SHELVE);
```
smgallo: These should be referenced as `self::STOP` and so on.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,132 @@
...
+
+        return $transformedRecord;
+    }
+
+    protected function getSourceQueryString()
+    {
+        $sql = parent::getSourceQueryString();
+
+        // We add a dummy row here to address gimmick wherein ETL discards the last row of data.
+        $unionValues = [];
```
smgallo: Lines 120 - 125 can be condensed into: `$unionValues = array_fill(0, $colCount, 0);`

smgallo:
@chakrabortyr Lets see if we can get this wrapped up today.

plessbd:
What is supposed to take precedence in the case of `getSourceQueryString`?  It looks like this one uses the previous sql when compared to #584.  We need to resolve this and close #584 if this is going to be doing the same work.

#602
```php
$sql = "SELECT * FROM ( $sql \nORDER BY instance_id ASC, event_time_utc ASC) a\nUNION ALL\nSELECT " .
            implode(',', $unionValues);
```
vs

#584
```php
$sql = "$sql \nUNION ALL\nSELECT " . implode(',', $unionValues) . "\nORDER BY 1 DESC, 2 ASC";
```

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        return $transformedRecord;
+    }
+
+    protected function getSourceQueryString()
+    {
+        $sql = parent::getSourceQueryString();
+
+        // We add a dummy row here to address gimmick wherein ETL discards the last row of data.
```
smgallo: Lets add a better explanation of what's actually going on.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -116,7 +116,8 @@ protected function getSourceQueryString()
     {
         $sql = parent::getSourceQueryString();
 
-        // We add a dummy row here to address gimmick wherein ETL discards the last row of data.
+        // The ETL process has a glitch where the last row of data is discarded at ingestion time. To work around
```
smgallo: Why is the last line discarded? This is not a general ETL issue, is it?

plessbd: Etl does not discard. This is to known when you are at the end since the etl doesn't have the ability to tell you. So this dummy record is placed there so that the state machine knows to flush it's state

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,126 @@
...
+        $this->_instance_state['end_time'] = $srcRecord['event_time_utc'];
+        $this->_instance_state['end_event_id'] = $srcRecord['event_type_id'];
+    }
+
+    /**
+     * @see ETL\Ingestor\pdoIngestor::transform()
+     */
+    protected function transform(array $srcRecord, $orderId)
+    {
+        if (!in_array($srcRecord['event_type_id'], $this->_all_event_ids)) {
```
jpwhite4: You need to flush the stored state as the first thing in this function.

The current code adds the extra row of zeros but the zero is not in the `$this->_all_event_ids` array so the stored state will not get flushed

open_xdmod/modules/xdmod/component_tests/lib/ETL/CloudStateReconstructorTransformIngestorTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+    private $fsm;
+
+    public function __construct() {
+        $this->fsm = new CloudStateReconstructorTransformIngestor();
+    }
+
+    protected function testValidTransformation() {
+        $this->fsm::transform($this->valid_event, 1);
+        $event = $this->fsm::transform($this->valid_end_event, 1);
```
smgallo: What is this syntax attempting to call a protected method of a class from outside of the class?

smgallo: What is this syntax attempting to call a protected method of a class from outside of the class?

open_xdmod/modules/xdmod/component_tests/lib/ETL/CloudStateReconstructorTransformIngestorTest.php
```diff
@@ -0,0 +1,81 @@
...
+    );
+
+    private $fsm;
+
+    public function __construct() {
+        $this->fsm = new CloudStateReconstructorTransformIngestor();
+    }
+
+    protected function testValidTransformation() {
+        $this->fsm::transform($this->valid_event, 1);
```
smgallo: This statement does nothing, did you forget to check the return value?

open_xdmod/modules/xdmod/component_tests/lib/ETL/CloudStateReconstructorTransformIngestorTest.php
```diff
@@ -0,0 +1,81 @@
...
+        
+        $this->assertEquals($this->valid_transform, $event);
+    }
+
+    protected function testInvalidTransformation() {
+        $this->assertEquals(array(), $this->fsm::transform($this->invalid_event, 1));
+    }
+
+    protected function testZeroTransformation() {
+        $this->assertEquals(null, $this->fsm::transform($this->zero_event, 1));
```
smgallo: I do not see where `CloudStateReconstructorTransformIngestor::transform()` can return a `null` value (nor should it).

chakrabortyr:
Closing this till I get it working so you aren't spammed with my changes

PR #607, opened 2018-07-06
Filter out Open Stack heartbeat events with stopped or deleted state

eiffel777:
In Open Stack there are heartbeat events that occur after the VM has been shut down. These heartbeat events have a state that is not active and need to be filtered out. If not filtered out then a second row for an instance is created in the cloud_event_transient table which causes incorrect data when aggregation is run. 

## Motivation and Context
Heartbeat events are events that let us know that a VM is still active. It is expected that heartbeat events only exist when a VM is running. This appears to not be the case for Open Stack. In Open Stack heartbeat events for instances are named compute.instance.exists and they all have a state property. 

When an instance is running the state property equals active. After an instance is shutdown one more heartbeat event for that instance occurs with a state of stopped or deleted. This extra heartbeat event after the instance ends causes the StateReconstructorTransormIngestor to add a second row for the instance to the cloud_event_transient table. 

When aggregation is run these extra rows for the instance cause some statistics, such as Number of VM's ended or Cores: Total, to count the details for these instances twice, leading to incorrect information.

## Tests performed
Tested in local docker

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_openstack/raw_event.json
+++ b/configuration/etl/etl_tables.d/cloud_openstack/raw_event.json
@@ -124,7 +124,7 @@
             		"type": "varchar(64)",
             		"nullable": true,
             		"default": null
-      	    },
+            },
             {
                 "name": "state",
                 "type": "varchar(64)",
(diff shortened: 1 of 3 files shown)


configuration/etl/etl_tables.d/cloud_openstack/raw_event.json
```diff
@@ -124,7 +124,13 @@
             		"type": "varchar(64)",
             		"nullable": true,
             		"default": null
-      	    }
+      	    },
+            {
```
chakrabortyr: Nitpicky, but let's keep the formatting consistent.

PR #608, opened 2018-07-09
Update User Organization on Login

ryanrath:
**NOTE: This PR depends on https://github.com/ubccr/xdmod/pull/572 ( the latest changes from this PR have been pulled into this one so that we could get things working ).**

**We now return you to your regularly scheduled PR description...**

## Description
These changes add the execution of `syncUserOrganization` to the standard login
process ( Or will, when the supporting PR's have been merged & this code
uncommented. ). The process goes something like this:
- The user's current organization ( `moddb.Users.organization_id` ) is different
than `Organizations::getOrganizationForUser`
 - Check if the user had elevated privileges for their previous organization.
    i.e. Did they have Center Director / Center Staff
    - If they did then make sure to revoke those privileges
  - Update the User's organization to the result of
  `Organizations::getOrganizationForUser`
  - Notify the admins that a user organization update was detected / performed.
    Include relevant User information.
  - Finally notify the user of the changes that were performed and how they can
    contact us if they have any questions or concerns.

## Motivation and Context
These changes are being introduced so that we have a way to keep the `moddb.Users.organization_id` up to date.

## Tests performed
Manual tests will be performed & automatic tests added where appropriate.

## Types of changes

## Checklist:


classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -74,4 +137,78 @@ public function logout(Request $request, Application $app)
...
+     */
+    private function syncUserOrganization(XDUser $user)
+    {
+        // =========================================================================================
+        // NOTE: The function body is commented until the classes / functions in other PR's
+        //       have been merged.
+        // =========================================================================================
+
+        /*
+        $userOrganization = $user->getOrganizationID();
```
smgallo: Lets change the variable names to make it clearer what we're looking at. Is value returned by `getOrganizationForUser()` the user's current organization and the value in XDUser what we have stored in XDMoD? If so maybe something like this? Not sure of the best names to use.
```
$userProfileOrganization = $user->getOrganizationID();
$dataWarehouseOrganization = Organizations::getOrganizationForUser($user->getUserID());
```

classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -74,4 +137,78 @@ public function logout(Request $request, Application $app)
...
+                        $user->getUsername(),
+                        $user->getEmailAddress(),
+                        Organizations::getNameById($userOrganization),
+                        Organizations::getNameById($currentOrganization),
+                        json_encode($originalAcls),
+                        json_encode($otherAcls)
+                    )
+                );
+            } else {
+                $this->emailLogger->notice(
```
smgallo: We certainly want an email if a user with elevated privs changed organizations and had those privs removed, but do we care if their organization simply changed? Don't want to get bombarded by emails.

classes/Models/Services/Organizations.php
```diff
@@ -29,22 +29,22 @@ public static function getOrganizationForUser($userId)
       FROM (
       SELECT DISTINCT
           u.organization_id,
-          2
+          3
```
smgallo: I think that we need an English description of how we determine the organization included in this function as a comment.

classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -167,11 +164,20 @@ private function syncUserOrganization(XDUser $user)
                 $originalAcls = $user->getAcls(true);
                 $otherAcls = array_diff($originalAcls, self::$CENTER_ROLES);
 
-                $user->setAcls($otherAcls);
+                // Make sure that they at least have 'usr'
+                if (empty($otherAcls)) {
```
smgallo: Do we require that all XDMoD users have the `usr` acl? What about program officers or someone with only the financial role? How will this affect existing users since now the `usr` role will automatically be added at login?

classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -196,21 +202,23 @@ private function syncUserOrganization(XDUser $user)
...
+                            \xd_utilities\getConfiguration('mailer', 'sender_email')
+                        ),
+                        'toAddress' => $user->getEmailAddress(),
+                        'fromAddress' => \xd_utilities\getConfiguration('general', 'tech_support_recipient'),
+                        'fromName' => '',
+                        'replyAddress' => \xd_utilities\getConfiguration('mailer', 'sender_email')
+                    )
+                );
+            } catch (\Exception $e) {
+                // come up with something to do if we can't send an email to the user.
```
smgallo: Log to the console/file?

classes/Rest/Controllers/AuthenticationControllerProvider.php
```diff
@@ -74,4 +139,86 @@ public function logout(Request $request, Application $app)
...
+     * @throws \Exception if there is a problem updating the provided users organization.
+     */
+    private function syncUserOrganization(XDUser $user)
+    {
+
+        $userOrganization = $user->getOrganizationID();
+        $currentOrganization = Organizations::getOrganizationForUser($user->getUserID());
+
+        if ($userOrganization !== $currentOrganization) {
+            $userOrganizationName = Organizations::getNameById($userOrganization);
```
jpwhite4: `getNameById()` may return null so you will need to handle that case. It will likely be less work to change the `getNameById()` to return a string such as "UNKNOWN" rather than `null`.

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,92 @@
...
+     * @throws \Exception if there is a problem executing sql statements.
+     */
+    public static function getOrganizationForUser($userId)
+    {
+        $sql = <<<SQL
+    SELECT src.organization_id
+    FROM (
+      SELECT user_org.*
+      FROM (
+      SELECT DISTINCT
```
jpwhite4: Why is the DISTINCT keyword used here? Under what circumstances will there be multiple ids returned from this query?

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,92 @@
...
+          u.organization_id,
+          3
+        FROM moddb.Users u
+        JOIN modw.organization o ON o.id = u.organization_id
+        WHERE u.id = :user_id
+        UNION
+        SELECT DISTINCT
```
jpwhite4: Why is the DISTINCT keyword used here? Under what circumstances will there be multiple ids returned from this query?

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,92 @@
...
+          o.id organization_id,
+          1
+        FROM moddb.Users u
+        JOIN modw.person p ON p.id = u.person_id
+        JOIN modw.organization o ON o.id = p.organization_id
+        WHERE u.id = :user_id
+        UNION
+        SELECT DISTINCT
```
jpwhite4: Why is the DISTINCT keyword used here? Under what circumstances will there be multiple ids returned from this query?

classes/Models/Services/Parameters.php
```diff
@@ -0,0 +1,56 @@
...
+
+        // We need to retrieve which dimensions this acl filters on. To do that we need to see how
+        // it's configured
+        $config = Config::factory();
+
+        // retrieve the roles section of the roles.json/.d config files.
+        $roles = $config['roles']['roles'];
+
+        try {
+            $aclConfig = $dimensions = $roles[$aclName];
```
jpwhite4: please check this line. Did you really intend to have two assignments here?

classes/Models/Services/Parameters.php
```diff
@@ -0,0 +1,56 @@
...
+            $dimensions = isset($aclConfig['dimensions']) ? $aclConfig['dimensions'] : array();
+        } catch (\Exception $e) {
```
jpwhite4: What circumstances will result in a general Exception  being thrown here?

classes/Models/Services/Organizations.php
ryanrath: No good reason, if I had to guess I copied / pasted the `user_acl_group_by_parametesr` query as the starting point for the others.

classes/Models/Services/Organizations.php
ryanrath: No good reason, if I had to guess I copied / pasted the `user_acl_group_by_parametesr` query as the starting point for the others.

classes/Rest/Controllers/AuthenticationControllerProvider.php
ryanrath: Yup sounds good. I'll commit that shortly

classes/Rest/Controllers/AuthenticationControllerProvider.php
ryanrath: *nod* removed

classes/Models/Services/Organizations.php
ryanrath: The short answer is: when a user is associated with multiple organizations. 

When `replace_role_parameters` is merged then this sub-query can be removed as a users organization will either `moddb.Users.organization_id` or `modw.Person.organization_id` ( via `moddb.Users.person_id` ).

classes/Rest/Controllers/AuthenticationControllerProvider.php
ryanrath: *nod* Will be added shortly.

classes/Models/Services/Parameters.php
ryanrath: If there is no configuration section for the provided `$aclName`.

classes/Models/Services/Parameters.php
ryanrath: I did, but it's easily re-writable: 
```php
$dimensions = isset($roles[$aclName]['dimensions']) ? $roles[$aclName]['dimensions'] : array();
```

classes/Rest/Controllers/AuthenticationControllerProvider.php
ryanrath: It's not required per se, but it is the base level of access for authenticated users. It will also only be added as a fall back to prevent a user from having no acls.

classes/Models/Services/Organizations.php
ryanrath: Definitely, I've added that in the latest commit. Let me know what you think.

classes/Models/Services/Parameters.php
ryanrath: This has been updated in the latest commits.

classes/Models/Services/Organizations.php
ryanrath: Granted, a user will not be able to be associated with multiple organizations moving forward, but that's why the `DISTINCT` is there.

classes/Models/Services/Parameters.php
```diff
@@ -28,8 +28,7 @@ public static function getParameters(\XDUser $user, $aclName)
         $roles = $config['roles']['roles'];
 
         try {
-            $aclConfig = $dimensions = $roles[$aclName];
-            $dimensions = isset($aclConfig['dimensions']) ? $aclConfig['dimensions'] : array();
+            $dimensions = isset($roles[$aclName]['dimensions']) ? $roles[$aclName]['dimensions'] : array();
```
smgallo: This statement will not throw an exception. if `$aclName` is not defined or `$roles[$aclName]['dimensions']` is not defined (or is `NULL`), `$dimensions = array()`. If you want to thrown an exception if `$aclName` is not defined do an explicit check:
```
if ( ! isset($roles[$aclName]) ) {
    throw new \Exception(...);
}
```

smgallo:
Apparently github doesn't like that I commented on a file that was changed... In `classes/Models/Services/Parameters.php` line 31:

This statement will not throw an exception. if `$aclName` is not defined or `$roles[$aclName]['dimensions']` is not defined (or is `NULL`), `$dimensions = array()`. If you want to thrown an exception if `$aclName` is not defined do an explicit check:
```
if ( ! isset($roles[$aclName]) ) {
    throw new \Exception(...);
}
```

ryanrath:
lol yah.... For some reason I thought PHP threw an exception if it tried to access an index that doesn't exist. On the plus side, less code is always good. Thanks!

smgallo:
You should still do the check in case `$aclName` is not defined though.

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,102 @@
...
+
+        return count($rows) > 0 ? $rows[0]['organization_id'] : UNKNOWN_ORGANIZATION_ID;
+    }
+
+    /**
+     * Retrieve the name for the organization identified by the provided id.
+     *
+     * @param int $organizationId the id of the organization whose name is to be retrieved.
+     *
+     * @return string|null Returns null if the organization could not be found, else the `name`
```
jpwhite4: Please update this comment

classes/Models/Services/Organizations.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $query = "SELECT o.name FROM modw.organization o WHERE o.id = :organization_id";
+        $params = array(
+            ':organization_id' => $organizationId
+        );
+
+        $db = DB::factory('database');
+
+        $rows = $db->query($query, $params);
+        return count($rows) > 0 ? $rows[0]['name'] : UNKNOWN_ORGANIZATION_NAME;
```
jpwhite4: Need to fix this line

ryanrath: D'oh, thanks! just pushed the fix.

configuration/etl/etl_data.d/jobs/xdw/organizations.json
```diff
@@ -0,0 +1,4 @@
+[
+  ["id", "abbrev", "name", "short_name", "long_name", "organization_origin_id"],
+  [-1, "UNK", "Unknown Organization", "Unknown", "UNK - Unknown Organization", -1]
```
jpwhite4: Where is the abbrev value used/displayed? I suggest changing this to UNKNOWN since UNK is entirely not obvious.

ryanrath: *nod* I was basically just following the format of our test data. But this could definitely be updated to be more intuitive.

ryanrath: does this work? 
```json
[-1, "unknown", "Unknown Organization", "Unknown", "unknown - Unknown Organization", -1]
```
again, just following the basic format of our test organization.

ryanrath: It looks like it's used in a couple of places: 
- Internal Dashboard User Admin areas
- Center Director -> Role Delegation 
- ... ?

smgallo: Lets capitalize "unknown" for consistency. Otherwise I think that's good.

ryanrath: Sounds good. Capitalized change in / building now

PR #626, opened 2018-08-03
Prompt for resource type in xdmod-setup

smgallo:
## Description

`xdmod-setup` add resource menu hard-codes a resource type of 1 (HPC). Prompt the user for the resource type, defaulting to HPC. Allowable types are currently HPC and Cloud. Use the `etc/resource_types.json` file to map from the string to `resouce_type_id`.

## Motivation and Context

Needed by @chakrabortyr and @eiffel777 for adding cloud resources during testing.

## Tests performed

Manual and shippable.

## Types of changes

## Checklist:

### DIFF
--- a/classes/OpenXdmod/Setup/AddResourceSetup.php
+++ b/classes/OpenXdmod/Setup/AddResourceSetup.php
@@ -33,6 +33,26 @@ class AddResourceSetup extends SetupItem
      */
     public function handle()
     {
+        $typeOptions = array();
+        $typeDescriptionText = "";
+        $availableTypes = json_decode(file_get_contents(CONFIG_DIR . '/resource_types.json'), true);
+
+        usort(
+            $availableTypes,
+            function ($a, $b) {
+                return strcmp($a['abbrev'], $b['abbrev']);
+            }
+        );
+
+        foreach ( $availableTypes as $type ) {
+            if ( 'UNK' == $type['abbrev'] ) {
+                continue;
+            }
+            // Note that Console::prompt() expects lowercase values for options
+            $typeOptions[] = strtolower($type['abbrev']);
+            $typeDescriptionText .= sprintf("%-10s - %s", $type['abbrev'], $type['description']) . PHP_EOL;
+        }
+
         $this->console->displaySectionHeader('Add A New Resource');
 
         $this->console->displayMessage(<<<"EOT"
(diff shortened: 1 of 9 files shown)


classes/OpenXdmod/Setup/AddResourceSetup.php
```diff
@@ -46,6 +46,7 @@ public function handle()
 
         $resource = $this->console->prompt('Resource Name:');
         $name     = $this->console->prompt('Formal Name:');
+        $type     = $this->console->prompt('Resource Type:', 'hpc', array('cloud', 'hpc'));
```
jtpalmer: There are many more options in [resource_types.json](https://github.com/ubccr/xdmod/blob/xdmod8.0/configuration/resource_types.json).  Why not list them all?

smgallo: I considered that, but we only have support for HPC and Cloud (and soon Storage) resources at the moment so I did not want to present an opportunity to confuse users during the install until we actually did something with it. If folks think it would be beneficial to list all of the abbreviations it's easy enough to do.

chakrabortyr: I'd say we err on the side of not adding additional sources of ambiguity.

jtpalmer: I plan on add something for storage and there are three resource types that qualify for that: Tape storage resource, Disk storage resource and Storage grid.

Is that going to be an issue?

I'd preferable just add them all since the data source is an ostensibly user-editable file.

Otherwise, we may as well not allow users to edit the resource types and not use all the XSEDE resource types.

smgallo: Fair enough, I will make that change.

jtpalmer: If @chakrabortyr doesn't think we should add them all, then maybe we shouldn't.  I don't want to create more work if it's just going to confuse the end-users.  We haven't really done anything with the resource types in the past, they're just displayed as a dimension so I want to be prepared if we're going to start giving them additional meaning and not just using them as display labels.

smgallo:
Made changes to display all resource types along with their descriptions based on discussion with @jtpalmer and @chakrabortyr.

PR #627, opened 2018-08-03
Update user organization admin

ryanrath:
## Description
Updating the code responsible for associating an XDMoD user w/ an Organization.

## Motivation and Context
We have updated the way in which we expect users to be associated to an organization. Previously a user could be associated with multiple organizations. But we found, upon inspection of our data that the only users who were using this functionality were Developers.

In an effort to simplify our code base / ways in which a user interacts with our system we have moved to requiring a user be associated with one and only one organization. This is to be accomplished by the user being associated with a 'Person'. This person's associated organization is then used to populate the Users organization.

## Tests performed
Testing Platforms: Open XDMoD && XSEDE
- Unit Tests
- Component Tests
- Integration Tests
- UI Tests
- Manual Tests ( per testing platform ): 
    - Login to the Internal Dashboard and ensure that all of the existing Users have had their organization set appropriately.
        - when testing on XSEDE the `acl-import` script must be run prior to testing.
  - Login with each existing user, ensure that the quick filters are displaying / populated as expected.

## Types of changes

## Checklist:


classes/XDAdmin.php
```diff
@@ -172,11 +172,8 @@ public function enumerateResourceProviders()
      *
      * @return array
      */
-    public function enumerateInstitutions($start, $limit, $nameFilter = NULL)
+    public function enumerateInstitutions($start, $limit = null, $nameFilter = NULL)
```
jpwhite4: It looks like the `$start` and `$limit` parameters are not used in this function now. Why did this change?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -695,14 +704,22 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
             listeners: {
                 activate: function () {
                     newUserRoleGrid.reset();
+                    cmbInstitution.getStore().load(
+                        {
+                            params: {
+                                start: 0,
+                                limit: 5000
```
jpwhite4: Why 5000?

classes/XDAdmin.php
ryanrath: This changed due to running into issues with setting organizations for users on form load ( by id ). Which was caused by not having all of the organizations retrieved when setting the components value.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: This should have been removed actually. It was part of the testing process to figure out why, when retrieving / setting an organization for a user, it would sometimes display nothing ( this is also related to the change you commented on above ).

html/controllers/user_admin/enum_institutions.php
```diff
@@ -1,39 +1,28 @@
...
+
+$xda = new XDAdmin();
+
+$name_filter = (isset($_POST['query'])) ? $_POST['query'] : NULL;
+
+list($institutionCount, $institutions) = $xda->enumerateInstitutions($name_filter);
+
+$institutionEntries = array();
+
+foreach ($institutions as $institution) {
```
smgallo: This appears to be making a copy of the array (with sequential numeric keys, but that's likely what's returned by the query anyway). Why not use `$returnData['institutions'] = $institutions;`?

html/controllers/user_admin/update_user.php
```diff
@@ -155,6 +156,9 @@ function ($carry, Acl $item) {
 
 // -----------------------------
 
+$user_to_update->setOrganizationID(Organizations::getOrganizationForUser($user_to_update->getUserID()));
```
smgallo: Do we need any error checking here or will things be handled properly if `getUserId()` or `getOrganizationForUser()` fail?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
smgallo: Should this still be removed?

classes/XDAdmin.php
```diff
@@ -186,13 +181,9 @@ public function enumerateInstitutions($start, $limit, $nameFilter = NULL)
             "SELECT COUNT(*) AS total_records FROM organization $filter"
```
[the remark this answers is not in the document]
ryanrath: Just the way it was written, but we can certainly just count the results.

classes/XDWarehouse.php
```diff
@@ -195,15 +195,26 @@ public function enumerateGridUsers(
      */
     public function resolveInstitutionName($institution_id)
     {
-        $instQuery = $this->_pdo->query(
-            "SELECT name FROM organization WHERE id = :id",
-            array('id' => $institution_id)
+        $query = <<<SQL
+SELECT name
```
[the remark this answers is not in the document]
ryanrath: Yup, just need to add an `ORDER BY id DESC` to make sure that if the id is found it appears before the unknown organization

html/controllers/user_admin/enum_institutions.php
ryanrath: no real reason, it's just how it was originally written. I'll update it.

classes/XDAdmin.php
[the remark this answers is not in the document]
smgallo: Might as well clean it up while we're in there.

html/controllers/user_admin/update_user.php
ryanrath: Any exceptions thrown on this line will handled by the Exception block on line 201.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: Yup, we're just returning all of the organizations now. I bumped the limit up during testing to see if the problems I was encountering were caused by the organization not being present ( due to it not being within the `limit` ) when attempting to set the value for this control.

classes/Models/Services/Organizations.php
```diff
@@ -88,4 +88,32 @@ public static function getNameById($organizationId)
...
+     */
+    public static function getOrganizations()
+    {
+        $db = DB::factory('database');
+        return $db->query(
+            "SELECT o.* FROM modw.organization o;"
+        );
+    }
+
+    public static function getOrganizationIdForPerson($personId)
```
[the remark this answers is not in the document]
ryanrath: kk

PR #634, opened 2018-08-14
Specify Custom End Dates + Union Bugfixes

chakrabortyr:
This is #602 over again, wanted to make sure everything worked correctly so I didn't spam you guys with commits.

## Description

`etl_overseer -r RESOURCE -e 'now - 7 days' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -e '2018-02-26 + 1 second' -p cloud-state-pipeline`
`etl_overseer -r RESOURCE -e '+ 14 days' -p cloud-state-pipeline`

## Motivation and Context
See above.

## Tests performed
Manual verification of data consistency, added component test to validate event reconstruction.

## Types of changes

## Checklist:


smgallo:
The ETL overseer already has an `-e | --end-date` option why not use that?

open_xdmod/modules/xdmod/component_tests/lib/ETL/CloudStateReconstructorTransformIngestorTest.php
```diff
@@ -0,0 +1,103 @@
...
+    );
+
+    private $fsm;
+
+    public function __construct()
+    {
+        $configFile = realpath(BASE_DIR . '/tests/artifacts/xdmod-test-artifacts/xdmod/etlv2/configuration/input/xdmod_etl_config_8.0.0.json');
+
+        // This needs to be explicitly defined here so PHP 5.4 doesn't complain
+        $definitionLoc = BASE_DIR . "/tests/artifacts/xdmod-test-artifacts/xdmod/etlv2/configuration/input/etl_action_defs_8.0.0.d/cloud_state.json";
```
smgallo: Rather than create a single use variable combine these 2 lines:
```
$this->options_array["definition_file"] =
    BASE_DIR .
    "/tests/artifacts/xdmod-test-artifacts/xdmod/etlv2/configuration/input/etl_action_defs_8.0.0.d/cloud_state.json";
```

classes/CCR/Loggable.php
```diff
@@ -129,7 +129,10 @@ public function logAndThrowException($message, array $options = null)
 
         $logMessage['message'] = $message;
 
-        $this->logger->log($logMessage, $logLevel);
+        if ($this->logger) {
```
smgallo: Why is this needed? If a logger has not been passed to the constructor it is set to `Logger::singleton('null')` in `setLogger()`.

chakrabortyr: I added this after manually running tests and seeing the Logger barf when the null Logger was used specifically in that function. Adding the conditional seemed to fix the issue, so I left it there as an added safety check.

smgallo: I have not run into the issue. Have you looked at my working tests? I'd like to remove the conditional unless it is needed.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @see ETL\Ingestor\pdoIngestor::transform()
+     */
+    protected function transform(array $srcRecord, &$orderId)
+    {
+        // We want to just flush when we hit the dummy row
+        if ($srcRecord['event_type_id'] === 0) {
+            return array($this->_instance_state);
```
jpwhite4: need an
```php
if ($this->instance_state !== null) {
   return array($this->_instance_state);
} else {
   return array();
}
```

open_xdmod/modules/xdmod/component_tests/phpunit.xml.dist
```diff
@@ -6,8 +6,8 @@
          cacheTokens="true"
          colors="true"
          convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
+         convertNoticesToExceptions="false"
```
jpwhite4: Why did this change?

classes/CCR/Loggable.php
```diff
@@ -130,6 +130,7 @@ public function logAndThrowException($message, array $options = null)
         $logMessage['message'] = $message;
 
         $this->logger->log($logMessage, $logLevel);
+
```
jpwhite4: don't make changes that just add whitespace to files that would otherwise be not modified.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,140 @@
...
+     * @see ETL\Ingestor\pdoIngestor::__construct()
+     */
+    public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log $logger = null)
+    {
+        parent::__construct($options, $etlConfig, $logger);
+
+        $this->_stop_event_ids = array(self::STOP, self::TERMINATE, self::SUSPEND, self::SHELVE);
+        $this->_start_event_ids = array(self::START, self::RESUME, self::STATE_REPORT, self::UNSHELVE);
+        $this->_all_event_ids = array_merge($this->_start_event_ids, $this->_stop_event_ids);
+        $this->_end_time = $etlConfig->getVariableStore()->endDate;
```
jpwhite4: You should do the 
```php
date('Y-m-d H:i:s', strtotime($etlConfig->getVariableStore()->endDate))
```
call here so that it only has to happen once and not every time there is a new instance.

open_xdmod/modules/xdmod/component_tests/phpunit.xml.dist
```diff
@@ -6,8 +6,8 @@
...
+         convertWarningsToExceptions="false"
```
jpwhite4: Please revert these changes.

PR #635, opened 2018-08-15
Add cloud setup

eiffel777:
Add functionality to move the templates/roles.d/cloud.json file to configuration/roles.d to and run appropriate acl related commands to enable to cloud realm

## Description
Not all groups that install XDMod have cloud resources and therefore we do not want to have the acl's for the cloud realm enabled until a cloud resource is added. To do this we place a configuration file titled cloud.json in the templates/roles.d/ folder. When a cloud resource is added for the first time this file is moved to configuration/roles.d/ and the appropriate acl commands are run to enable the cloud realm. 

When any subsequent cloud resource is added these files are moved and acl commands are run only if the templates/roles.d/cloud.json is different from configuration/roles.d/cloud.json

## Tests performed
Tested in local docker

## Types of changes

## Checklist:


classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -145,11 +145,75 @@ public function addResource(array $resource)
...
+     * Save the current list of resources. If a cloud resource exists in the list of resources
+     * add the cloud acl files and run acl config commands
      */
     public function save()
     {
         $this->saveJsonConfig($this->resources,     'resources');
         $this->saveJsonConfig($this->resourceSpecs, 'resource_specs');
+
+        $cloud_resources_exist = array_filter($this->resources, function($resource){
+            return $resource['resource_type_id'] == 5;
```
smgallo: Rather than hardcoding the id here, does it make sense to look it up in the `resource_types.json` file as we do in `addResource()`?

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -145,11 +145,75 @@ public function addResource(array $resource)
...
+        }
+    }
+
+    /**
+     * Checks to see if the cloud.json files in configuration/datawarehouse.d and configuration/roles.d
+     * match the cloud.json files in templates/datawarehouse.d and templates/roles.d
+     */
+    private function doCloudAclFilesMatch()
+    {
+        $datawarehouse_config = CONFIG_DIR.'/datawarehouse.d/cloud.json';
```
smgallo: Add space between constant and dot.

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -145,11 +145,75 @@ public function addResource(array $resource)
...
+
+        return ((md5(file_get_contents($datawarehouse_config)) == $datawarehouse_config_template) || (md5(file_get_contents($roles_config)) == $roles_config_template)) ? true : false;
+    }
+
+    /**
+     * Moves cloud.json files from templates/datawarehouse.d and templates/roles.d to configuration/roles.d
+     * and configuration/datawarehouse.d and then runs acls-xdmod-management, acl-config and acl-import to
+     * enable the Cloud realm.
+     */
+    private function addCloudAcls()
```
smgallo: If we delete a cloud resource, are the cloud ACLs removed?

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -145,11 +145,75 @@ public function addResource(array $resource)
...
+    {
+        $datawarehouse_config_dir = CONFIG_DIR.'/datawarehouse.d';
+        $roles_config_dir = CONFIG_DIR.'/roles.d';
+
+        $datawarehouse_config_template_dir = TEMPLATE_DIR.'/datawarehouse.d';
```
smgallo: Add space between constant and dot.

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -145,11 +145,66 @@ public function addResource(array $resource)
...
+        }
+    }
+
+    /**
+     * Checks to see if the cloud.json files in configuration/datawarehouse.d and configuration/roles.d
+     * match the cloud.json files in templates/datawarehouse.d and templates/roles.d
+     */
+    private function doCloudAclFilesMatch()
+    {
+        $roles_config = CONFIG_DIR.'/roles.d/cloud.json';
```
smgallo: Add space between constant and dot here and elsewhere.

classes/OpenXdmod/Setup/ResourcesSetup.php
eiffel777: @smgallo Is it likely that the resource type id will change? I would think that it would be unlikely that it would change right?

smgallo: It's unlikely, but it is a user editable file. We should discuss this as a group regarding what should happen but for now leave it as it is.

PR #643, opened 2018-08-17
fix cloud ingestion and table keys

plessbd:
This is a few changes that @jpwhite4 @chakrabortyr and I worked on for the cloud realm.

I dont want these fixes to get lost in the shuffle.


plessbd:
@chakrabortyr says he is doing this in https://github.com/chakrabortyr/xdmod/tree/RC_MultiResFix and will also be adding unit tests

PR #654, opened 2018-08-28
Update xdmod setup tcl script

eiffel777:
Updating the xdmod-setup.tcl and adding xdmod-upgrade-add-cloud-resource.tcl file to programmatically create cloud resources for testing instead of manually add a resource to the database. Also adding etl actions to jobs-cloud-common.json to ingest resources from the resources.json config file.

## Description
In xdmod-setup.tcl a series of prompts were added so that a cloud resource can be added for testing the same way that regular HPC resources are added in that file. When the upgrade tests are run a file called xdmod-upgrade-add-cloud-resource.tcl was added to add a cloud resource for running the upgrade tests.

Once the resources are added in using the .tcl files they then need to be imported into the database. To do this 4 actions where added to the jobs-cloud-common.json file based off the information in #651 on which actions need to be run to import resources from the resources.json file.

## Tests performed
Tested in local docker manually and ran all component, integration and regression tests for both a fresh install and an upgrade

## Types of changes

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
+++ b/open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
@@ -28,7 +28,7 @@ then
     xdmod-import-csv -t names -i $REF_DIR/names.csv
     xdmod-ingestor
     php /root/bin/createusers.php
-    #Adding open stack resource since there is no way to automatically add a cloud resource.
+    #Updating minmaxdate table so data for cloud realm shows up
     mysql -e "UPDATE modw.minmaxdate SET max_job_date = '2018-07-01';"
     #Ingesting cloud data from references folder
     php /usr/share/xdmod/tools/etl/etl_overseer.php -p jobs-common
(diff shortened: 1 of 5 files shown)


open_xdmod/modules/xdmod/integration_tests/scripts/xdmod-upgrade-add-cloud-resource.tcl
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env expect
+# Expect script that runs xdmod-upgrade to upgrade an already installed Open
+# XDMoD instance.
+
+#-------------------------------------------------------------------------------
+
+# Helper functions
+proc provideInput { prompt response } {
```
jpwhite4: I really dislike cut and paste duplication of code. Much prefer to put common functions in a file and then source it from all the different places it is needed

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
```diff
@@ -28,23 +28,15 @@ then
     xdmod-import-csv -t names -i $REF_DIR/names.csv
     xdmod-ingestor
     php /root/bin/createusers.php
-    #Copying roles file so Cloud realm shows up
-    mkdir -p /etc/xdmod/roles.d
-    cp $BASEDIR/../../../../../templates/roles.d/cloud.json /etc/xdmod/roles.d/
     #Adding open stack resource since there is no way to automatically add a cloud resource.
```
plessbd: update the comment, since this isnt updating the resource anymore just updating the min max table since the current min max has not been updated to support multiple realms.

Same for it in upgrade path.

PR #661, opened 2018-08-30
Updating Backup Tables to not be Temporary

ryanrath:
## Description
The backup tables will now be fully fledged tables instead of temporary tables.
This will ensure that if the `acl-config` process stops before completion, but
after the backups have been performed, these tables will still be available to
reconstruct their source tables.

## Motivation and Context
Slightly more robust than the previous implementation

## Tests performed
Manual Tests were performed:

Testing Environments:
  - Open XDMoD Fresh Install
  - Open XDMoD Upgrade
  - XSEDE upgrade

Test:
  - Spin up docker && install / upgrade as appropriate
  - `mysql` and check contents of the source tables and `bkup` tables. Ensure that they contain the same number of records and the values make sense. 

## Types of changes

## Checklist:


bin/acl-config
```diff
@@ -381,12 +381,21 @@ function backupNonManagedTables(iDatabase $db, array $tables)
...
-
+    $handle = $db->handle();
     foreach($tables as $table => $queries) {
         $creationQuery = $queries['create'];
+        $quotedTable= $handle->quote($table . '_bkup');
+
+        $exists = count($db->query("SHOW TABLES LIKE $quotedTable")) > 0;
+
+        if ($exists) {
+            $db->execute("DROP TABLE ${table}_bkup");
```
jtpalmer: This would be simpler if it was just `DROP TABLE IF EXISTS ${table}_bkup`.

ryanrath: Thanks, I'll be pushing the change momentarily. Just running through the tests once more.

jpwhite4:
If the acl-config process does crash how does one recover the tables?

ryanrath:
It runs twice in a row? Which create table statements do you think will fail? There's really only one set of tables actually `CREATE`d in `acl-config` which are the non-managed backup tables and each creation is preceded by a `DROP TABLE IF EXISTS ...`.

jpwhite4:
Surely that defeats the purpose of this change if you always delete the table first?

ryanrath:
@jpwhite4 I've added the `recover` flag + accompanying code changes per our conversation yesterday ( and tested in Docker on OpenXDMoD / XSEDE ). Please let me know if there's anything else you'd like me to address. Thanks!

bin/acl-config
```diff
@@ -23,11 +23,13 @@ $opts = array(
     't' => 'dryrun',
     'v' => 'verbose',
     'd' => 'debug',
+    'r' => 'recover',
```
smgallo: Need to add help documentation for this option. Also, the `-h` docs should be formatted similar to other xdmod applications with the flags and optional arguments on the first line and the description indented on the line beneath:

```
-o, --option *tag=value*
    Description of the option
```

bin/acl-config
```diff
@@ -85,7 +91,9 @@ try {
  **/
 function main()
 {
-    global $dryRun, $log, $verify, $logLevel;
+    global $dryRun, $log, $verify, $logLevel, $recover;
+
+    $schema = 'moddb';
```
smgallo: This variable is only used in the call to `tablesExist()` below, why set it as a variable?

bin/acl-config
```diff
@@ -317,20 +325,44 @@ SQL;
         )
     );
 
-    backupNonManagedTables($db, $tablesToBeBackedUp);
+    $nonManagedBackupTables = array_reduce(
```
smgallo: In the data structure above this, it would be clearer if the array keys were `create_sql` and `insert_sql` (or `populate_sql`) rather than simply `create` and `populate`.

bin/acl-config
```diff
@@ -317,20 +325,44 @@ SQL;
...
+
+    $backupTablesExist = tablesExist($db, $nonManagedBackupTables, $schema, false);
+
+    if ($backupTablesExist && !$recover) {
+        $log->err('Backup tables detected! ');
+        $log->err('Unable to continue!');
+        $log->err('This may indicate a failed previous run.');
+        $log->err('Please see the XDMoD Troubleshooting documentation to resolve.');
+        exit(1);
+    } else if (!$backupTablesExist && $recover) {
```
smgallo: Is it necessary to stop here, or can we issue a warning that no recovery tables were found and continue?

bin/acl-config
```diff
@@ -317,20 +325,44 @@ SQL;
...
+    $nonManagedBackupTables = array_reduce(
+        array_keys($tablesToBeBackedUp),
+        function($carry, $item) {
+            $carry[] = "${item}_bkup";
+            return $carry;
+        },
+        array()
+    );
+
+    $backupTablesExist = tablesExist($db, $nonManagedBackupTables, $schema, false);
```
smgallo: Am I missing something, or is recovery never performed? `$recover` is only checked here in this conditional and I don't see action taken if it is set.

bin/acl-config
ryanrath: To indicate what the string value is being used for. I can certainly change it if you'd prefer.

bin/acl-config
ryanrath: It's not necessary for the execution of the script, I can change it to warn the user and continue.

smgallo: I think that would be the better option.

bin/acl-config
smgallo: Lets make it a constant instead and define it at the top of the script.

bin/acl-config
ryanrath: Kinda? The recovery is performed iff the backup tables are not overwritten ( via `backupNonManagedTables(...)`. 

So the state of these variables during a valid recovery would be:

- `backupTablesExist: true`
- `recovery: true` 

Which when used in: 
```php 
if ($backupTablesExist && !$recover) {
        $log->err('Backup tables detected! ');
        $log->err('Unable to continue!');
        $log->err('This may indicate a failed previous run.');
        $log->err('Please see the XDMoD Troubleshooting documentation to resolve.');
        exit(1);
} elseif (!$backupTablesExist) {
    if ($recover) {
        $log->warning('Recover mode specified but no backup tables exist.');
        $log->warning('Script execution will continue as normal.');
    }
    backupNonManagedTables($db, $tablesToBeBackedUp);
}
```

means that the `backupNonManagedTables` function is never called and that when execution reaches: 
```php
// Now re-populate the user related tables.
repopulateNonManagedTables($db, $tablesToBeBackedUp);
```
The pre-existing backup tables will be used and the recovery performed.

smgallo: If the backup tables already exist, how do you know if they have or have not been overwritten?

ryanrath: Because they are removed at the end of a successful run.

ryanrath: And at that point in the execution they would not have been populated

PR #667, opened 2018-09-07
Acl script consolidation

ryanrath:
## Description
These changes ensure that the only script that needs to be executed to bring
things up to date w/ Acls is `acl-config`.

Since `acl-config` now handles the execution of `acl-xdmod-management` and
`acl-import`, we can remove the scripts themselves and any references to them in
other code.

Select sql scripts have been removed from the `acl-imports` ETL process as they are no longer required.

## Motivation and Context
Simplifying the user experience interacting with the Acl system and removing code / scripts that are no longer needed. 

## Tests performed
Manual and Automated tests on Docker containers were conducted

## Types of changes

- [X] Code Restructuring / Simplification

## Checklist:


bin/acl-config
```diff
@@ -1991,6 +1998,42 @@ function verifyDatabase()
...
 }
 
+/**
+ * Attempt to run the ETL process identified by $processName.
+ *
+ * @param string $processName the name of the ETL process that is to be executed.
+ * @param bool $quiet         Defaults to true, if set to false then additional logging output may be
+ *                            generated
+ */
+function runEtlProcess($processName, $quiet=true)
```
plessbd: Instead of executing an external program we have started to call `EtlOverseer` directly
for an example see: https://github.com/ubccr/xdmod/blob/2a376565ff4daf64a6f93a9ff6e076a97522678c/classes/OpenXdmod/DataWarehouseInitializer.php#L285-L323
more of an FYI than a change requested

ryanrath: coolio, per our conversation I'll work on a PR after this one is merged that extracts this function and replaces code that executes etl pipelines / actions with the extracted function.

bin/acl-config
```diff
@@ -436,7 +439,10 @@ SQL;
 
     wipeManagedTables($db);
 
-    createManagedTables();
```
jpwhite4: If you are no longer calling this function then it needs to be removed from the code.

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -259,11 +259,7 @@ private function removeRolesFile($realm)
      */
     private function updateAcls()
     {
-        $manageAcls = new AclEtl(['section' => 'acls-xdmod-management']);
```
jpwhite4: The AclEtl class needs to be removed from the source code if it is no longer being used.

classes/OpenXdmod/Setup/AdminUserSetup.php
```diff
@@ -26,6 +27,17 @@ public function handle()
         $lastName     = $this->console->prompt('Last name:');
         $emailAddress = $this->console->prompt('Email address:');
 
+        // Retrieve the organization that XDMoD is currently setup for so that the Admin user can
+        // be associated with it. If we are unable to retrieve / lookup the organization then
+        // default to the 'Unknown' organization.
+        try {
```
jpwhite4: Please can you explain why this is changing.

Dockerfile
```diff
@@ -1 +1 @@
-FROM       tas-tools-ext-01.ccr.xdmod.org/xdmod-centos7:open7.5.1-v5
```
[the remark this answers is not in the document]
ryanrath: It was the same as w/ the sso user retrieval PR. The new one just had updated the user creation scripts to have valid organizations.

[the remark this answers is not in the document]
jpwhite4: This will have to wait until the 8.0.0 docker - we should not be putting 8.0 features into a 7.5 image.

classes/OpenXdmod/Setup/AdminUserSetup.php
ryanrath: At the time I thought it was desirable for all of our dynamically created users to have an appropriate organization value ( hence the docker version update etc. ). It's really not needed though. I've reverted this code in the latest commit.

classes/OpenXdmod/Setup/ResourcesSetup.php
ryanrath: Yup, removed in latest commits

bin/acl-config
ryanrath: Yup, removed in latest commits

bin/acl-config
ryanrath: @plessbd I've updated this code to utilize the `EtlConfiguration / EtlOverSeer` code.

bin/acl-config
```diff
@@ -110,6 +114,9 @@ function main()
         exit($result);
     }
 
+    // Make sure that all of the required tables are present
+    runEtlProcess('acls-xdmod-management');
```
plessbd: is there a reason we are doing this here and at line 447?
What is the point of running this twice?

ryanrath: This one is to ensure that any table changes are applied prior to acl-config doing anything. Line 447 is so that the tables we dropped in the line above it are re-created.

plessbd: Seems like there should be a better way to do this, but that would be a different PR anyway.

jpwhite4:
Don't forget to update the documentation in the docs/ directory to remove the references to the removed commands.

bin/acl-config
smgallo: You can run the TableManagement action and use the same config file. This will only update tables referenced in that file.

docs/configuration.md
```diff
@@ -46,36 +46,15 @@ XDMoD is installed you must create the databases manually.  Likewise, if you
...
-  
-this script executes a series of sql statements that are stored in 
-etc/etl/etl_sql.d/acls/xdmod/\*.sql.
+#### Acl Database Setup / Population
+ 
+This step will run immediately after you have setup the database that XDMoD will 
+be using and does not require any additional input. It is responsible for creating 
+and populating the tables required by the Acl framework.
+      
+If you're XDMoD Installation requires modifications to the acl tables
```
jpwhite4: s/you're/your/

docs/configuration.md
```diff
@@ -46,36 +46,15 @@ XDMoD is installed you must create the databases manually.  Likewise, if you
...
-etc/etl/etl_sql.d/acls/xdmod/\*.sql.
+#### Acl Database Setup / Population
+ 
+This step will run immediately after you have setup the database that XDMoD will 
+be using and does not require any additional input. It is responsible for creating 
+and populating the tables required by the Acl framework.
+      
+If you're XDMoD Installation requires modifications to the acl tables 
+(etc/etl/etl_tables.d/acls/xdmod/<table>.json) then running this step again or 
+the 'acl-config' bin script is required.
```
jpwhite4: use backticks rather than single quotes

bin/acl-config
```diff
@@ -2184,14 +2172,84 @@ SQL;
...
+    global $dryRun, $log;
+
+    $log->debug(
+        sprintf(
+            'Running ETL pipeline "%s" with parameters %s',
+            $processName,
+            json_encode($params)
+        )
+    );
+    if (!$dryRun) {
```
jpwhite4: Suggest passing the dryRun configuration setting to the EtlOverseer and have it execute the pipeline in dryr un mode.

PR #672, opened 2018-09-13
Hardening Password Reset Process

ryanrath:
## Description
Harden the password reset process.

**NOTE:** When testing this change it is important to look at emails generated during user creation, triggering a password reset via the internal dashboard, and login -> 'Forgot password'. 

## Motivation and Context
MD5 can be brute forced fairly easily in the time before the link is expired. We now use SHA256 to calculate the hashes.

## Tests performed
Manual Testing via docker

## Types of changes

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -2686,38 +2686,6 @@ SQL;
 
     }//_generateToken
 
-
-    // --------------------------------
-
-    /*
-     *
-     * @function validateRID
-     * (determines whether the reset identifier (RID) is valid -- used for account password updating)
-     *
-     * @param string $rid
-     *
-     * @return array:  if first element in the array is VALID, then the array structure will be (VALID, id, first_name)
-     *                 if first element in the array is INVALID, then the array structure will be (INVALID)
-     *
-     */
-
-    public static function validateRID($rid)
-    {
-
-        $pdo = DB::factory('database');
-
-        $accountCheck = $pdo->query("SELECT id, first_name FROM Users WHERE MD5(CONCAT(username, password_last_updated)) = :rid", array(
-            ':rid' => $rid,
-        ));
-
-        if (count($accountCheck) > 0) {
-            return array('status' => VALID, 'user_id' => $accountCheck[0]['id'], 'user_first_name' => $accountCheck[0]['first_name']);
-        } else {
-            return array('status' => INVALID);
-        }
-
-    }//validateRID
-
     /*
      * @function isSSOUser
      *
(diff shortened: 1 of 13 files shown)


ryanrath:
@jpwhite4 Agreed, I've just pushed a commit based on your email. I've completed manual tests for: 
  - Happy Path
  - Cryptographic Splicing ( Attempting to change the data within the token to reset a different users password ) fails.
  - That expiration is enforced

These were done both in a fresh install & upgrade. I do still need to test XSEDE and OpenXDMoD w/ SUPREMM.

But if you could take a look when you have a second that would be great.

classes/Models/Services/Users.php
```diff
@@ -418,4 +418,89 @@ public static function isPrincipalInvestigator(XDUser $user)
...
+        if ($now >= $expiration) {
+            $expirationDate = date('Y-m-d H:i:s', $expiration);
+            $log->debug("RID Token expired for: $userId | expired: $expirationDate");
+            return $results;
+        }
+
+        try {
+            $user = XDUser::getUserByID($userId);
+            $expected = explode('|', Users::generateRID($user, $expiration))[2];
+            $valid = \Symfony\Polyfill\Php56\Php56::hash_equals($expected, $hash);
```
jpwhite4: change this to 
```
$valid = hash_equals($expected, $hash);
```

composer.lock
```diff
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
```
jpwhite4: We need to not change all these dependencies this late in the development cycle.

html/controllers/user_auth/update_pass.php
```diff
@@ -16,7 +18,7 @@
 	
 	// -----------------------------
 	
-   $validationCheck = XDUser::validateRID($_POST['rid']);
```
jpwhite4: Why did validateRID get added to Users? Also why is it still in XDUser?

ryanrath: - Because I felt that's where it belonged.
- Because I forgot to remove it.

classes/XDUser.php
```diff
@@ -2989,4 +2957,88 @@ public function setOrganizationID($organizationID)
...
+
+        if ($now >= $expiration) {
+            $expirationDate = date('Y-m-d H:i:s', $expiration);
+            $log->debug("RID Token expired for: $userId | expired: $expirationDate");
+            return $results;
+        }
+
+        try {
+            $user = XDUser::getUserByID($userId);
+            $expected = explode('|', $user->generateRID($expiration))[2];
```
jpwhite4: getUserByID could return null -> you need to check for null return value

classes/XDUser.php
```diff
@@ -2989,4 +2957,88 @@ public function setOrganizationID($organizationID)
...
+
+    /**
+     * Generate an RID value from the provided $user. This value is intended to be used in
+     * authenticating / authorizing a password reset. If an $expiration value is provided, that will
+     * be used instead of generating one via the 'email_token_expiration' portal settings value.
+     *
+     * @param int|null    $expiration the date after which this rid is considered invalid.
+     * @return string in the form "userId|expiration|hash"
+     * @throws Exception If there are any missing configuration properties that this function relies
+     * on. These include: email_token_expiration, email_token_expiration_format, and
```
jpwhite4: thisd documentation needs to be updated

classes/XDUser.php
ryanrath: Inside the try?

html/controllers/user_auth/update_pass.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Models\Services\Users;
```
jpwhite4: remove this change

html/password_reset.php
```diff
@@ -7,14 +7,12 @@
    */
 
 	require_once dirname(__FILE__).'/../configuration/linker.php';
-
```
jpwhite4: remove this change

html/controllers/user_admin/pass_reset.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 use CCR\MailWrapper;
+use Models\Services\Users;
```
jpwhite4: remove this change

classes/XDUser.php
jpwhite4: the try will not catch the fatal error of dereferencing the null value. How did you test this case?

classes/XDUser.php
```diff
@@ -2989,4 +2957,89 @@ public function setOrganizationID($organizationID)
...
+    public function generateRID($expiration = null)
+    {
+        $expiresIn = \xd_utilities\getConfiguration('general', 'email_token_expiration');
+        $appSecret = \xd_utilities\getConfiguration('general', 'application_secret');
+
+        $userId = $this->_id;
+        $userSecret = $this->_password;
+
+        if ($expiration === null) {
+            $expiration = time() + $expiresIn;
```
chakrabortyr: What happens when both expiration and the config setting are null? Should this have some default value?

ryanrath: So, what happens when somebody goes in and manually removes these settings?

html/controllers/user_auth/update_pass.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+
```
chakrabortyr: Let's remove this.

html/password_reset.php
```diff
@@ -1,174 +1,208 @@
 <!DOCTYPE HTML>
```
jpwhite4: Please revert these change

ryanrath: The reformatting? Or the validation checking?

open_xdmod/modules/xdmod/component_tests/lib/RIDTest.php
```diff
@@ -0,0 +1,110 @@
...
+        );
+    }
+
+    /**
+     * @dataProvider provideGenerateRID
+     *
+     * @param array $options
+     */
+    public function testValidateRID(array $options) {
+
```
jpwhite4: Please double check this function.

classes/XDUser.php
```diff
@@ -2989,4 +2957,89 @@ public function setOrganizationID($organizationID)
...
+
+        try {
+            $user = XDUser::getUserByID($userId);
+            if ($user === null) {
+                return $results;
+            }
+            $expected = explode('|', $user->generateRID($expiration))[2];
+            $valid = hash_equals($expected, $hash);
+
+            $results['status'] = $valid ? 'success' : INVALID;
```
jpwhite4: update_pass.php checks for $validationCheck['status'] == VALID

classes/XDUser.php
ryanrath: resolving per our conversation.

PR #686, opened 2018-09-25
Update user emails to address by name.

jpwhite4:
The use of the word "Greetings" to start notification emails increases
the spam filtering score and makes it look like a phishing email. It
also, to my mind is a bit unprofessional. This changes it to address the
user directly with their name.

The 'greetings' also reminds me of the catchphrase of the Joey character from the mid 80s
sitcom Bread. https://www.youtube.com/watch?v=GSGWnzzRgT0


jpwhite4:
We will be rewriting much of this code in the next couple of days so this request will be OBE.

PR #688, opened 2018-09-26
Added Locale Specific Password Expiration Times in Forgot Password Email

chakrabortyr:
… template.

## Description
Addresses one of my gripes in Asana about our forgot password emails not explicitly stating when the reset link expires.

## Motivation and Context
Minimizes support questions, user friendliness.

## Tests performed
Sent a bunch of emails to myself in Docker

## Types of changes

## Checklist:

### DIFF
--- a/html/controllers/user_admin/pass_reset.php
+++ b/html/controllers/user_admin/pass_reset.php
@@ -45,6 +45,7 @@ try {
             'first_name'           => $user_to_email->getFirstName(),
             'username'             => $userName,
             'reset_url'            => $resetUrl,
+            'expiration'           => strftime("%c %Z", explode('|', $rid)[1]),
             'maintainer_signature' => MailWrapper::getMaintainerSignature(),
             'subject'              => "$page_title: Password Reset",
             'toAddress'            => $user_to_email->getEmailAddress()
--- a/html/controllers/user_auth/pass_reset.php
+++ b/html/controllers/user_auth/pass_reset.php
@@ -46,6 +46,7 @@ try {
             'first_name'           => $user_to_email->getFirstName(),
             'username'             => $user_to_email->getUsername(),
             'reset_url'            => $resetUrl,
+            'expiration'           => strftime("%c %Z", explode('|', $rid)[1]),
             'maintainer_signature' => MailWrapper::getMaintainerSignature(),
             'subject'              => "$page_title: Password Reset",
             'toAddress'            => $user_to_email->getEmailAddress()
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -2978,7 +2978,8 @@ SQL;
         $userSecret = $this->_password;
 
         if ($expiration === null) {
-            $expiration = time() + $expiresIn;
+            // set default expiration time of 10 minutes if we don't get one from configuration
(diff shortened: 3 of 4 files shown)


html/controllers/user_admin/pass_reset.php
```diff
@@ -34,6 +34,8 @@
     $userName = $user_to_email->getUsername();
 
     $rid = $user_to_email->generateRID();
+    $expireDate = strftime("%c %Z", explode('|', $rid)[1]);
+    $expiration = $expireDate ? $expireDate : "N/A";
```
jpwhite4: Under what circumstances would expireDate not be set? In these circumstances is 'N/A' the appropriate text to send to the user?

chakrabortyr: The real gem that is the PHP manual [doesn't document this](https://secure.php.net/manual/en/function.strftime.php), however, strftime() will return a bool when you pass in a nonsensical value to it. 

In my mind, it's appropriate to send "N/A" to the user, as the end user need not be aware of configuration snafu.

smgallo: "N/A" implies that there is no expiration date. If we don't know the date then "Unknown" is more appropriate.

html/controllers/user_auth/pass_reset.php
```diff
@@ -35,6 +35,8 @@
 
 try {
     $rid = $user_to_email->generateRID();
+    $expireDate = strftime("%c %Z", explode('|', $rid)[1]);
+    $expiration = $expireDate ? $expireDate : "Unknown";
```
jpwhite4: Under what circumstances will `$expireDate` not be set.

chakrabortyr: Sorry, to clarify: it will never not be set, it would be false.`$expireDate` could be false if you set the `email_token_expiration` attribute in the portal settings to empty or some other nonsense value.

chakrabortyr: Nvm, it's not as robust as I thought. We're going to need to do some extra validation on these timestamps as well...

jpwhite4: I still can't reproduce the $expireDate being false. Please can you tell me what I should set email_token_expiration to in my config file to get $expireDate to be false.

I tried the following:

If email_token_expiration is not set in the config file then this code will never get called (since the generateRID() will throw an exception). If the email_token_expiration is set to a non-numeric string then the timestamp will still be a number and strftime() will still return a string.

```
<?php
var_dump(strftime("%c %Z", time() + false));
var_dump(strftime("%c %Z", time() + true));
var_dump(strftime("%c %Z", time() + "A non numeric string"));
var_dump(strftime("%c %Z", time() + ""));
```

chakrabortyr: Blah, brain malfunction. That's actually not possible,

html/controllers/user_admin/pass_reset.php
```diff
@@ -34,6 +34,8 @@
     $userName = $user_to_email->getUsername();
 
     $rid = $user_to_email->generateRID();
+    $expireDate = strftime("%c %Z", explode('|', $rid)[1]);
```
jpwhite4: no need for the `$expireDate` variable here.

PR #689, opened 2018-09-26
Unlock org for unknown person

ryanrath:
## Description
- create_user.php: 
  - The organization associated with the newly created user was solely based off of the person they were mapped to. The organization and person mappings are now independent of each other. 
- update_user.php
  - The organization associated with the user to be updated was solely based off of the person they were mapped to. The organization and person mappings are now independent.
- TGUserDropDown.js
  - If there is a callback specified, it should be called. Regardless of whether or not the value has changed.
- SectionExistingUsers.js
  - Ensure that when the 'Unknown' person is mapped that the institution drop down is enabled for manual override.
- SectionNewUser.js
  - Ensure that when the 'Unknown' person is mapped that the institution drop down is enabled for manual override.

## Motivation and Context
We did not account for needing to manually override a User's associated
organization when they are mapped to the 'Unknown' person. These changes enable
that on the front end and make sure that they are correctly taken into account
on the back end.

## Tests performed
Manual Tests ( automated tests to follow )

New User
  - Create a new User
  - Map a Person other than 'Unknown' 
  - Ensure that the 'Institution' drop down is correctly populated and unable to be modified.
  - Map the 'Unknown' Person
  - Ensure that the 'Institution' drop down is able to be modified.

Existing User ( Not 'Unknown' )
  - Double click on an existing user not mapped to 'Unknown'
  - Change their person to 'Unknown'
  - Ensure that you can now update their 'Institution'
  - Select a new 'Institution'
  - Save changes
  - Close the 'Create / Manage Users' dialog window.
  - Double click on the same user
  - Ensure that their person / institution has not changed.
  - Force refresh the browser
  - Double click on the same user
  - Ensure that their person / institution has not changed.

Existing User ( 'Unknown' ) 
  - Double click on an existing user who is mapped to 'Unknown'
  - Change their person to anybody not == 'Unknown'
  - Ensure that the "Institution' drop down has been updated correctly
  - Ensure that you are not able to update the 'Institution' drop down.
  - Save changes
  - Close the 'Create / Manage Users' dialog window.
  - Double click on the same user
  - Ensure that their person / institution has not changed.
  - Force refresh the browser
  - Double click on the same user
  - Ensure that their person / institution has not changed.

## Types of changes

## Checklist:

### DIFF
--- a/html/controllers/user_admin/create_user.php
+++ b/html/controllers/user_admin/create_user.php
@@ -73,7 +73,7 @@ try {
         $_POST['last_name'],
         array_keys($acls),
         ROLE_ID_USER,
-        Organizations::getOrganizationIdForPerson($_POST['assignment']),
+        $_POST['institution'],
         $_POST['assignment']
     );
     $newuser->setUserType($_POST['user_type']);
@@ -97,6 +97,14 @@ try {
         }
     }
     // =============================
+
+    // 'institution' now corresponds to a Users organization and will always be present, not only
+    // when a user has been assigned the campus champion acl. This means we need to update the logic
+    // that gates  the `setInstitution` function call to include a check if the user has been
+    // assigned the Campus Champion acl.
+    if (isset($_POST['institution']) && in_array(ROLE_ID_CAMPUS_CHAMPION, array_keys($acls))) {
+        $newuser->setInstitution($_POST['institution']);
+    }
 
     $page_title = \xd_utilities\getConfiguration('general', 'title');
     $site_address = \xd_utilities\getConfigurationUrlBase('general', 'site_address');
--- a/html/controllers/user_admin/update_user.php
+++ b/html/controllers/user_admin/update_user.php
@@ -110,6 +110,9 @@ if (isset($_POST['user_type'])) {
 
 }
 
+// Store this users original set of acls before they are possibly modified below.
+$originalAcls = $user_to_update->getAcls(true);
+
 // ===========================================
 // Make sure that we're not attempting to enable / disable the user before
 // processing 'acls'
@@ -154,20 +157,22 @@ if (!isset($_POST['is_active'])) {
     } // if (isset($_POST['acls'])) {
 }
 
-// -----------------------------
-
-$user_to_update->setOrganizationID(Organizations::getOrganizationForUser($user_to_update->getUserID()));
-// -----------------------------
-
+// 'institution' now corresponds to a Users organization and is not only present when they are a
+// campus champion. This means we need to make sure that the User's organization_id is populated but
+// that the old behavior of having `setInstitution` called w/ the 'institution' value is still
+// retained as this will have an effect on the results of CampusChampionRole's `getIdentifier`
+// function ( which is often displayed to the user ).
 if (isset($_POST['institution'])) {
+    $user_to_update->setOrganizationID($_POST['institution']);
 
-    if ($_POST['institution'] == -1) {
+    $oldCampusChampion = in_array(ROLE_ID_CAMPUS_CHAMPION, $originalAcls);
+    $newCampusChampion = in_array(ROLE_ID_CAMPUS_CHAMPION, array_keys($acls));
+
+    if ($newCampusChampion && !$oldCampusChampion) {
+        $user_to_update->setInstitution($_POST['institution']);
+    } elseif (!$newCampusChampion && $oldCampusChampion) {
         $user_to_update->disassociateWithInstitution();
-    } else {
-        $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
-        $user_to_update->setInstitution($_POST['institution'], $isPrimary);
     }
-
 }//if (isset($_POST['institution']))
 
 try {
--- a/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
+++ b/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
@@ -912,7 +912,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
         var mnuActions = new Ext.menu.Menu({
             plain: true,
             showSeparator: false,
-            cls: 'no-icon-menu',
+            cls: 'no-icon-menu existing_users_action_menu',
 
             items: [
                 {
--- a/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
+++ b/html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
@@ -203,6 +203,14 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
             /* eslint-disable no-use-before-define */
             settingsAreDirty = true;
             saveIndicator.show();
+
+            // If the currently selected person is the Unknown person then make sure we
+            // enable the institution drop down for manual overrides.
+            if (cmbUserMapping.getValue() === '-1') {
+                cmbInstitution.setDisabled(false);
+            } else {
+                cmbInstitution.setDisabled(true);
+            }
             /* eslint-enable no-use-before-define */
         };
 
@@ -1164,6 +1172,13 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
                         }
 
                         cmbInstitution.initializeWithValue(json.user_information.institution, json.user_information.institution_name);
+
+                        // If the current person mapped is the Unknown person then make
+                        // sure to enable the institution drop down for manual override.
+                        // we don't need an else as it's already setDisabled(true) above.
+                        if (cmbUserMapping.getValue() === '-1') {
+                            cmbInstitution.setDisabled(false);
+                        }
 
                         /**
                          * acls are in the form:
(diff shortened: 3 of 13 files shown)


html/controllers/user_admin/update_user.php
```diff
@@ -154,20 +154,8 @@ function ($carry, Acl $item) {
...
 if (isset($_POST['institution'])) {
-
-    if ($_POST['institution'] == -1) {
-        $user_to_update->disassociateWithInstitution();
-    } else {
-        $isPrimary = isset($acls) ? array_key_exists(ROLE_ID_CAMPUS_CHAMPION, $acls) : false;
-        $user_to_update->setInstitution($_POST['institution'], $isPrimary);
-    }
-
+    $user_to_update->setOrganizationID($_POST['institution']);
```
jpwhite4: Why is the institution no longer being set here?

html/gui/js/TGUserDropDown.js
```diff
@@ -135,8 +135,8 @@ CCR.xdmod.ui.TGUserDropDown = Ext.extend(Ext.form.ComboBox, {
 
                         var value = json.results[valueProperty];
 
-                        if (comp.getValue() !== value && callback !== undefined) {
-                            callback(comp.getValue(), value);
+                        if (callback !== undefined) {
```
jpwhite4: How come the callback must be called every time now? If the value didn't change then surely the enable/disable state of the drop down also does not change.

html/controllers/user_admin/update_user.php
ryanrath: Because I like to break things? In actuality this shouldn't have been removed and I've added it back in w/ additional checks ( we only want to call this if the user has been assigned the campus champion acl ). I've also added it back to `create_user.php` w/ the same checks.

html/gui/js/TGUserDropDown.js
ryanrath: Yeah, that was a debug change that shouldn't have made it in. Revert incoming.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -45,7 +45,16 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
             width: 150,
             cascadeOptions: {
                 component: cmbInstitution,
-                valueProperty: 'id'
+                valueProperty: 'id',
+                callback: function (control, oldValue, newValue, originalValue) {
```
jpwhite4: What is the "originalValue" function argument for? Where does it come from? The call in TGUser does not provide it and it appears to be unused here.

html/gui/js/TGUserDropDown.js
```diff
@@ -136,7 +136,7 @@ CCR.xdmod.ui.TGUserDropDown = Ext.extend(Ext.form.ComboBox, {
                         var value = json.results[valueProperty];
 
                         if (comp.getValue() !== value && callback !== undefined) {
-                            callback(comp.getValue(), value);
+                            callback(comp, comp.getValue(), value);
```
jpwhite4: What are the consequences of changing the arguments to the callback function? Will this break existing code that was expecting the first argument to be the new value?

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: Just  more debugging code, removing. Also going to run through the changes and see if I can find / cleanup any more code that shouldn't have made it in.

html/controllers/user_admin/update_user.php
jpwhite4: Why do you no longer call `disassociateWithInstitution()`? If this function is no longer required then it should be removed from the code. (However, I suspect that you do need to call it to make sure that the ACLs are updated properly in the case where a user loses the campus champion role).

html/gui/js/TGUserDropDown.js
ryanrath: *What are the consequences of changing the arguments to the callback function?*
There aren't any as the only places the callback is provided are `SectionNewUser.js` and `SectionExistingUser.js`. Of these, only `SectionNewUser.js` utilize(d) any of the arguments provided to the callback. And of those arguments, the only ones it used were the newly added `component` and the `newValue`. Now,  With that being said, this change doesn't really need to happen as we have the component that needs to be updated  in scope when the callback is defined. I'll be updating that shortly.

*Will this break existing code that was expecting the first argument to be the new value?*
Nope, see above*

html/controllers/user_admin/update_user.php
ryanrath: It's not really the ACLs that need it but `CampusChampionRole` does still need it so that it's `getIdentifier(true)` function returns the correct value.

ryanrath: Just pushed a commit w/ `disassociateWithInstitution` back in & some added logic to detect when it should be called ( i.e. when a user has the cc acl removed ).

jpwhite4: Still confused why any of the [set|disassociateWith]Institution logic is being changed at all. What bug is being fixed?

ryanrath: Short answer is that the institution logic is being changed because the `institution` now means something different than it did pre-"Users are directly associated with an Organization". 

Previously this parameter was only present if the user had been assigned the Campus Champion role ( but still contains `organization_id`). Now it is always present. To account for this change it is necessary to change when `[set|disassociateWith]Institution` is called.

In so far as bugs are concerned, during user creation:
  - In OpenXDMoD: none
  - In XSEDE: new Campus Champion Users would have their 'Top Role' display incorrectly. It wouldn't include the organization that they are associated with.
    - NOTE: basically the result of `CampusChampionRole::getIdentifier(true)` would not be correct.

during user update:
  - In OpenXDMoD: none
  - In XSEDE: After a user would be updated w/ the Campus Champion role ( and that's their most privileged role ) the 'Top Role' value in 'My Profile' would be incorrect.
    - NOTE: see during user creation*

ryanrath:
I'm working on some UI tests for this one. After they're done we'll have the infrastructure in place to create more Internal Dashboard UI tests.

html/gui/js/CCR.js
```diff
@@ -1322,7 +1322,7 @@ CCR.xdmod.ui.userManagementMessage = function (msg, success) {
     new Ext.ux.window.MessageWindow({
         title: 'User Management',
         width: 300,
-        html: '<b style="color: ' + styleColor + '">' + msg + '</b>' || 'No information available',
+        html: '<b style="color: ' + styleColor + '" class="user_management_message">' + msg + '</b>' || 'No information available',
```
jpwhite4: Not sure that we should be changing this in this pull request. Is there a bug with the display?

ryanrath: There wasn't a bug with the display. The change is to support ease of selection for testing.

html/controllers/user_admin/create_user.php
```diff
@@ -133,6 +141,6 @@ function ($carry, Acl $item) {
 
 $returnData['success'] = true;
 $returnData['user_type'] = $_POST['user_type'];
-$returnData['message'] = 'User <b>'.$_POST['username'].'</b> created successfully';
+$returnData['message'] = 'User '.$_POST['username'].' created successfully';
```
jpwhite4: How com this changed?

ryanrath: Same reason that `update_user.php` was changed. Specifically, to make it easier ( Possible? ) to select the element that ends up containing `$returnData['message'] to verify that the user has been created successfully during testing. If you happen to have some thoughts on how that could be accomplished without this change I'm definitely all for it.

ryanrath:
@jpwhite4 I've squashed the changes into to two commits, one for the functional changes and the other that contains the UI tests and related changes. Hoping that helps with the review. Thank you for taking a look at things.

html/controllers/user_admin/create_user.php
jpwhite4: Here you go:
```javascript
successNotification: function (username) { 
    return module.exports.selectors.modalDialogByTitle('User Management') +
            '//b[text()[1][contains(., "User")] and text()[2][contains(., "created successfully")]]/b[text() = "' + 
            username + '"]/ancestor::node()[1]';
} 
```
the definition of `modalDialogByTitle` is in another comment response below.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -540,6 +548,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
             }
 
             Ext.Msg.show({
+                cls: 'delete_user',
```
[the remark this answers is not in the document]
ryanrath: Sweet, I'll add that in there.

html/controllers/user_admin/create_user.php
ryanrath: Sweet, I'll add this. I will also note that the `<b>` tags aren't actually providing any additional styling. The message itself is already in bold. But this does mean that the file doesn't have to change.

PR #691, opened 2018-10-02
Updating who is notified when SSO unknown org occurs

ryanrath:
## Description
- We are no longer notifying the user when we are unable to identify which
  organization they are associated with.
- We are now **always** notifying the admin when we assign a user the unknown
  organization.
  - Since we're always notifying the admin we no longer need the
  `email_admin_sso_unknown_org` property.

## Motivation and Context
We should make sure to only notify the people who can do something about a particular situation. 

## Tests performed
Manual Tests were performed: 
- Sign in via SSO with:
  - A User that has an organization that will not be identified
- Check that an email was sent correctly to the admin address

- Sign in via SSO with:
  - A User that has an organization that will be identified
  - Check that no email was sent

- Sign in via SSO with:
  - A User that will be matched with a valid person 
  - Check that no email was sent.

## Types of changes

## Checklist:

### DIFF
--- a/classes/Authentication/SAML/XDSamlAuthentication.php
+++ b/classes/Authentication/SAML/XDSamlAuthentication.php
@@ -57,21 +57,6 @@ Unable to Identify Users Organization.
 Additional Setup Required.
 EML;
 
-    const USER_EMAIL_SUBJECT = 'XDMoD SSO User: Additional Actions Required';
-
-    const USER_EMAIL_BODY = <<<EML
-
-Dear %s
-
-This email is notify you that XDMoD was unable to determine which organization to associate you with.
-Administrative Users have been notified that additional setup will be required. You may not have full
-access until this additional setup is complete.
-
-Thank you,
-
-XDMoD SSO User Creation
-EML;
-
     private $logger = null;
 
     public function __construct()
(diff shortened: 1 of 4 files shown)


docs/simpleSAMLphp.md
```diff
@@ -76,8 +76,6 @@ Conditionally, if your installation supports multiple organizations then you mus
 There are two configuration properties in `portal_settings.ini` that support and interact with this SAML property.
```
plessbd: this needs to change since there are no longer two configuration properties.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -311,41 +296,28 @@ public function getLoginLink()
...
-            $emailAdminOnUnknownOrg = \xd_utilities\getConfiguration('sso', 'email_admin_sso_unknown_org') === 'on';
-        } catch (Exception $e) {
-            $this->notify(
-                $techSupportRecipient,
-                $techSupportRecipient,
-                '',
-                $techSupportRecipient,
-                'Error retrieving XDMoD configuration property "email_admin_sso_unknown_org" form portal_settings.ini',
-                $e->getMessage()
+        // We only notify admins iff we were unable to identify which organization the user should
```
plessbd: i believe `iff` should be `if`

PR #697, opened 2018-10-11
allow order to be set on FOS

plessbd:
the field of science hierarchy was not getting the order set.
This will automatically set it.

### DIFF
--- a/configuration/etl/etl.d/hpcdb-xdw.json
+++ b/configuration/etl/etl.d/hpcdb-xdw.json
@@ -54,7 +54,8 @@
     }, {
         "name": "field-of-science-hierarchy",
         "definition_file": "jobs/xdw/field-of-science-hierarchy.json",
-        "description": "field of science hierarchy records"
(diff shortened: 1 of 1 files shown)


jtpalmer:
I'm not familiar with the query optimizations.  How does not optimizing this query result in the order being set?

plessbd:
It forces allowSingleDatabaseOptimization (https://github.com/ubccr/xdmod/blob/xdmod8.0/classes/ETL/Ingestor/pdoIngestor.php#L975) to return `false` which makes it go down the `multiDatabaseIngest` path and allows the following code to run:

https://github.com/ubccr/xdmod/blob/xdmod8.0/classes/ETL/Ingestor/pdoIngestor.php#L755-L772

plessbd:
Yeah @jtpalmer and I were talking about how `optimize_query` in this case doesn't accurately describe what it is doing

smgallo:
It's moved beyond what it was originally intended to do.

PR #698, opened 2018-10-11
add fos-hierarchy to docker;update test artifacts

plessbd:
Adds a fos-hierarchy to the docker builds and updates the test artifacts to take these into account.


plessbd:
this will wait until there is an 8.0 docker;  this requires the hierarchy and group-to-hierarchy to be doe as part of the initial install.

plessbd:
done as part of #724

PR #701, opened 2018-10-12
Update Sign On Panel

plessbd:
## Description
Force for the local login to be collapsed when using SSO with an option to not if provided.
Fix sign in form saving user and pass by browser when using local login.

Since I was already near some of the stuff in index.php I added a few new lines to make the javascript that is output a little cleaner.

## Motivation and Context

Users and developers complained

## Tests performed

Manually verified the outcome when SSO enabled and not and various combinations of portal setting
## Types of changes


configuration/portal_settings.ini
```diff
@@ -50,6 +50,7 @@ multiple_service_providers = "off"
 
 [sso]
 force_default_organization = "on"
+colapse_local_login = "true"
```
smgallo: I think that we should start adding one line comments for each new setting that we add to the settings file. Eventually, we should go back and add them for all settings.

html/gui/js/CCR.js
```diff
@@ -978,6 +978,9 @@ CCR.xdmod.ui.actionLogin = function (config, animateTarget) {
 
                 if (success) {
                     decodedResponse = CCR.checkDecodedJSONResponseSuccess(data);
+                    if (Ext.isGecko) {
```
smgallo: Why are you singling out the Geico mascot?

html/index.php
```diff
@@ -312,10 +312,20 @@ function isReferrer($referrer)
                 // This will catch when a configuration directory does not exist if it is set in the environment level
             }
             if ($auth && $auth->isSamlConfigured()) {
+                $ssoColapse = 'true';
```
smgallo: Collapse should have 2 "l"s

html/gui/js/CCR.js
```diff
@@ -1090,6 +1113,15 @@ CCR.xdmod.ui.actionLogin = function (config, animateTarget) {
 
     if (CCR.xdmod.isSSOConfigured) {
         loginItems.push(SSOLoginFrm);
+        if (CCR.xdmod.SSOcolapseLocalLogin) {
+            localLoginFrm.collapsible = true;
```
smgallo: Who didn't want to spell out "Form"?!?

html/gui/js/CCR.js
plessbd: Because I dont like how he licks his eyeball, it freaks me out.

Seriously, because isGecko actually refers to Firefox and it is the one that has to have the form submit happen to allow the password save functionality of the browser triggered.

html/gui/js/CCR.js
plessbd: I am going to blame: https://github.com/ubccr/xdmod/pull/297

configuration/portal_settings.ini
```diff
@@ -48,6 +48,11 @@ singlejobviewer = "off"
 ; set to 'off' in Open XDMoD until support has been added.
 multiple_service_providers = "off"
 
+[sso]
+; used to show local sign in portion of the sign in modal
+; setting to true will show it.
+show_local_login = "false"
```
smgallo: All of the other entries in this file use `on` or `off`, we should be consistent.

plessbd: this just makes me want to change them all to true or false...

smgallo: Yeah, but then will all of the place where they're used still work or will you need to change all of them too.

configuration/portal_settings.ini
```diff
@@ -48,6 +48,12 @@ singlejobviewer = "off"
 ; set to 'off' in Open XDMoD until support has been added.
 multiple_service_providers = "off"
 
+[sso]
+force_default_organization = "on"
+; used to show local sign in portion of the sign in modal
+; setting to on will show it.
```
chakrabortyr: I think this should just be one line: "Set to on to enable the local user option in login modal."

plessbd: done please re review

PR #702, opened 2018-10-12
Allow search panel scroll bars on small displays

jpwhite4:
The job viewer search panel is hardcoded to ~1000 x 600 px and cannot be resized and scrollbars are disabled. It gets aligned to the top left so is unusable on a screen that is shorter than ~600px since the buttons at the bottom are not displayed. 

This change makes it so that the panel window will show scrollbars if necessary so that the ui component is usable on small displays.

### DIFF
--- a/html/gui/js/modules/job_viewer/SearchPanel.js
+++ b/html/gui/js/modules/job_viewer/SearchPanel.js
@@ -65,6 +65,7 @@ XDMoD.Module.JobViewer.SearchPanel = Ext.extend(Ext.Panel, {
             Ext.apply(this, {
                     layout: 'table',
                     width: 1000,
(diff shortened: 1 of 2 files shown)


html/gui/js/modules/job_viewer/SearchPanel.js
```diff
@@ -65,6 +65,7 @@ XDMoD.Module.JobViewer.SearchPanel = Ext.extend(Ext.Panel, {
             Ext.apply(this, {
                     layout: 'table',
                     width: 1000,
+                autoScroll: true,
```
jtpalmer: Indentation looks odd here, but I'm guessing this is the correct indentation and everything else is wrong.

PR #706, opened 2018-10-17
Make "Show raw data" for multiple realms configurable

jpwhite4:
There were a couple of places where the realm setting was hardcoded. This change makes the information about the raw data realms loaded from (modular) configuration files and makes the realm selectable in the job viewer.


classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1761,17 +1768,26 @@ private function processHistoryRequest(Application $app, XDUser $user, $realm, $
...
     {
+        $config = \Xdmod\Config::factory();
+
+        $rawstats = $config['rawstatistics'];
+
+        $results = array();
+
+        if (isset($rawstats['realms'])) {
+            foreach($rawstats['realms'] as $realm => $realmconfig) {
+                $results[] = array('dtype' => 'realm',
```
plessbd:
```suggestion
                $results[] = array(
                    'dtype' => 'realm',
```
This should be on the next line

(I dont think this suggestion thing works right though it is showing some html stuff in there that I dont trust...)

plessbd: This should be on the next line
```suggestion
                $results[] = array(
                    'dtype' => 'realm',
```

PR #711, opened 2018-10-25
Add resource_id to where clause for cloud instance etl action

eiffel777:
Add resource_id to where clause for cloud instance etl action for joining to both the account table and the user table

## Motivation and Context
If the resource_id is not used in the where clause it is possible that the wrong account or user can be assigned to an instance if the same project name or username exists for two resources.

## Tests performed
Tested in local docker manually and ran all component, integration and regression tests

## Types of changes

## Checklist:


jpwhite4:
Are there any other places in the cloud code where resource_id condition is absent?

eiffel777:
@jpwhite4 It looks like it was missing from the action for volumes as well. I've updated it.

PR #712, opened 2018-10-25
Update Mongo Driver to a currently maintained version

jpwhite4:
## Description
The nodejs mongo driver version support period ended (a while ago). This updates to a supported version.

Of course, the API changed in between versions so there are some code changes that accompany the driver update.


jpwhite4:
See also https://github.com/ubccr/xdmod-supremm/pull/128

PR #716, opened 2018-11-01
Add new functionality to Usage chart requests

ryanrath:
# Note: This is for 8.1

## Description
Added the ability for Usage chart requests to provide non-numeric values for
'_filter' parameters. The new feature will attempt to translate these
non-numeric values into appropriate id values that can be used during chart
generation. Specifically this new feature is to support integration w/
Coldfront.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -1160,8 +1160,7 @@ class Usage extends Common
 
             $stmt = $db->prepare($query);
             $stmt->execute(array(':value' => $usageFilterValue));
-
-            return implode(',', $stmt->fetchAll(PDO::FETCH_COLUMN, 0));
+            return $stmt->fetch()[0];
         }
 
         return $usageFilterValue;
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -1174,7 +1174,7 @@ class Usage extends Common
             }
 
             // If a string was provided but no id(s) were found then exception out.
-            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterValue));
+            throw new Exception(sprintf("Invalid value detected for filter '%s': %s", $usageFilterType, $usageFilterValue));
         }
 
         return array($usageFilterValue);
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -1174,7 +1174,7 @@ class Usage extends Common
             }
 
             // If a string was provided but no id(s) were found then exception out.
-            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterType));
+            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterValue));
         }
 
         return array($usageFilterValue);
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -1158,7 +1158,7 @@ class Usage extends Common
                 ',',
                 array_reduce(
                     $rows,
-                    function($carry, $item) {
+                    function ($carry, $item) {
                         $carry[] = $item['value'];
                         return $carry;
                     },
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -1174,7 +1174,7 @@ class Usage extends Common
             }
 
             // If a string was provided but no id(s) were found then exception out.
-            throw new Exception("Invalid filter value detected");
+            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterType));
         }
 
         return array($usageFilterValue);
--- a/classes/DataWarehouse/Access/Usage.php
+++ b/classes/DataWarehouse/Access/Usage.php
@@ -976,15 +976,17 @@ class Usage extends Common
             $usageFilterValues = explode(',', $usageValue);
 
             foreach ($usageFilterValues as $usageFilterValue) {
-                $usageFilterValue = $this->translateFilterValue($usageFilterType, $usageFilterValue);
+                $translatedValues = $this->translateFilterValue($usageFilterType, $usageFilterValue);
 
-                $meFilters[] = array(
-                    'id' => "$usageFilterType=$usageFilterValue",
-                    'value_id' => $usageFilterValue,
-                    'dimension_id' => $usageFilterType,
-                    'realms' => array($usageRealm),
-                    'checked' => true,
-                );
+                foreach($translatedValues as $translatedValue) {
+                    $meFilters[] = array(
+                        'id' => "$usageFilterType=$translatedValue",
+                        'value_id' => $translatedValue,
+                        'dimension_id' => $usageFilterType,
+                        'realms' => array($usageRealm),
+                        'checked' => true,
+                    );
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


docs/software-requirements.md
```diff
@@ -76,7 +76,7 @@ added with this command for CentOS 7:
     # yum install httpd php php-cli php-mysql php-gd php-mcrypt \
                   gmp-devel php-gmp php-pdo php-xml \
                   php-pear-MDB2 php-pear-MDB2-Driver-mysql \
-                  java-1.7.0-openjdk java-1.7.0-openjdk-devel \
+                  java-1.8.0-openjdk java-1.8.0-openjdk-devel \
```
jpwhite4: I don't think this documentation change belongs in this pull request.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1112,87 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+         * This function will execute the provided $query with $values as the set of parameters. If
+         * any records are found then their 'value' columns are returned.
+         *
+         * @param string $query the sql query to be executed
+         * @param array $values the one parameter that will be provided to $query.
+         *
+         * @return int[]|array all current queries return an int array. If no rows are found then
+         * an empty array is returned.
+         */
+        $lookupValue = function ($query, array $values) {
```
jpwhite4: Note that the PHP PDO already has a function that will return the values of one column in an array. 
```php
$lookupvalue = function ($query, array $values) {
    $stmt = $db->prepare($query);
    $stmt->execute($values);

    return implode(',', $stmt->fetchAll(PDO::FETCH_COLUMN, 0) );
};
```

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1112,87 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+     *                                 values: pi, resource, project
+     * @param mixed  $usageFilterValue the value to be translated
+     * @return int|string an int will be returned if, the $usageFilterValue is not numeric and a
+     * record is found. Else, the unmodified $usageFilterValue is returned.
+     */
+    private function translateFilterValue($usageFilterType, $usageFilterValue)
+    {
+
+        /**
+         * Anonymous function to prevent code duplication. $query is expected to minimally fulfill
```
jpwhite4: I note that this anonymous function does not prevent code duplication since it only ever needs to be called once per call to translateFilterValues. In fact it results in more lines of code than simply having the code inline at the bottom of the switch statement and having the switch statement set the value of the $query.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1112,87 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+                        $carry[] = $item['value'];
+                        return $carry;
+                    },
+                    array()
+                )
+            );
+        };
+
+        if (!is_numeric($usageFilterValue)) {
+            $usageFilterValues = explode(',', $usageFilterValue);
```
jpwhite4: The `$usageFilterValue` variable already went through an `explode()` on line 976. Therefore it cannot have any more commas to be exploded again and the bulk of this code is unnecessary. Please can you double check this.

docs/software-requirements.md
ryanrath: This has been reverted.

classes/DataWarehouse/Access/Usage.php
ryanrath: Very cool, thanks! This will be included in the next commit.

classes/DataWarehouse/Access/Usage.php
ryanrath: Yeah... best intentions and all... Thanks for the catch. The code will be updated and actually simplified in the next set of commits.

classes/DataWarehouse/Access/Usage.php
ryanrath: Double checked, you are correct, this value doesn't need to be exploded again. Here is the function body w/ updates so far ( Just finishing up writing integration tests before committing ) : 
```php
        // If the value is already a number then just return it. We do this because this function is
        // about tranlsating a non-numeric value to a numeric one ( i.e. an id )
        if (is_numeric($usageFilterValue)) {
            return $usageFilterValue;
        }

        $query = null;

        switch ($usageFilterType) {
            case 'pi':
                // NOTE: At the time of writing there is no instance where multiple records have
                // the same username but different person_id values.
                $query = "SELECT DISTINCT sa.person_id AS value FROM modw.systemaccount sa WHERE sa.username = :value;";
                break;
            case 'resource':
                $query = "SELECT id AS value FROM modw.resourcefact WHERE code = :value";
                break;
            case 'project':
                $query = "SELECT account_id AS value FROM modw_cloud.account WHERE display = :value";
                break;
        }

        if ($query !== null) {

            $stmt = $db->prepare($query);
            $stmt->execute(array(':value' => $usageFilterValue));

            return implode(',', $stmt->fetchAll(PDO::FETCH_COLUMN, 0));
        }

        return $usageFilterValue;
```

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1112,58 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+                break;
+        }
+
+        if ($query !== null) {
+            $db = DB::factory('database');
+
+            $stmt = $db->prepare($query);
+            $stmt->execute(array(':value' => $usageFilterValue));
+
+            return implode(',', $stmt->fetchAll(PDO::FETCH_COLUMN, 0));
```
jpwhite4: I don't think you want to implode all results here. The rest of the code will only work if there is a single return value.

jpwhite4:
What is the expected behavior of the code if a string value is supplied that does not match in the database. I.e if I ask for filtering by a PI that does not exist.

ryanrath:
@smgallo Thoughts? What currently occurs is that that particular filter / value combo is ignored. So, `resource_filter=frearson,robertson&pi_filter=superunknownpi` is the same as `resource_filter=frearson,robertson`

smgallo:
Ideally, filtering on an unknown entity would generate an error from the underlying code.

classes/DataWarehouse/Access/Usage.php
smgallo: As per discussion earlier with @jpwhite4 and @ryanrath an upcoming commit will return an array of zero or more elements that can be iterated over to properly add the filters.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1156,14 +1155,38 @@ private function translateFilterValue($usageFilterType, $usageFilterValue)
         }
 
         if ($query !== null) {
+            $filterValueIsString = preg_match('/[\'"]+/', $usageFilterValue) === 1;
```
smgallo: This regex will match quotes anywhere in the string not just those surrounding it. It should be anchored at the start of the string and shouldn't have the `+` (e.g., `/^[\'"]/`).

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1156,14 +1155,38 @@ private function translateFilterValue($usageFilterType, $usageFilterValue)
...
+
+            $value = $filterValueIsString
+                ? preg_replace('/[\'"]+/', '', $usageFilterValue)
```
smgallo: This regex will trim any quote character from the string, which may not be what we want if we allow passing of a person's name in the future as it will remove apostrophes from anywhere in the string. We would be better served using `trim($usageFilterValue, '\'"')`.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1156,14 +1155,38 @@ private function translateFilterValue($usageFilterType, $usageFilterValue)
...
-
-            return implode(',', $stmt->fetchAll(PDO::FETCH_COLUMN, 0));
+            $stmt->execute(array(':value' => $value));
+
+            $rows = $stmt->fetchAll(PDO::FETCH_COLUMN, 0);
+
+            if ($rows !== false) {
+                $rowCount = count($rows);
+
+                if ($rowCount === 0 && !$filterValueIsString) {
```
smgallo: Since we are now forcing the caller to specify whether the value is an integer id or a string that needs lookup in a dimension table, we can get rid of this logic. If the value is not enclosed in quotes we can return immediately.

classes/DataWarehouse/Access/Usage.php
ryanrath: It should also be anchored at the end probably `/^[\'"].*[\'"]$/`

classes/DataWarehouse/Access/Usage.php
ryanrath: Nice, didn't know that PHP's trim allowed a char list to be trimmed. I'll update that.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1115,68 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+
+            $rows = $stmt->fetchAll(PDO::FETCH_COLUMN, 0);
+
+            // We need to test if there was an error ( bool returned ) because count(bool) === 1.
+            if ($rows !== false && count($rows) > 0) {
+                return $rows;
+            }
+
+            // If a string was provided but no id(s) were found then exception out.
+            throw new Exception("Invalid filter value detected");
```
smgallo: Lets include the invalid filter in the error message:
```suggestion
            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterValue));
```

ryanrath: *facepalm* Ignore the most recent commit, added the type instead of the value.

classes/DataWarehouse/Access/Usage.php
ryanrath: Update is in the latest version of the PR

ryanrath:
Alrighty, @jpwhite4 @smgallo I think I've got things up to date w/ the comments on here / what we've talked about ( and updated w/ xdmod8.1 ). If you could take another look at things and let me know if there's any other changes you'd like to see before merging that would rock.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -1107,4 +1115,68 @@ private function generateFilename($title, $start_date, $end_date, $isTimeseries)
...
+
+            $rows = $stmt->fetchAll(PDO::FETCH_COLUMN, 0);
+
+            // We need to test if there was an error ( bool returned ) because count(bool) === 1.
+            if ($rows !== false && count($rows) > 0) {
+                return $rows;
+            }
+
+            // If a string was provided but no id(s) were found then exception out.
+            throw new Exception(sprintf("Invalid filter value detected: %s", $usageFilterValue));
```
smgallo: One last (nitpicky) thing that I should have mentioned before - lets add the type in there as well:
```suggestion
            throw new Exception(sprintf("Invalid filter value detected for '%s': %s", $usageFilterType, $usageFilterValue));
```

PR #717, opened 2018-11-01
Remove MinMaxDate Table from TimeAggregationUnits

chakrabortyr:
# Note: This is for 8.1

## Description
This is phase 1 of removing the minmaxdate table. Instead we calculate date ranges by directly hitting the aggregate tables for the respective aggregation period. If this doesn't have any hiccups, a subsequent pull request will eliminate the minmaxdate table from the ETL process.

## Motivation and Context
Since we have non HPC Job realms, we don't want to predicate date ranges on one specific realm.

## Tests performed
Ran regression tests on Docker + made sure behavior was consistent with what was there prior.

## Types of changes

## Checklist:


classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -177,25 +169,29 @@ public static function registerAggregationUnits()
...
+     * @param $fact_table: The realm we are generating aggregation units for. Used in min/max calculations.
      *
      * @return class a subclass of TimeAggregationUnit based on $time_period requested.
      * @throws Exception if $time_period is not registered
      *
      * TimeAggregationUnit subclasses must be registed using TimeAggregationUnit::registerUnit first
      *
      */
-    public static function factory($time_period, $start_date, $end_date)
+    public static function factory($time_period, $start_date, $end_date, $fact_table)
```
smgallo: This is not the fact table, it is the aggregation table.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -71,43 +74,32 @@ public function getDateRangeIds($start, $end){
...
+
+        $unit = $this->getUnitName();
+
+        // this is needed because we can't be consistent about anything.
+        $unit_id = $unit . '_id';
+        $query = 'SELECT
+        MIN(' . $unit_id . ') as minPeriodId,
+        MAX(' . $unit_id . ')as maxPeriodId
+    FROM
+        ' . $this->getFactTable() . $unit . ' p;';
```
smgallo: You should check that the fact table has indeed been set since that is not guaranteed.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -12,6 +12,9 @@
     //The unit name is passed to this class via the constructor by extending subclasses
     private $_unit_name;
 
+    // This is begotten at query time from the factory method
+    private $_fact_table;
```
smgallo: This is not the actual table, this is the prefix.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -177,25 +169,29 @@ public static function registerAggregationUnits()
      * @param $time_period: the name of the time aggregation unit. ie: day, week, month, quarter.
      * @param $start_date: if time_period is auto this is used to figure out aggregation unit
      * @param $end_date: if time_period is auto this is used to figure out aggregation unit
+     * @param $fact_table: The realm we are generating aggregation units for. Used in min/max calculations.
```
smgallo: This comment is incorrect. This is not the realm.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -71,43 +74,32 @@ public function getDateRangeIds($start, $end){
...
-        if(null === $dateResult[0]['minPeriodId'] || null === $dateResult[0]['minPeriodId']){
-            return array(-1,-1);
+
+        $unit = $this->getUnitName();
+
+        // this is needed because we can't be consistent about anything.
+        $unit_id = $unit . '_id';
+        $query = 'SELECT
+        MIN(' . $unit_id . ') as minPeriodId,
+        MAX(' . $unit_id . ')as maxPeriodId
```
smgallo: add a space before "as"

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -71,43 +74,32 @@ public function getDateRangeIds($start, $end){
...
+    FROM
+        ' . $this->getFactTable() . $unit . ' p;';
+
+        $dateResult = \DataWarehouse::connect()->query($query, $queryParams);
```
smgallo: The query no longer uses `$queryParameters`, remove them here and clean them up above.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -71,43 +74,32 @@ public function getDateRangeIds($start, $end){
...
+
+        if (null === $dateResult[0]['minPeriodId'] || null === $dateResult[0]['minPeriodId']) {
```
smgallo: In what cases will `null` be returned?

classes/DataWarehouse/Query/TimeAggregationUnit.php
chakrabortyr: That is what the null statement below should nominally handle, however the setDuration function (where this is called) is only called after aggregation units have been initialized, which is (as far as I can tell) only after we have a database schema + table name.

classes/DataWarehouse/Query/TimeAggregationUnit.php
chakrabortyr: It nominally shouldn't unless something is wrong with the aggregation tables, however I left this in to make the code changes as minimal as possible.

classes/DataWarehouse/Query/TimeAggregationUnit.php
smgallo: I believe that is incorrect. If the fact table has not been set then this code will result in an error.

classes/DataWarehouse/Query/TimeAggregationUnit.php
smgallo: Add a comment saying "NULL will only be returned if the aggregation table is empty"

classes/DataWarehouse/Query/Query.php
```diff
@@ -103,7 +103,7 @@ public function __construct($realm_name,
         $this->setRealmName($realm_name);
         $this->setQueryGroupname($query_groupname);
 
-        $this->_aggregation_unit = \DataWarehouse\Query\TimeAggregationUnit::factory($aggregation_unit_name,$start_date, $end_date);
+        $this->_aggregation_unit = \DataWarehouse\Query\TimeAggregationUnit::factory($aggregation_unit_name,$start_date, $end_date, "{$datatable_schema}.{$datatable_name}_by_");
```
plessbd: if this is being used by SQL we should `quote` the schema and table names, not that this can easily be done here since this string gets concatenated.  So I would recommend passing them in separately so they can be properly quoted

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -177,25 +160,30 @@ public static function registerAggregationUnits()
...
+     *  ie "jobfact_by_"
      *
      * @return class a subclass of TimeAggregationUnit based on $time_period requested.
      * @throws Exception if $time_period is not registered
      *
      * TimeAggregationUnit subclasses must be registed using TimeAggregationUnit::registerUnit first
      *
      */
-    public static function factory($time_period, $start_date, $end_date)
+    public static function factory($time_period, $start_date, $end_date, $aggregate_table_prefix)
```
plessbd: this is not actually the table prefix, this is the `schema.table prefix`

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -12,6 +12,9 @@
     //The unit name is passed to this class via the constructor by extending subclasses
     private $_unit_name;
 
+    // This is begotten at query time from the factory method
```
smgallo: Fix this comment. It says nothing about what the variable holds.

classes/DataWarehouse/Query/TimeAggregationUnit.php
```diff
@@ -63,51 +66,32 @@ public function getUnitName()
...
+        $unit_id = $unit . '_id';
+        $query = 'SELECT
+        MIN(' . $unit_id . ') as minPeriodId,
+        MAX(' . $unit_id . ') as maxPeriodId
+    FROM
+        ' . $this->getAggTablePrefix() . $unit . ' p;';
+
+        $dateResult = \DataWarehouse::connect()->query($query);
+
+        // this should not occur unless something is wrong with the aggregate table(s)
```
smgallo: Specifically, if the aggregate table is empty. Please update.

smgallo:
You will also need to remove the ETL that creates the table and remove references in the XSEDE and other repos.

chakrabortyr:
> You will also need to remove the ETL that creates the table and remove references in the XSEDE and other repos.

Quoting the description: "If this doesn't have any hiccups, a subsequent pull request will eliminate the minmaxdate table from the ETL process."

Just wanted to make it a separate PR.

classes/DataWarehouse/Query/TimeAggregationUnit.php
chakrabortyr: yep, clarified.

PR #719, opened 2018-11-02
Add exception code to logAndThrowException

smgallo:
# Note: This is for 8.1

## Description

`Loggable:logAndThrowException()` includes a message but does not include an error code. If an exception was provided in the options list, use the error code from that exception. Note that PDOException uses a string exception code (typically a five characters alphanumeric identifier defined in the ANSI SQL standard) while Exception uses an int. For these cases, use the driver specific error code instead so that we can propagate the error code with the newly thrown Exception.

For a mapping of MySQL specific error codes to ANSI-SQL identifiers see https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html

## Motivation and Context

@eiffel777 can utilize this information to decide whether or not cloud data is present and should be ingested/aggregated.

## Tests performed

Manually ran tests.

## Types of changes

## Checklist:


smgallo:
@chakrabortyr This PR includes tests to ensure `logAndThrowException()` is behaving properly.

PR #727, opened 2018-11-10
fix defaults so the etl doesn't keep trying to alter the tables

plessbd:
the etlv2 process keeps seeing these two tables as needing to be altered, this should fix that.

### DIFF
--- a/configuration/etl/etl_tables.d/jobs/xdw/queue.json
+++ b/configuration/etl/etl_tables.d/jobs/xdw/queue.json
@@ -20,7 +20,6 @@
             {
                 "name": "queue_origin_id",
                 "type": "char(50)",
-                "default": "NULL",
                 "nullable": true
             }
         ],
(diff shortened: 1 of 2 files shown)


configuration/etl/etl_tables.d/jobs/xdw/queue.json
```diff
@@ -20,7 +20,6 @@
             {
                 "name": "queue_origin_id",
                 "type": "char(50)",
-                "default": "NULL",
```
smgallo: Use `"default": null`

plessbd: this line is getting deleted, are you saying you want me to keep it in with `"deafult": null`

smgallo: No, just wanted to point out that `"NULL"` is a string and `null` is a null in the database.

PR #728, opened 2018-11-10
updates that will hopefully make shippable run faster

plessbd:
Added in to docker
* npm cache
* chromedriver
  *  npmrc to use local file 
* mysql optimizations
* ssl keys generated once at docker build instead of for each samlsetup
Updated samlsetup and dependencies accordingly


plessbd:
well nope...

PR #730, opened 2018-11-16
Enable show raw data and Job Viewer for the Jobs realm

jpwhite4:
This pull request enables the Job Viewer functionality for the Jobs realm. This one is quite big, but mostly new code. I decided to include a minor internal API change (there will be a corresponding pull request for SUPREMM to update to use the new API). The API change is in classes/Rest/Controllers/WarehouseControllerProvider.php

integration tests have been added to test the new code.

Couple of things to think about:
- what access controls are appropriate for the jobs realm raw data? My implementation uses the same access controls as the SUPReMM realm based on the rationale that while aggregated jobs data is public, the details are not.
- In the current ETLv2 infrastucture forces me to create a seperate definition files for the _by_day stuff even though there is really not much differences. For example:
```diff
--- configuration/etl/etl_action_defs.d/jobs/hpc-aggregation-day.json	2018-11-16 11:08:43.000000000 -0500
+++ configuration/etl/etl_action_defs.d/jobs/hpc-aggregation.json	2018-11-16 10:42:34.000000000 -0500
@@ -1,7 +1,7 @@
 {
     "#": "Aggregation of HPC job records and tasks ingested from the XDCDB",
     "table_definition": {
-        "$ref": "${table_definition_dir}/jobs/xdw/jobfact_by_day.json#/table_definition"
+        "$ref": "${table_definition_dir}/jobs/xdw/jobfact_by_.json#/table_definition"
     },

     "#": "The aggregation period query determines which periods need to be aggregated based on added or modified",
@@ -77,8 +77,7 @@
             "node_time": "COALESCE(SUM(task.node_count * ${wallduration_case_statement}), 0)",
             "sum_node_time_squared": "COALESCE(SUM( CAST(POW(task.node_count * ${wallduration_case_statement}, 2) AS DECIMAL(36,4)) ), 0)",
             "sum_weighted_expansion_factor": "SUM( ((task.wallduration + task.waitduration) / task.wallduration) * task.node_count * COALESCE(${wallduration_case_statement}, 0))",
-            "sum_job_weights": "SUM(task.node_count * COALESCE(${wallduration_case_statement}, 0))",
-            "job_id_list": "GROUP_CONCAT(task.job_id)"
+            "sum_job_weights": "SUM(task.node_count * COALESCE(${wallduration_case_statement}, 0))"
         },
         "groupby": [
             "${AGGREGATION_UNIT}_id",
```
the `$ref` does not really improve the situation much because the bulk of the file is the column definintions.


smgallo:
For the duplication of the "by-day" file, do you mean that the bulk is in the `records` section of the `source_query`? Something to think about is perhaps being able to reference a JSON object and then add/override additional items.

jpwhite4:
Yes. I'd like to be able to reference the base file and then just insert an extra couple of columns to the table definition and add one line to the aggregation query.

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
```diff
@@ -13,6 +13,11 @@ cp -r $REF_SOURCE /var/tmp/
 set -e
 set -o pipefail
 
+if ! grep -q group_concat_max_len /etc/my.cnf.d/server.cnf;
```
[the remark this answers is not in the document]
plessbd: should be able to remove this now.

classes/DataWarehouse/Query/Jobs/JobMetadata.php
```diff
@@ -0,0 +1,66 @@
...
+        if ($job == null) {
+            return array();
+        }
+
+        return array(
+            \DataWarehouse\Query\RawQueryTypes::ACCOUNTING => true
+        );
+    }
+
+    public function getJobSummary($user, $jobid)
```
smgallo: Why do all of these methods return empty arrays? If there is a reason it should be documented in this class.

classes/DataWarehouse/Query/Query.php
```diff
@@ -725,6 +725,15 @@ public function getCountQueryString()
         $data_query .= ") as a WHERE a.total IS NOT NULL";
         return $data_query;
     }
+
+    protected function nextPdoIndex($value)
```
smgallo: Need documentation here as well.

classes/OpenXdmod/Migration/DatabasesMigration.php
```diff
@@ -150,4 +150,39 @@ protected function requestMysqlAdminCredentials()
...
+    /**
+     * Run an etlv2 action.
+     *
+     * This helper function provides the necessary boilerplate to execute
+     * an etlv2 pipeline.
+     *
+     * @param array $scriptOptions the options to pass to the etlv2 class
+     */
+    protected function runEtlv2(array $scriptOptions) {
+        if (empty($scriptOptions['chunk-size-days'])) {
```
smgallo: I've created a task in Asana to work on putting code like this into a central location  https://app.asana.com/0/806625283031961/917407032604343

open_xdmod/modules/xdmod/build.json
```diff
@@ -71,6 +71,7 @@
             "configuration/processor_buckets.json",
             "configuration/rawstatistics.json",
             "configuration/rawstatistics.d",
+            {"configuration/rawstatistics.d/z_jobs.json": "rawstatistics.d/z_jobs.json"},
```
smgallo: What does "z_jobs" stand for?

classes/DataWarehouse/Query/Jobs/JobMetadata.php
jpwhite4: Because there is no job summary data for the Jobs realm.

smgallo: OK, lets document that in the class.

open_xdmod/modules/xdmod/build.json
plessbd: not that it matters, but the convention i usually see for this is`\d+_\w`

PR #732, opened 2018-11-21
Update events that are used to determine VM session starts and stops

eiffel777:
There are more events that are VM sessions start and stop events than we previously realized. This PR adds 6 new events, 3 start events and 2 end events. The start events added are unsuspend, unpause and power_on. The end events added are power_off and pause. Resume events in openstack were are also mapped to the RESUME event.

During testing a bug was also found in the aggregation where if an instance had more that one session then the details of the first session was added to the cloud_events_transient table and any other sessions just updated the first row instead of adding a new one because of the keys on the cloud_events_transient table. To fix this I am adding the start_time and end_time field to primary key on that table.

## Motivation and Context
Adding the new stop and start events gives us more accuracy on how VM's are being used

## Tests performed
Updated regression tests and ran them. Also, manually checked to make sure the new aggregation was correct.

## Types of changes

## Checklist:


configuration/etl/etl_tables.d/cloud_common/cloud_transient.json
```diff
@@ -105,15 +105,19 @@
                 "name": "PRIMARY",
                 "columns": [
                     "resource_id",
-                    "instance"
+                    "instance",
+                    "start_time",
+                    "end_time"
```
jpwhite4: Why were these columns added?

jpwhite4: I.e. why were both start and end added? Surely the start_time is sufficient to generate uniqueness. If you include the end_time does this have a problem when the data are aggregated tomorrow (since the end time of an instance may have changed - but not the start time).

jpwhite4: I note that the cloud transient action that populates this table uses "resource_id ASC, instance_id ASC, start_time ASC" as the order by clause.

eiffel777: After thinking about it, I think that just using start time would be enough but as it is I don't think there will be an issue when you run aggregation the next day. The cloud_events_transient is truncated everytime aggregation is run so if you have a different end time for a session than the day before you will only get one row for that session with the new end time not two rows with the same start time but different end times.

I have no real strong feelings about this either way since I think both will work fine. If someone wants me to change it to just start_time I can do that.

jpwhite4: I thought the plan was to update the code to not truncate the tables every day. If you add the end_time in then it will just have to be removed again soon.

eiffel777: I know there was talk of making it so that you don't have to re-aggregate all the cloud data every time you run aggregation which would mean that the cloudfact tables would no longer be truncated everyday. I'm not sure what affect that would have on the cloud_events_transient table. If it's the case that it would no longer be truncated every time aggregation is run then the end_time would need to be removed.

jpwhite4: Of course the cloud_events_transient table would have to not be truncated - this table stores the facts. What design did you have in mind where the cloud_events_transient table could be truncated every day?

eiffel777: I don't have a design in mind. I don't think I'm the one that is going to make this change so I haven't really thought about it which is why I said I wasn't sure how any changes to how the aggregation is done would affect the cloud_events_transient table. Since it seems that we are going to change how the aggregation happens and that means the cloud_events_transient will no longer be truncated then I can remove the end_time.

classes/OpenXdmod/Migration/Version800To810/DatabasesMigration.php
```diff
@@ -2,9 +2,148 @@
...
+
+        $console = Console::factory();
+        $console->displayMessage(<<<"EOT"
+There have been updates to cloud aggregation statistics to make the data more accurate.
+IF you have the Cloud realm enabled it is recommended that you re-ingest and aggregate your cloud data.
+If you have a large amount of cloud data this may take multiple hours.
+EOT
+        );
+        $runaggregation = $console->promptBool(
+            'Do you want to run cloud aggregation now?',
```
plessbd: We should do some basic timing to allow them to know how long this might take.
This also runs ingestion.
We should check if the cloud is enabled before we bother prompting (already talked about that this ability is in another PR, so this can get updated at that point for all the things...)

classes/OpenXdmod/Migration/Version800To810/DatabasesMigration.php
```diff
@@ -2,9 +2,148 @@
...
+                    'The directory you specified does not seem to exist. Please enter the directory of your cloud logs to use for re-ingestion and aggregation of the Cloud realm.';
+
+                $cloudLogDirectory = $console->prompt(
+                    $cloudLogDirectoryPromptMessage,
+                    false
+                );
+                $cloudLogDirectoryEntryAttempts++;
+            }
+
+            $cloudResourceName = $console->prompt(
```
plessbd: Does this handle multiple cloud types?

Generic and openStack?

classes/OpenXdmod/DataWarehouseInitializer.php
```diff
@@ -353,7 +353,7 @@ public function aggregate(
      *
      * @param string $realm The realm you are checking to see if exists
      */
-    private function realmEnabled($realm)
+    public function realmEnabled($realm)
```
plessbd: I want to name this `isRealEnabled`
reasoning:
`isRealmEnabled` to me means it is going to tell you yes or no
vs `realEnabled` makes me think of things like [session_name](http://php.net/manual/en/function.session-name.php) which can be used to set things.

PR #739, opened 2018-11-30
Add cloud data to xdmod-shredder and xdmod-ingestor

eiffel777:
This PR adds support for shredding and ingesting cloud event logs in either the OpenStack or Eucalyptus(soon to be renamed Generic Cloud) log file formats. 

For shredding the cloud log files two shredder classes are added that implement the shredDirectory function that is used when xdmod-shredder is called using the -d option. Shredding cloud logs by file using the -i flag is not supported and throws an exception. 

For ingesting the shredded data a new command line option called datatype is added that allows you to specify what sort of data you want to ingest. Currently only openstack and genericcloud are supported. 

The aggregate option now takes an optional parameter to specify what data you want to aggregate. Currently this is either jobs or cloud. If no parameter is specified it will attempt to aggregate both jobs and cloud data. 

If no command line options are specified when running `xdmod-ingestor` then it will try to ingest both cloud and jobs data. There may be cases where only the jobs realm is being used but `xdmod-ingestor` with no command line options is run. In this example `xdmod-ingestor` will attempt to ingest jobs data and cloud data. In order to prevent errors from trying to ingest cloud data that doesn't exist the new function realmEnabled is used in order to only attempt ingestion of realms that are enabled.

There are some pipeline actions that needed to exist outside of the jobs pipelines becuase of foreign key constraints on tables used to ingest resources. This was done for actions pertaining to ingesting organizations and resource_types from their respective config files.

The runEtlPipeline function was added to /ETL/Utilities class. It is similar to the runEtlPipeline function in DataWarehouseInitializer but is not private and does not require you to instantiate the DataWarehouseInitializer class and the two other classes that are passed to the DataWarehouseInitializer constructor but not used in runEtlPipeline. In a subsequent pull request I will remove the runEtlPipeline function from the DataWarehouseInitializer class and point anything that uses it to the ETL\Utilities::runEtlPipeline().

## Motivation and Context
We want to have a unified way of shredding and ingesting all types of data that xdmod needs.

## Tests performed
Manual testing in docker and added new tests for shredder classes that test to make sure if a directory that does not exist is passed to the shredder then false is returned.

## Types of changes

## Checklist:


bin/xdmod-ingestor
```diff
@@ -108,6 +111,7 @@ function main()
                 break;
             case 'aggregate':
                 $aggregate = true;
+                $realmToAggregate = $value;
```
jpwhite4: According to the docs the $value for an optional argument is false if it is not specified. Assuming the documentation is correct you'll need to check for === false rather than === null on line 287. Please confirm.

eiffel777: @jpwhite4 Yeah, that does seem to be the case. It should be changed now.

bin/xdmod-ingestor
```diff
@@ -57,6 +57,9 @@ function main()
         array('', 'ingest-staging'),
         array('', 'ingest-hpcdb'),
 
+        //Type of data that is being ingested.
```
plessbd: nitpicking, but I like a space after the `//` also matches the others in the file

bin/xdmod-ingestor
plessbd: do we need both `aqggregate` and `realmToAggregate`

bin/xdmod-ingestor
```diff
@@ -130,6 +134,10 @@ function main()
             case 'ingest-hpcdb':
                 $ingestHpcdb = true;
                 break;
+            case 'datatype':
+                $datatype = true;
```
plessbd: do we need both `datatype` and `datatypeValue`

bin/xdmod-shredder
```diff
@@ -233,7 +233,7 @@ function main()
         $logger->notice("Job errors written to '$jobErrorLogFile'");
     }
 
-    if (!$dryRun) {
+    if (!$dryRun && in_array($format, array('pbs', 'slurm', 'lsf', 'sge', 'uge'))) {
```
jpwhite4: Please can you explain the reason for this change.

classes/ETL/Utilities.php
```diff
@@ -323,4 +323,54 @@ public static function quoteVariables(array $variables, VariableStore $variableS
 
         return $localVariableMap;
     }  // quoteVariables()
+
+    public static function runEtlPipeline(array $pipelines, $logger, array $params = array())
```
plessbd: It feels like this should be a separate pull request that also removes similar code from the code base:

https://github.com/ubccr/xdmod/blob/6149358969bac0ec16250a529b6a9990505fa662/classes/OpenXdmod/Migration/Etlv2Migration.php#L54

https://github.com/ubccr/xdmod/blob/6149358969bac0ec16250a529b6a9990505fa662/classes/OpenXdmod/DataWarehouseInitializer.php#L322

https://github.com/ubccr/xdmod/blob/133a3ae3047d465323e43a5330ab46314bba678b/classes/OpenXdmod/Setup/DatabaseSetup.php#L163

https://github.com/ubccr/xdmod/blob/133a3ae3047d465323e43a5330ab46314bba678b/bin/acl-config#L2251

Specifically so we dont have the many places this is done in our code base and we dont forget about it.

eiffel777: Here is the branch, https://github.com/eiffel777/xdmod/tree/move-runEtlPipeline-references-to-etl-utilities-class, that has the changes talked about in the last paragraph of the description. It's a bit behind xdmod8.1 right now so there is a some extra code in the diff that can be ignored

classes/OpenXdmod/Shredder/Openstack.php
```diff
@@ -7,14 +7,14 @@
 
 namespace OpenXdmod\Shredder;
 
+use CCR\DB\iDatabase;
+
 class OpenStack extends aCloud
```
plessbd: Should the filename match the class name `OpenStack` vs `Openstack`

eiffel777: Thanks for the catch. It's changed now.

classes/OpenXdmod/DataWarehouseInitializer.php
```diff
@@ -275,6 +348,17 @@ public function aggregate(
...
 
+    /**
+     * Check to see if a realm exists in the realms table
+     *
+     * @param string $realm The realm you are checking to see if exists
+     */
+    private function realmEnabled($realm)
+    {
+        $realms = $this->warehouseDb->query("SELECT * FROM moddb.realms WHERE display = :realm", [':realm' => $realm]);
+        return (count($realms) > 0) ? true : false;
```
plessbd:
```suggestion
         return (count($realms) > 0);
```

configuration/etl/etl.d/jobs_cloud_eucalyptus.json
```diff
@@ -20,7 +20,13 @@
...
         },
 
-        "jobs-cloud-eucalyptus": {
+        "jobs-cloud-ingest-eucalyptus": {
+            "namespace": "ETL\\Ingestor",
+            "options_class": "IngestorOptions",
+            "truncate_destination": false,
+            "enabled": true
+        },
+        "jobs-cloud-extract-eucalyptus": {
```
plessbd: should this be `eucalyptus` or should it be `generic` since we are starting to call them all generic instead of euca I would prefer we get all of them to be that way.

eiffel777: @plessbd I've already done some work on changing them over to be generic on this branch, https://github.com/eiffel777/xdmod/tree/change-eucalyptus-to-generic-cloud. Once the two pull requests I have up now are approved I will put a PR in for that branch.

bin/xdmod-shredder
```diff
@@ -234,11 +234,16 @@ function main()
     }
 
     if (!$dryRun) {
-        $logger->notice('Normalizing data');
+        $logger->notice('Normalizing data!');
```
plessbd: I am so excited we are normalizing this data, before I was just fine.

PR #746, opened 2018-12-12
Remove UserRoleParameters & Simplify user_acl_group_by_parameter usage

ryanrath:
## Description
This PR is a part of the role migration process. In it I have gone through and removed all usage of both `user_acl_group_by_parameters` && `UserRoleParameters`. Both direct usage ( via a sql query ) and indirect usage ( calling a function that relies on querying either of the tables ) were removed. Where appropriate I have also removed code that was no longer being used. 

## Motivation and Context
Now that we directly associate Users w/ an organization these two tables are no longer necessary. This allows a fairly large chunk of code to be removed ( note PR addition / removal stats of 331 & 3192 respectively ) and what remains to be simplified. 

## Tests performed
- All automated tests were performed in the OpenXDMoD docker container w/ both a fresh install && and upgrade. 
- Manual tests were performed to ensure that the Center Director promotion / demotion feature still works as expected. 
- SSOLogin tests were updated so that the automated organization updating could be tested.
- New tests were not added as the existing test suite exercises what changes were introduced. 

## Types of changes

- [X] Maintenance / Refactoring

## Checklist:


jpwhite4:
I don't think we should remove the ability to associate parameters with a given ACL. This is a useful feature that has direct application: for example a 'domain champion' in XSEDE would want to be able to see all jobs in the given field of science.

html/controllers/role_manager/downgrade_member.php
```diff
@@ -14,16 +14,18 @@
 try {
 
     $activeUser = \xd_security\getLoggedInUser();
-    $organization = $activeUser->getActiveOrganization();
-    $memberUserId = $member->getUserID();
+    $organization = $activeUser->getOrganizationID();
 
     // An eligible user must be associated with the currently logged in users center.
-    if (!Users::userIsAssociatedWithCenter($memberUserId, $organization)) {
+    if ($activeUser->getOrganizationID() !== $organization) {
```
jpwhite4: The value of `$organization` is set on line 17 to be the result of a call to `$activeUser->getOrganizationID()`. Therefore this if statement can never return false and is not needed at all.

ryanrath: It is, it just should have been `if ($activeUser->getOrganizationID() !== $member->getOrganizationID()) {`. That was a flub on my part. I'll be committing that change + re-instating uagbp table & infrastructure shortly.

ryanrath:
Per conversations w/ @smgallo / @jpwhite4 I'm going to close this PR and push a new one that leaves user_acl_group_by_parameters in place in addition to a re-working of the commits to make it easier to review.

PR #749, opened 2018-12-14
Make logrotate.d file not override global settings

mjuwit:
Moved common logrotate parameters inside of the brackets for the various logfiles.

## Description
logrotate.d confs are loaded via "dumb" include statement in the logrotate.conf which "Reads the file given as an argument as if it was included inline where the include directive appears." (from logrotate manpage).

Previous version was applying the 'rotate', 'weekly', 'compress', 'missingok', and 'datetext' parameters outside of the mustache brackets for the individual log files.  This would override any settings made in the main system 'logrotate.conf' file before the include statement, thus potentially altering global logrotate behavior for the system.

## Motivation and Context
xdmod shouldn't alter global logrotate.conf parameters (thus potentially altering logrotate behavior for the entire system without knowledge of the system administrators).

## Tests performed
Running this version on our sytsems -- had to change the file as it was altering our global logrotate configuration.

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


jtpalmer:
Thank you for bringing this issue to our attention.  I didn't realize those directives would override other files.

PR #757, opened 2018-12-21
Update to support simplesaml session naming

plessbd:
SimpleSAMLphp uses its own session name to store its information.
This fix makes it so that this actually works and keeps XDMoD working.

Also I hate waiting 3 seconds, I changed it to that when initially doing this to make testing easier, this sets it back down so that we aren't just sitting to sit.


classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -84,10 +93,20 @@ public function __construct()
      */
     public function isSamlConfigured()
     {
-        $this->_isConfigured = count($this->_sources) > 0 ? true : false;
+        if($this->_isConfigured === null){
```
chakrabortyr: why did this change?

classes/XDSessionManager.php
```diff
@@ -248,6 +248,13 @@ public static function logoutUser($token = "")
         // authentication (via tokens) trip the first Exception as the
         // result of invoking resolveUserFromToken($token)
         session_destroy();
+
+        try {
+            $auth = new Authentication\SAML\XDSamlAuthentication();
+            $auth->logout();
+        } catch (InvalidArgumentException $ex) {
+         // This will catch when a configuration directory does not exist if it is set in the environment level
```
chakrabortyr: Shouldn't we log something here? Empty catch statements hurt my soul.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -84,10 +93,20 @@ public function __construct()
...
+            $this->_isConfigured = (count($this->_sources));
```
chakrabortyr: Couldn't this just be:

```suggestion
            return count($this->_sources);
```

And then we can get rid of the variable.

classes/Authentication/SAML/XDSamlAuthentication.php
plessbd: So that the call doesnt have to do the count if it has already done it once.   If sources is a LARGE set of providers only iterate over it once.

classes/XDSessionManager.php
plessbd: Nope, this is specifically when apache or nginx have been set up to to have an alternate saml config directory and that directory does not exist, so we ignore it as saml isnt set up and we dont have to do anything with it.  This is the same as what is being done in login.php

classes/Authentication/SAML/XDSamlAuthentication.php
plessbd: answered in other comment

classes/XDSessionManager.php
smgallo: Lets add "this is specifically when apache or nginx have been set up to to have an alternate saml config directory and that directory does not exist, so we ignore it as saml isnt set up and we dont have to do anything with it" to the comment.

classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -84,10 +93,20 @@ public function __construct()
      */
     public function isSamlConfigured()
```
jpwhite4: How about:
```php
pubilc function isSamlConfigured()
{
    return !empty($this->_sources);
}
```

docs/simpleSAMLphp.md
```diff
@@ -47,10 +47,8 @@ You will need to modify the `config.php` file and make sure you modify the `meta
   ),
   ...
 ```
-two other keys that might need to be set if you are having errors
+If you are having errors you might need to check the trusted domains setting
```
jpwhite4: How does the admin know if they need to check the settings? Please enumerate the conditions when they might need to and might not need to check them.

plessbd: If you are having errors

We had at least one person that had to change the `trusted domains setting`, I never figured out exactly what their setup was that made them have to do this.  But since I had run into it once, I put it in there.

PR #762, opened 2019-01-02
Change eucalyptus to generic cloud

eiffel777:
Changing the name of the Eucalyptus etl pipelines to Generic since these pipelines can be used of systems other than just Eucalyptus and is the one we tell users to use if the don't have an OpenStack system.

## Tests performed
Ran all test in docker and manually tested ingestion and aggregation

## Types of changes

## Checklist:


configuration/etl/etl_action_defs.d/cloud_common/cloud_metrics_aggregation.json
```diff
@@ -1,5 +1,5 @@
 {
-    "#": "Aggregation of Cloud Data ingested from Eucalyptus",
+    "#": "Aggregation of Cloud Data ingested from OpenStack and generic cloud logs",
```
jpwhite4: This pipeline is agnostic to the source of the data so there is not need to explicitly list the currently supported data sources here.

configuration/etl/etl.d/jobs_cloud_generic.json
```diff
@@ -30,7 +36,7 @@
 
     "#": "Current Cloud job ingestion",
 
-    "jobs-cloud-generic": [
+    "jobs-cloud-ingest-generic": [
```
smgallo: Can we add comments in the file so it's clearer what the difference between the "jobs-cloud-ingest-generic" and "jobs-cloud-extract-generic" pipelines are doing?

eiffel777: @smgallo I just added some comments to clarify the difference between the pipelines.

PR #765, opened 2019-01-03
Fix change tracking while editing existing users

ryanrath:
## Description
Previously, the method used to determine whether or not changes had been made
while editing an existing user was manipulation of a "global" ( to
SectionExistingUsers ) variable, `settingsAreDirty`. This caused problems when
reverting a previously updated value to it's original value, after which the UI
still indicated that changes had been made.

To resolve this problem we have replaced the `settingsAreDirty` boolean with
a call to a new function `userSettingsDirty` in `inDirtyState`.
`userSettingsDirty` returns true based on whether any of the User Setting fields
has a value different from their original value.

To finish off the change, each location that set `settingsAreDirty` has been
replaced with a call to `updateSaveIndicator`. Which just shows or hides the
save indicator based on `inDirtyState`.

## Motivation and Context
Intuitive UI is good UI.

## Tests performed
Manual testing was performed for both a fresh_install && upgrade

## Types of changes

## Checklist:


html/controllers/user_admin/create_user.php
```diff
@@ -56,6 +56,8 @@ function ($carry, Acl $item) {
     \xd_response\presentError("Acl information is required");
 }
 
+$sticky = isset($_POST['sticky']) ? (bool)$_POST['sticky'] : false;
```
jpwhite4: The $_POST['sticky'] will come in as a string that is set to 'true' or 'false' the `(bool)` cast operator will not convert this to a php boolean. You could use `filter_var($_POST['sticky'], FILTER_VALIDATE_BOOLEAN)` in place of the bool cast. 

Ref: http://php.net/manual/en/language.types.boolean.php

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -201,14 +199,16 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
 
         var comboChangeHandler = function (f, newValue, oldValue) {
             /* eslint-disable no-use-before-define */
-            settingsAreDirty = true;
-            saveIndicator.show();
-            if (cmbUserMapping.startValue !== cmbUserMapping.getValue() || cmbInstitution.startValue !== cmbInstitution.getValue()) {
+            if ((!Ext.isEmpty(cmbUserMapping.originalValue) && String(cmbUserMapping.originalValue) !== String(cmbUserMapping.getValue())) ||
+                (!Ext.isEmpty(cmbInstitution.originalValue) && String(cmbInstitution.originalValue) !== String(cmbInstitution.getValue()))
```
jpwhite4: Why the String cast here?

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -704,28 +705,60 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
...
+                        if (Ext.isEmpty(field.originalValue)) {
+                            if (!Ext.isEmpty(field.startValue)) {
+                                field.originalValue = field.startValue;
+                            } else if (!Ext.isEmpty(value)) {
+                                field.originalValue = value;
+                            }
+                        }
+
+                        // We explicitly cast each side to string so that we can
+                        // utilize the more stringent `===`.
```
jpwhite4: Why? Under what circumstances would the data types be different? What data types are being set in these variables?

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -1279,7 +1313,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
                                 if (resp === 'no') {
                                     self.resetDirtyState();
                                     // eslint-disable-next-line no-use-before-define
```
jpwhite4: This eslint disable directive is not needed. Remove it

html/internal_dashboard/js/admin_panel/SectionNewUser.js
```diff
@@ -551,7 +551,7 @@ XDMoD.CreateUser = Ext.extend(Ext.form.FormPanel, {
             handler: function () {
                 var institution = (cmbInstitution.getValue().length > 0) ? cmbInstitution.getValue() : -1;
                 var mapped_user_id = (cmbUserMapping.getValue().length > 0) ? cmbUserMapping.getValue() : -1;
-
+                var manualOverride = mapped_user_id === -1 && institution !== -1;
```
jpwhite4: Please can you explain this logic. It looks to me that manualOverride will always be set to false since the mapped_user_id can never be -1 because the user must be mapped before this function is called.

jpwhite4:
When I try to view an existing user (wiith this code) the dialog box incorrectly shows: There are unsaved changes to this account.

Steps:
-Open Internal Dashboard
-Select User Management Tab
-Click "Create and Manage Users" button
-Click "Current Users" tab in dialog
-Select "External Users" in filter drop down
-Click on some of the users in turn.

jpwhite4:
Also the "No discard changes" button is borked too. You click it and nothing happens. I assume this is related to the change tracking errors.

html/internal_dashboard/js/admin_panel/SectionNewUser.js
ryanrath: A mapped_user_id of -1 is the 'Unknown' person. So we set manualOverride to true if the user has selected the 'Unknown' person but has selected something other than the default 'Unknown' organization.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
ryanrath: They're different when the field being inspected is the User Type field as it has a starting value of 1 but when `getValue` is called it returns a string.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
ryanrath: Thanks, removed in latest commit.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
ryanrath: This has been removed in the latest commit.

html/controllers/user_admin/create_user.php
ryanrath: sweet thanks! Updated in latest commit.

ryanrath:
@jpwhite4 I've addressed your questions / comments in the latest commits. If you could give these changes a once over when you get a moment that would be great. Thanks!

html/internal_dashboard/js/admin_panel/SectionNewUser.js
jpwhite4: The user interface does not allow the Unknown person to be set and does not allow the organisation to be overridden. How did you test this code change to ensure that the manualOverride flag was set correctly?

ryanrath: Could I get a set of steps to reproduce? When I create a new user via "Create & Manage Users" -> "New User" I am definitely able to select the 'Unknown' user. I actually wrote tests to ensure this works in `automated_tests/test/specs/xdmod/internalDashboard.js`.

jpwhite4: Ah. I see what you were getting at. I was referring to the code path scenario:

line 553 when the cmbUserMapping.getValue().length returns 0 then the mapped_user_id variable will be set to -1.

ryanrath: Oooo ok, thanks! I've gone back over that bit of code and I'm currently going to be testing the following: 

Removing SectionNewUser.js@552-524 and replacing it with:
```js 
                var mappedUser = parseInt(cmbUserMapping.getValue());
                if (!Number.isInteger(mappedUser)) {
                    CCR.xdmod.ui.userManagementMessage('You must select a valid Person for this user to be mapped to.');
                    return;
                }

                var institution = parseInt(cmbInstitution.getValue());
                if (!Number.isInteger(institution)) {
                    CCR.xdmod.ui.userManagementMessage('You must select a valid Institution for this uesr to be mapped to.');
                    return;
                }

                var manualOverride = mappedUser === -1 && institution !== -1;
```

ryanrath: @jpwhite4 Here's the commit where the changes were made: https://github.com/ubccr/xdmod/pull/765/commits/a9dad208b73ad3d9f620a06dd1382cdba8e5f4af

jpwhite4:
As of commit a61c6e5 toggling the ACL assignment checkboxes does not update the 'unsaved changes' text properly.

jpwhite4:
Also the unsaved changes message is not reset properly when you switch users.

ryanrath:
Hi Joe, 

Commit https://github.com/ubccr/xdmod/pull/765/commits/6a662fd3dd5bb2c96d53316899fce0f19881806f Takes care of the acl assignment checkbox problem.

Commit https://github.com/ubccr/xdmod/pull/765/commits/4d37caf9b531e55600090f3ac4883f51c91a27c3 Takes care of the a problem with consecutive changes to User Type not being flagged properly.

Commit https://github.com/ubccr/xdmod/pull/765/commits/4cf69cff3cca0f7caa1ec7ee8fcd57f74c6de131 Takes care of the message not being updated upon first selecting a new user.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -283,7 +297,7 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
                             self.resetDirtyState();
 
                             // eslint-disable-next-line no-use-before-define
```
jpwhite4: remote this eslint disable - it is no longer applicable.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -668,10 +686,8 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
                         return;
                     }
 
-                    settingsAreDirty = true;
-
                     // eslint-disable-next-line no-use-before-define
```
jpwhite4: remove this eslint disable - it is no longer applicable.

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -116,12 +116,24 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
...
-
         self.setCallback = function (callback) {
             user_update_callback = callback;
         };
 
+        /**
+         * Updates the save indicator based on whether or not changes have been
+         * made.
+         */
+        self.updateSaveIndicator = function () {
```
jpwhite4: What is the reason for assigning this function as a property of the Panel? I would have thought that, since it is effectively a private function it should just be a variable.
```js
var updateSaveIndicator = function () {
...
```

html/internal_dashboard/js/admin_panel/SectionExistingUsers.js
```diff
@@ -1174,11 +1226,16 @@ XDMoD.ExistingUsers = Ext.extend(Ext.Panel, {
                             mnuItemPasswordReset.show();
 
                             cmbUserType.setDisabled(false);
+                            cached_user_type = String(json.user_information.user_type);
                             cmbUserType.setValue(cached_user_type);
+                            cmbUserType.originalValue = cmbUserType.getValue();
                         }
-                        stickyCheckbox.setValue(Boolean(json.user_information.sticky));
-                        cmbInstitution.initializeWithValue(json.user_information.institution, json.user_information.institution_name);
+                        var sticky = Boolean(json.user_information.sticky);
```
jpwhite4: out of interest. Why the Boolean cast here? Isn't the json.user_information.sticky property always going to be a json boolean?:q

ryanrath: Yeah... I think this was left over from when I was debugging while adding the `sticky` property. I was running into some problems w/ the `sticky` value coming back as an int.

PR #768, opened 2019-01-04
Don't automatically rewind DirectoryScanner file handle

smgallo:
## Description

When running the DirectoryScanner ETL DataEndpoint in docker using the latest overlay2 storage driver and PHP 5.4, the `DirectoryScannerTest::testPatternFilters()` automated tests would fail because an incorrect number of files (exactly 2 times the expected number) was returned from a directory scan. When using the aufs driver (as shippable.com does) this did not manifest.

The error occurs when the `CallbackFilterIterator::rewind()` method is called twice in a row. The set of records returned by the iterator is duplicated. Since we iterate over the set using `foreach` and a `rewind()` is called under the hood by PHP it is not necessary to explicitly call it ourselves.

**Note that this bug does not manifest in PHP7 running in docker.**

## Motivation and Context

Rescue developers from insanity wondering why tests that have been working for years are now failing.

## Tests performed

Manual docker testing and automated testing in docker.

## Types of changes

## Checklist:

### DIFF
--- a/classes/ETL/DataEndpoint/DirectoryScanner.php
+++ b/classes/ETL/DataEndpoint/DirectoryScanner.php
@@ -611,9 +611,6 @@ class DirectoryScanner extends aDataEndpoint implements iStructuredFile, iComple
 
         $this->handle = $iterator;
 
-        // Rewind the handle so it is ready to use.
-        $this->handle->rewind();
-
         return $this->handle;
 
     }  // connect()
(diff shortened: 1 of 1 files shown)


smgallo:
Even if I run the integration tests locally against the v2 docker using the unmodified 8.1 branch I get failures. Either I have something screwed up or something else is wonky.

PR #782, opened 2019-01-23
Updating Configuration class to support merging objects w/ Tests

ryanrath:
## Description
- Updated `Configuration\Configuration` w/ the ability to merge objects. This is
  so that `Configuration\Configuration` can eventually replace `Config`.
- Added tests to `EtlConfigurationTest` that ensure
  `Configuration\Configuration` can generate the same json as `Config`.
  - The reason these tests have been placed here instead of `ConfigurationTest`
    is documented here:
    https://app.asana.com/0/807629084565719/1101232922862525/f
- Updated the unit tests `bootstrap.php` so that we can use the
  integration_tests `TestFiles` helper class.

## Motivation and Context
`Config` currently supports a couple of features that `Configuration\Configuration` does not. 
- Merging .d configuration files that contain '+' prefixed property names
- Handling the 'extends' key / value pair
- Retrieving which configuration options have been provided by which module.

This PR implements the first of these features ( Merging '+' prefixed properties ). Future PR's will address the remaining features. 

## Tests performed
All existing tests were performed:
- unit
- component
- integration
- regression
- UI

New tests were added to ensure, on a basic level, that `Configuration` can generate the same output as `Config` when presented with the same base configuration files. In this case both `roles.json` and `datawarehouse.json` are tested.

## Types of changes

## Checklist:


classes/Configuration/Configuration.php
```diff
@@ -515,6 +541,41 @@ protected function merge(Configuration $localConfigObj, $overwrite = false)
...
+                $existing->$property = $incoming->$property;
+            } else {
+                $existingValue = $existing->$property;
+
+                if (is_object($existingValue) && is_object($incomingValue)) {
+                    $existing->$property = $this->mergeObjects($existingValue, $incomingValue);
+                } elseif (is_array($existingValue) && is_array($incomingValue)) {
+                    $existing->$property = array_merge($existingValue, $incomingValue);
+                } else {
+                    $existing->$property = $incomingValue;
```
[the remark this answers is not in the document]
smgallo: Actually, if we fix up this method can't we just call it from the `merge()` method and get rid of the if-then-else there?

classes/Configuration/Configuration.php
```diff
@@ -490,12 +490,38 @@ protected function merge(Configuration $localConfigObj, $overwrite = false)
         // it.
 
         foreach ( $localConfigObj->getTransformedConfig() as $k => $v ) {
+
+            // Normalize incoming keys starting w/ '+'
+            if (substr($k, 0, 1) === '+') {
```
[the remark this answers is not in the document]
ryanrath: \*facepalm\* yeah, completely forgot about that part of the convo. I'll add the transformer and push it up shortly.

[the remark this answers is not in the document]
smgallo: You should be able to use the CommentTransformer with minimal changes to suit this.

classes/Configuration/PlusKeyTransformer.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
```
smgallo: We need a docblock here explaining what the plus key represents and how we're transforming it (e.g., why it is being removed - because merge is the default behavior of Configuration).

classes/Configuration/Configuration.php
```diff
@@ -390,6 +390,7 @@ protected function preTransformTasks()
     {
         $this->addKeyTransformer(new CommentTransformer($this->logger));
         $this->addKeyTransformer(new JsonReferenceTransformer($this->logger));
+        $this->addKeyTransformer(new PlusKeyTransformer($this->logger));
```
smgallo: Is there a better name for this transformer that hints at its function?

ryanrath: lol I'm sure there is, I honestly wasn't sure what to go with. Maybe, `MergeKeyTransformer` or something similar?

smgallo: Maybe `StripMergePrefixTransformer` ?

ryanrath: That sounds good. I'll update the name + add the docblock in the next commit.

PR #789, opened 2019-02-01
Configuration Code - Consolidation

ryanrath:
## Description
- Added support for merging `+` prefixed configuration keys.
- Added support for merging based on the `extends` configuration key.
- Added tests to exercise this new functionality.

## Motivation and Context
This PR brings us closer to feature parity w/ the `Config` class and thus closer to finishing the consolidation of XDMoD's code that deals with configuration files. The only feature left to implement is being able to tell which module has supplied which portions of a configuration unit. 

## Tests performed
For docker ( fresh install / update ):
- Unit
- Component 
- Integration
- Regression ( REG_TEST_ALL=1)
- UI

## Types of changes

## Checklist:


classes/Configuration/Configuration.php
```diff
@@ -560,10 +583,149 @@ protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomi
...
+     * NOTE: After an `extends` property has been successfully merged, it will itself be unset and
+     * therefor not present in the resultant object.
+     *
+     * @param stdClass $config    the base object that will be recursively traversed to find and
+     * resolve instances of the `extends` property.
+     * @param string[] $extenders the array of property keys whose values define an `extends`
+     * property.
+     * @throws Exception if unable to resolve the merging of an extender and it's extendee.
+     */
+    protected function resolveExtends(\stdClass &$config, array $extenders)
```
smgallo: Objects are passed by reference so you don't need the `&` in `\stdClass &$config`

classes/Configuration/Configuration.php
```diff
@@ -560,10 +583,149 @@ protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomi
...
+    /**
+     * Find $key in this `Configuration`s `transformedConfig` and return it's value. If not found,
+     * null will be returned.
+     *
+     * @param string        $key      the key being looked for in $incoming ( or `transformedConfig`
+     *                                if $incoming is not set. )
+     * @param stdClass|null $incoming the stdClass to be searched for key.
+     * @return mixed|null `$key`s value if found, else null.
+     */
+    protected function findByKey($key, $incoming = null)
```
smgallo: What happens if there are multiple keys with the same name in the transformed config? For example:
```
{
    default: {
        some: thing,
        usr: smgallo
    },
    usr: {
        ...
    }
}
````

File: classes/Configuration/Configuration.php [hunk 3]
```diff
@@ -560,10 +583,149 @@ protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomi
...
+     * property.
+     * @throws Exception if unable to resolve the merging of an extender and it's extendee.
+     */
+    protected function resolveExtends(\stdClass &$config, array $extenders)
+    {
+        foreach($config as $k => $v) {
+            if (in_array($k, $extenders)) {
+                $target = $this->findByKey($v->extends);
+                if ($target === null) {
+                    $this->logAndThrowException("Unable to find 'extends' target $v");
```
smgallo: This won't represent well if `$v` is not a scalar. You'll get "Unable .... (array))".

classes/Configuration/Configuration.php
```diff
@@ -560,10 +583,149 @@ protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomi
...
+                }
+                $config->$k = $this->mergeLocal($v, $target, "extending-{$v->extends}", false, false);
```
smgallo: `"extending-{$v->extends}"` isn't an actual filename is it?

ryanrath: Nope, just a unique-ish identifier to be used if an error occurs.

classes/Configuration/Configuration.php
ryanrath: Updated in the latest version ( in `XdmodConfiguration` now ) to use sprintf + json_encode to make for a more understandable exception message.

classes/Configuration/Configuration.php
```diff
@@ -560,10 +583,149 @@ protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomi
...
+     * @return array in the form:
+     * array(
+     *   'extender' => 'extendee'
+     * )
+     */
+    protected function findExtends(\stdClass $config, $parentKey = null)
+    {
+        $extends = array();
+        foreach($config as $k => $v) {
+            if ($k === 'extends' && is_string($v)) {
```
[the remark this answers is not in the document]
ryanrath: Yup, this has been moved to a new child class called `XdmodConfiguration`. I've pulled the `'extends'` out into a class constant so if we want to change it to '$extends' it will be really easy.

classes/Configuration/Configuration.php
```diff
@@ -512,27 +512,50 @@ protected function merge(Configuration $localConfigObj, $overwrite = false)
...
-    protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomingFileName, $overwrite)
+    protected function mergeLocal(\stdClass &$existing, \stdClass $incoming, $incomingFileName, $overwrite = false, $overwriteScalar = true)
     {
         foreach($incoming as $property => $incomingValue) {
 
             if ( $overwrite || ! isset($existing->$property) ) {
                 $existing->$property = $incoming->$property;
             } else {
-                $existingValue = $existing->$property;
+                $existingValue = &$existing->$property;
```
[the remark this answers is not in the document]
ryanrath: So that changes to `$existingValue` will persist. This occurs when both `$existingValue` and `$incomingValue` are arrays.

bin/acl-config
```diff
@@ -128,10 +128,18 @@ function main()
...
+        $log,
+        array(
+                'local_config_dir' => CONFIG_DIR . '/modules.d'
+        )
+    );
+    $moduleConfig->initialize();
 
     // Module Retrieval
-    $modules = $config['modules'];
+    $modules = json_decode(json_encode($moduleConfig->getTransformedConfig()), true);
```
smgallo: Lets add a couple words as to why we need to encode/decode the config.

bin/acl-config
```diff
@@ -128,10 +128,18 @@ function main()
     // roles/datawarehouse
     $results = array();
 
-    $config = Config::factory();
+    $moduleConfig = new \Configuration\Configuration(
```
smgallo: This is using `\Configuration\Configuration` to set `$moduleConfig` but there is a `\Configuration\ModuleConfiguration` below. Why isn't that used here?

bin/acl-config
```diff
@@ -141,86 +149,111 @@ function main()
...
                 }
             }
         }
 
         // Process the realms provided by this module, if any.
-        if (count($realms) > 0 && isset($realms['realms'])) {
-            foreach ($realms['realms'] as $realm => $data) {
-                $groupBys = isset($data['group_bys']) ? $data['group_bys'] : null;
-                $statistics = isset($data['statistics']) ? $data['statistics'] : null;
+        if (count(get_object_vars($realms)) > 0 && property_exists($realms, 'realms')) {
```
smgallo: Do we need `count(get_object_vars($realms)) > 0`?

classes/Configuration/ModuleTransformer.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
```
smgallo: We need a docblock here explaining what this transformer does.

classes/Configuration/XdmodConfiguration.php
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * This function will handle the `extends` keyword for the provided $source object. This will
+     * result in merging properties from the `extends` target into the object that defined the
+     * `extends` property. After the processing is done the `extends` property will be removed.
+     *
+     * @param $source
+     * @throws Exception
+     */
+    protected function handleExtendsFor($source)
```
smgallo: Will `$source` always be a `stdClass`? If so, lets use type hinting.

classes/Configuration/XdmodConfiguration.php
```diff
@@ -0,0 +1,224 @@
...
+     * array(
+     *   'extender' => 'extendee'
+     * )
+     */
+    protected function findExtends(\stdClass $config, $parentKey = null)
+    {
+        $extends = array();
+        foreach ($config as $k => $v) {
+            if ($k === self::EXTENDS_KEYWORD && is_string($v)) {
+                $extends[$parentKey] = $v;
```
smgallo: What happens when we set `$extends[null] = $v`?

classes/Configuration/ModuleConfiguration.php
```diff
@@ -0,0 +1,461 @@
...
+     *
+     * @param Configuration $localConfigObj
+     * @param bool $overwrite
+     * @return Configuration
+     */
+    protected function merge(Configuration $localConfigObj, $overwrite = false)
+    {
+        parent::merge($localConfigObj, $overwrite);
+
+        if ($localConfigObj instanceof ModuleConfiguration) {
```
smgallo: The code here looks identical to `Configuration::merge()` except for this conditional. Is it necessary to call the parent?

bin/acl-config
ryanrath: Heh, this is a little confusing given the naming. I'll try to explain: So `$moduleConfig` describes the data ultimately being retrieved. We use `\Configuration\Configuration` as opposed to `\Configuration\ModuleConfiguration` because `\Configuration\Configuration` supplies all of the functionality we need. In this case, we just need merging of local config files. 

`\Configuration\ModuleConfiguration` allows us to view a specific set of configuration data through the lens of a specific module so that we can setup the correct associations in the db.

bin/acl-config
ryanrath: kk

bin/acl-config
ryanrath: doesn't seem like it, I'll go ahead and remove it.

classes/Configuration/XdmodConfiguration.php
ryanrath: So this one's an interesting one, I've done some local testing which I'll detail below:

For the scenario that I believe you're referencing, I ran the following:
```php
<?php 

function findExtends(\stdClass $config, $parentKey = null)
{
    $extends = array();
    foreach ($config as $k => $v) {
        if ($k === 'extends' && is_string($v)) {
            $extends[$parentKey] = $v;
        } elseif (is_object($v)) {
            $extends = array_merge($extends, findExtends($v, $k));
        }
    }
    return $extends;
}

$empty = new stdClass();
$empty->extends = 'default';

print_r( findExtends($empty) );
```

and I ended up with: 
```
Array
(
    [] => default
)
```

After plugging in `$this->resolveExtends($source, $extenders);` and the code to support it ( I wanted to see if this code would exception out when it was passed the malformed data ). I ended up with: 
```
stdClass Object
(
    [extends] => default
)
```

So I think my slightly longer answer to the implied question of, how does the code react when $parentKey is null and it encounters an 'extends' keyword? would be that, it goes along its merry old way and processes things just fine. I don't think that the user / admin who has the funktastic  configuration files is going to get the behavior they expect. But the code won't fall over, at least at this point, if they do.

classes/Configuration/ModuleConfiguration.php
ryanrath: the call to `parent::merge();` ensures that `$this->transformedConfig` is populated. The `ModuleConfiguration` specific portion of `merge` makes sure that `annotatedConfig` is populated. 

Basically I just wanted to make sure that if the `ModuleConfiguration` is upcast / used as an `XdmodConfiguration` it would still function properly.

classes/Configuration/XdmodConfiguration.php
ryanrath: I could come up with a quick json schema that describes, generically, how our config files should be setup, so that we can throw a quick error pre-processing if they have something funky going on.

smgallo: I just want to be sure that we don't end up with some sort of strange or undefined behavior if `findExtends()` gets called somewhere without the 2nd argument such as in `handleExtendsFor()`

classes/Configuration/XdmodConfiguration.php
```diff
@@ -75,7 +75,7 @@ protected function processExtends()
      * result in merging properties from the `extends` target into the object that defined the
      * `extends` property. After the processing is done the `extends` property will be removed.
      *
-     * @param $source
+     * @param stdClass $source
      * @throws Exception
      */
     protected function handleExtendsFor($source)
```
smgallo: Should that be `protected function handleExtendsFor(\stdClass $source)` ?

ryanrath: yup, incoming in latest commit.

PR #796, opened 2019-02-08
Remove unused columns

chakrabortyr:
## Description
Removed start event ids from the cloud transient table. Also made the wallduration datatype consistent between transient and aggregate tables.

## Motivation and Context
Event ids are no longer part of any unique constraints.

## Tests performed
Made sure the existing numbers weren't changed.

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_common/cloud_transient.json
+++ b/configuration/etl/etl_tables.d/cloud_common/cloud_transient.json
@@ -20,11 +20,6 @@
                 "nullable": false
             },
             {
-                "name": "start_event_id",
-                "type": "int(11)",
-                "nullable": false
-            },
-            {
                 "name": "start_event_type_id",
                 "type": "int(11)",
                 "nullable": false
--- a/configuration/etl/etl_action_defs.d/cloud_common/cloud_transient.json
+++ b/configuration/etl/etl_action_defs.d/cloud_common/cloud_transient.json
@@ -9,7 +9,6 @@
             "start_event_type_id": "e.start_event_id",
             "end_time": "e.end_time",
             "end_event_type_id": "e.end_event_id",
-            "start_event_id": "ev.event_id",
             "resource_id": "it.resource_id",
             "instance": "it.provider_identifier",
             "instance_type": "itt.instance_type",
--- a/configuration/etl/etl_tables.d/cloud_common/event.json
+++ b/configuration/etl/etl_tables.d/cloud_common/event.json
@@ -63,7 +63,6 @@
                 "name": "PRIMARY",
                 "columns": [
                     "resource_id",
-                    "event_id",
                     "instance_id",
                     "event_time_utc",
                     "event_type_id",


configuration/etl/etl_tables.d/cloud_common/cloudfact_by_.json
```diff
@@ -102,7 +102,7 @@
                 "comment": "FACT: Number of sessions ended over a given period."
             },{
                 "name": "wallduration",
-                "type": "decimal(18,0)",
+                "type": "bigint(18)",
```
jpwhite4: is 18 digits big enough?

smgallo: Why the change from decimal to bigint?

plessbd: What is the point of `(18)`?

Seems arbitrary in the case of an int.

https://dev.mysql.com/doc/refman/5.5/en/numeric-type-overview.html
```
M indicates the maximum display width for integer types. The maximum display width is 255. Display width is unrelated to the range of values a type can contain, as described in Section 11.2, “Numeric Types”. For floating-point and fixed-point types, M is the total number of digits that can be stored. 
```

chakrabortyr: Yes, as per our conversation it should scale to a sufficiently high order of magnitude.

plessbd: 18 digits in this case is just a display and seems arbitrary, I would just remove it and use the default (which for an unsigned int is 20 anyway)

If your going to make it consistent make it make sense, I would say update the others to match as well.

configuration/etl/etl_tables.d/cloud_common/cloudfact_by_.json
```diff
@@ -62,17 +62,17 @@
...
             },{
                 "name": "core_time",
-                "type": "bigint(42)",
+                "type": "bigint(20) unsigned",
                 "nullable": false,
                 "comment": "FACT: Core hours reserved by a VM."
             },{
                 "name": "memory_reserved",
-                "type": "bigint(42)",
+                "type": "bigint(20) unsigned",
```
plessbd: please update the comment as to what the value stored in here is (bytes, KB, MB, GB)
Same with storage below

plessbd: Reason for that is, it doesnt make sense to store memory in bytes, and if it is stored in GB it doesnt make sense to store this as a bigint when we wouldnt be able to ever have that large of a number.

Same with storage...

configuration/etl/etl_tables.d/cloud_common/cloud_transient.json
```diff
@@ -71,12 +66,12 @@
             },
             {
                 "name": "start_time_ts",
-                "type": "bigint(18)",
+                "type": "bigint(20) unsigned",
```
plessbd: technically the start_time_ts could be negative, but I dont think we have any timestamps from pre 1970...

plessbd:
If you are going to make changes other than remving columns the title of this should probably be written as "database table cleanup/updates"

configuration/etl/etl_tables.d/cloud_common/cloudfact_by_.json
jpwhite4: What about systems that have different word sizes? E.g. the UNIVAC III has 25 bit word size.

jpwhite4:
Also won't someone please think of the children!

configuration/etl/etl_tables.d/cloud_common/cloudfact_by_.json
plessbd: Just to note
`bigint(20)` is how it is stored if you dont specify the display width, so `bigint(20)` will need to be what is stored in teh JSON file to allow for the detection of changes in etl v2

PR #797, opened 2019-02-13
Add cloud user and system account group by

eiffel777:
## Description
This pull request adds group by's for the person and system username for the cloud realm. The person group by will show the first and last name of a person associated with a username. The System Username group by will show the username of the person associated with the data being shown. In order to use these group by's usernames from the cloud event logs files are loaded into the modw.systemaccount and modw.person tables using pipelines added to jobs_cloud_generic.json and jobs_cloud_openstack.json. The new pipelines take actions from existing pipelines used for adding usernames from job files and modifies them for use in the cloud realm. In order for the first and last name to be shown when using the User group by the `xdmod-import-csv` command with the `-t names` option should be used, which is the same method for loading full names for Jobs data.

Each VM session will now have a person and system username associated with it. A sessions may have multiple events by multiple people during its run. We have choosen to use the person who started the session as the person who is responsible for the session and the resources used by it. 

A new ConfigFilesMigration.php file has made for upgrading from a previous installation of xdmod as the new group by's need to be added to the roles.json file. This file checks to see if the cloud.json file exists in the CONFIG_DIR/roles.d folder and if it does the group by's are added to the public and default roles. If the cloud.json file is not found an exception is thrown. Since the cloud.json file exists in the roles.d folder the writeJsonConfigFile function in /classes/OpenXdmod/Migration/ConfigFilesMigration.php cannot be used so I added a function called writeJsonPartialConfigFile that will write a partial json config file to whatever partial config file subdirectory that is specified. When upgrading all previous events in the database will default to -1 for the person and system username. If a person re-ingests their events those events will be updated to have the correct person and system username.

The user group by for the cloud realm was already created but was not active. To activate it entries were added for in to the roles.json and datawarehouse.json files.

## Tests performed
Manually tested in docker and new regressions tests for each group by have been added and the integration tests have been modified to account for the new group by's.

## Types of changes

## Checklist:


classes/OpenXdmod/Migration/Version800To810/ConfigFilesMigration.php
```diff
@@ -0,0 +1,72 @@
...
+     * Data from cloud.json file in CONFIG_DIR/roles.d folder
+     */
+    private $cloudRolesFile;
+
+    /**
+     * Execute the migration.
+     */
+    public function execute()
+    {
+        $this->setCloudRolesFile();
```
jpwhite4: Please explain why this config migration exists. You should be able to update the roles.d/cloud.json file and the RPM will install it in the config directory.

classes/OpenXdmod/Migration/Version800To810/ConfigFilesMigration.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Find cloud.json file in config directory and assign it to $this->clouRolesFile
+     */
+    private function setCloudRolesFile()
+    {
+        $rolesConfigFolder = $this->config->getPartialFilePaths('roles');
+
+        if($cloudFile = array_search(CONFIG_DIR."/roles.d/cloud.json", $rolesConfigFolder) === false){
+            throw new Exception("cloud.json file not found in roles.d folder");
```
jpwhite4: What happens to this exception? Will it completely kill the upgrade process? If so how does the user recover?

classes/OpenXdmod/Migration/Version800To810/ConfigFilesMigration.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Add a group by to a role in the cloud.json file
+     *
+     * @param $groupBy Name of group by that is being added
+     * @param $role Role that group by is being added to
+     */
+    private function addCloudRolesGroupBy($groupBy, $role)
+    {
+        if(!array_key_exists($role, $this->cloudRolesFile['+roles'])){
+            throw new Exception("Role not found in cloud.json file");
```
jpwhite4: What happens to this exception? Will it completely kill the upgrade process? If so how does the user recover?

configuration/etl/etl.d/jobs_cloud_generic.json
```diff
@@ -31,6 +31,12 @@
             "options_class": "IngestorOptions",
             "truncate_destination": false,
             "enabled": true
+        },
+        "jobs-cloud-import-users-generic": {
+            "namespace": "ETL\\Ingestor",
+            "options_class": "IngestorOptions",
+            "truncate_destination": false,
+            "enabled": true
```
jpwhite4: `enabled: true` is the default and does not need to be specified

configuration/etl/etl.d/jobs_cloud_generic.json
```diff
@@ -124,6 +130,128 @@
...
+          "description": "OpenStack cloud account data",
+          "class": "DatabaseIngestor",
+          "definition_file": "cloud_generic/union_user_pi.json",
+          "endpoints": {
+              "destination": {
+                  "type" : "mysql",
+                  "name" : "Cloud DB",
+                  "config" : "datawarehouse",
+                  "schema": "mod_shredder",
+                  "create_schema_if_not_exists": "true"
```
jpwhite4: If the setting is the same as the default then you should not override it. E.g. create_schema_if_not_exists is specifed as true in the defaults at the top of the file.

db/migrations/8.0.0-8.1.0/modw_cloud.sql
```diff
@@ -4,7 +4,7 @@ USE modw_cloud;
 -- somme OpenStack events. The events that already exist in the staging and event
 -- need to have their mappings updated. Some of the updates include mapping the
 -- compute.instance.power_on.start event to POWER_ON_START event and
--- compute.instance.resume.start to REQUEST_RESUME and compute.instance.resume.end 
+-- compute.instance.resume.start to REQUEST_RESUME and compute.instance.resume.end
```
jpwhite4: As a general rule I prefer to not have changes that only change whitespace. This is because it makes it more difficult for code reviewers to review correctly. For example, in this case, this file shows in the list of changed files relating to adding the user and system group bys. However there is no need fo this file to change at all for the  purpose of this pull request.

classes/DataWarehouse/Query/Cloud/GroupBys/GroupByPerson.php
```diff
@@ -128,29 +128,4 @@ public function pullQueryParameterDescriptions(&$request)
             "select long_name as field_label from modw.person  where id in (_filter_) order by order_id"
         );
     }
```
jpwhite4: How come this is being removed?

classes/OpenXdmod/DataWarehouseInitializer.php
```diff
@@ -207,7 +207,7 @@ public function ingestCloudDataOpenStack()
         if( $this->isRealmEnabled('Cloud') ){
             try{
                 $this->logger->notice('Ingesting OpenStack event log data');
-                Utilities::runEtlPipeline(array('jobs-cloud-extract-openstack'), $this->logger);
+                Utilities::runEtlPipeline(array('jobs-cloud-import-users-openstack', 'jobs-cloud-extract-openstack'), $this->logger);
```
jpwhite4: Is there a scenario where you would run the jobs-cloud-extract-* pipeline and not run the import-users one? If not then I suggest that you put the import users actions into the existing pipeline.

classes/OpenXdmod/Migration/ConfigFilesMigration.php
```diff
@@ -193,6 +193,23 @@ protected function writeJsonConfigFile($name, array $data)
...
     }
 
+    /**
+     * Write the contents of a partial config file.
+     *
+     * @param string $directory The directory where the partial config file lives (without ".d")
+     * @param string $name The config file name (without ".json").
+     * @param array $data The data to store in the config file.
+     */
+    protected function writeJsonPartialConfigFile($directory, $name, array $data)
```
jpwhite4: Question the need for another dedicated function that writes a json config file. I note that we already have a `JSON::saveFile()` that does exactly the same thing. The code that calls this function already knows about how to construct the path to the file. In fact you even use the sister `JSON::loadFile()` function a few lines before calling this one.  See `classes/OpenXdmod/Migration/Version800To810/ConfigFilesMigration.php` lines 36 and 51

```
Json::loadFile(CONFIG_DIR."/roles.d/cloud.json");

chakrabortyr:
LGTM once you address @jpwhite4's outlying comment.

jpwhite4:
Don't forget to squash and merge

PR #803, opened 2019-02-18
add some summary charts for the cloud realm

plessbd:
This adds two default summary tab charts when the cloud realm is enabled.

### DIFF
--- a/templates/roles.d/cloud.json
+++ b/templates/roles.d/cloud.json
@@ -26,8 +26,77 @@
                     "group_by": "vm_size_memory"
                 },
                 {
-                  "realm": "Cloud",
-                  "group_by": "submission_venue"
+                    "realm": "Cloud",
+                    "group_by": "submission_venue"
+                }
+            ],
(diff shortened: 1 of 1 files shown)


templates/roles.d/cloud.json
```diff
@@ -26,8 +26,79 @@
...
+                        "data": [{
+                            "combine_type": "stack",
+                            "display_type": "column",
+                            "filters": {
+                                "data": [],
+                                "total": 0
+                            },
+                            "group_by": "project",
+                            "has_std_err": true,
+                            "id": 0.00000000000004,
```
smgallo: Remind me what the heck this `id` field is for...

plessbd: For the fun of it!
Upon discussion with @jpwhite4 we think it has to do with this being copied from the usage page where these "ID's" are used to store data in relation to the treeview.  These have been removed from this pull request and has had no ill effects.

PR #804, opened 2019-02-19
Update with some of the options that are avalbile

plessbd:
These should have been added when the options were added, but they were not, and I need to keep reminding myself what they are, so I added some help.

### DIFF
--- a/bin/xdmod-ingestor
+++ b/bin/xdmod-ingestor
@@ -333,6 +333,12 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
+    --aggregate *realm*
+        If specified optional realm to aggregate (job, cloud, storage)
+
+    --datatype [openstack, genericcloud, storage]
+        Ingest one specific datatype
+
     -q, --quiet
         Output warning level and above log messages.
 
--- a/bin/xdmod-ingestor
+++ b/bin/xdmod-ingestor
@@ -334,10 +334,11 @@ Usage: xdmod-ingestor [-v]
         Output debug level and above log messages.
 
     --aggregate[=*realm*]
-        If specified optional realm to aggregate (job, cloud, storage)
+        If a *realm* (job, cloud, storage) is specified aggregate that realm
+        otherwise aggregate all realms.
 
-    --datatype [openstack, genericcloud, storage]
-        Ingest one specific datatype
+    --datatype *type*
+        Ingest one specific datatype (openstack, genericcloud, storage).
 
     -q, --quiet
         Output warning level and above log messages.
--- a/bin/xdmod-ingestor
+++ b/bin/xdmod-ingestor
@@ -333,7 +333,7 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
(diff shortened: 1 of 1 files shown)


bin/xdmod-ingestor
```diff
@@ -333,6 +333,12 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
+    --aggregate *realm*
+        If specified optional realm to aggregate (job, cloud, storage)
+
+    --datatype [openstack, genericcloud, storage]
```
smgallo: The OCD in me wants this to be `cloud-openstack, cloud-generic, storage`

jpwhite4:
What about the rest of the options????!!!!?!?!?!?!?!??!?!?!?!?!?

bin/xdmod-ingestor
```diff
@@ -333,6 +333,12 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
+    --aggregate *realm*
+        If specified optional realm to aggregate (job, cloud, storage)
```
jtpalmer: You should indicate that the realm is optional and that if you do specify the realm that you need to use `=` as the separator and not space.

smgallo: @jtpalmer Is that the case with the other options as well?

bin/xdmod-ingestor
```diff
@@ -333,6 +333,12 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
+    --aggregate *realm*
```
jtpalmer:
```suggestion
    --aggregate[=*realm*]
```

bin/xdmod-ingestor
```diff
@@ -333,6 +333,12 @@ Usage: xdmod-ingestor [-v]
     --debug
         Output debug level and above log messages.
 
+    --aggregate[=*realm*]
+        If specified optional realm to aggregate (job, cloud, storage)
+
+    --datatype [openstack, genericcloud, storage]
+        Ingest one specific datatype
```
jtpalmer:
```suggestion
        Ingest one specific datatype.
```

bin/xdmod-ingestor
jtpalmer: This is the only option that optionally accepts an argument.

smgallo: I'm wondering why this one requires an `=` rather than a space like the other arguments.

jtpalmer: And the requirement to use `=` for optional arguments is from my reading of the [`getopt`](http://php.net/getopt) documentation:

```
Note: Optional values do not accept " " (space) as a separator.
```

bin/xdmod-ingestor
plessbd: That would be @eiffel777 who chose those, I'm just documenting it ;)

plessbd:
@jpwhite4 if you provide me a list of the other options in the format of

```
--optionName *optionalOptionvalue*
    Description of option
```

I am sure I can add them in ;)

PR #805, opened 2019-02-20
Deterministic reconstruction with no non-lifecycle events

chakrabortyr:
## Description
Guarantees a deterministic order for events received by the event reconstructor. Also no longer passes events to the reconstructor which do not pertain to the lifecycle of a VM session (starts/stops/heartbeats)

## Motivation and Context
We need a guaranteed, deterministic order for events to arrive in and the removal of non lifecycle events makes reconstruction simpler by not querying for unnecessary events.

## Tests performed
Made sure the numbers didn't change.

## Types of changes

## Checklist:


classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -137,7 +137,7 @@ protected function getSourceQueryString()
         $colCount = count($this->etlSourceQuery->records);
         $unionValues = array_fill(0, $colCount, 0);
 
-        $sql = "$sql \nUNION ALL\nSELECT " . implode(',', $unionValues) . "\nORDER BY 1 DESC, 2 DESC, 3 ASC";
+        $sql = "$sql WHERE event_type_id IN (" . implode(',', $this->_all_event_ids) . ")\nUNION ALL\nSELECT " . implode(',', $unionValues) . "\nORDER BY 1 DESC, 2 DESC, 3 ASC, 4 DESC";
```
jpwhite4: This change results in dead code on lines 104-106.

jpwhite4: The events are typically in the database in start -> stop order. This leads me to think that it would be better to order by `4 ASC`. Why did you choose `4 DESC`?

jpwhite4: Or maybe it doesn't matter. In any case did you have a reason for choosing `4 DESC`

jpwhite4:
After a bit of detective work it appears that we really have to start using better time precision. I found cases where the 4 ASC version is wrong and cases where the 4 DESC version is wrong. The data below shows the database rows for a particular instance computed with 4ASC and 4DESC.
`4ASC`
```
2018-07-23 17:03:54	START	2018-07-23 20:00:36	POWER_OFF_END
2018-10-11 12:36:26	POWER_ON_END	2018-10-17 10:19:52	POWER_OFF_END
```
verses `4DESC`
```
2018-07-23 17:03:54	START	2018-07-23 20:00:36	POWER_OFF_END
2018-07-23 20:00:36	RESUME	2018-10-17 10:19:52	POWER_OFF_END
2018-10-17 10:19:52	RESUME	2018-10-18 06:59:33	TERMINATE
```
The actual data is below and matches the 4ASC version in this case (there are other examples in the same data where the desc version gets it right and asc gets it wrong). 
```
   "event_type": "compute.instance.create.start",     "generated": "2018-07-23T17:03:40.138716",
   "event_type": "compute.instance.create.end",     "generated": "2018-07-23T17:03:54.929757",
   "event_type": "compute.instance.power_off.start",     "generated": "2018-07-23T20:00:31.285698",
   "event_type": "compute.instance.power_off.end",     "generated": "2018-07-23T20:00:36.452734”,

   "event_type": "compute.instance.power_on.start",     "generated": "2018-10-11T12:36:25.728863",
   "event_type": "compute.instance.power_on.end",     "generated": "2018-10-11T12:36:26.979636",
   "event_type": "compute.instance.power_off.start",     "generated": "2018-10-17T10:19:48.574615",
   "event_type": "compute.instance.power_off.end",     "generated": "2018-10-17T10:19:52.752060",
```

Of course the real problem is that we are throwing away the microseconds from the data.

chakrabortyr:
I'd be interested in seeing the numbers the integration tests generate when we switch from DESC to ASC especially if the row counts are drastically different. My intuition is that the integration tests will then fail, I'll confirm tomorrow.

The proper fix is probably to stop throwing out information, given the frequency which events can occur in the same second.

PR #806, opened 2019-02-20
Replace `Config` Usage w/ `XdmodConfiguration`

ryanrath:
**I'm going to apologize up front for the size of this PR but it turns out we use configuration files fairly extensively. On the plus side the vast majority of changes you'll be seeing are the same, just in different places.**

## Description
This PR removes all usage `Xdmod\Config` that is not in a file that itself will be removed soon ( `aRole` I'm looking at you! ) and replaces it w/ the newly consolidated `XdmodConfiguration`. `XdmodConfiguration` is derived from the same base class, `Configuration\Configuration` as the code utilized by ETLv2. After this PR has been merged all that should be left is to remove `Xdmod\Config` and the its associated tests.

In general the changes found in this PR fit one of a few use cases:

**Standard Usage / Replacement**

*Old Code:*
```php
<?php

// instance of \Log or CCR\Log
$logger;

$config = Config::factory();

$specificConfig = $config['specific'];

// do stuff w/ $specificConfig and use $logger to log
```

*New Code:*
```php
<?php

$configFile = new \Configuration\XdmodConfiguration(
    'specific.json',
    CONFIG_DIR,
    $logger,
    array(
        'local_config_dir' => implode(DIRECTORY_SEPERATOR, array(CONFIG_DIR, 'specific.d))
    )
);
$configFile->initialize();

$specificConfig = json_decode($configFile->toJson(), true);

// do the same stuff w/ $specificConfig as in the old code.
```

**`getFilePath`|`getConfigDirPath` usage / replacement**

*Old Code:*
```php
<?php

$config = Config::factory();

$configFilePath = $config->getFilePath('config-file.json');

// Do something with $configFilePath
```

*New Code:*
```php
<?php

$configFilePath = implode(
    DIRECTORY_SEPARATOR,
    array(
      CONFIG_DIR,
      'config-file.json'
    )
);

// Do same things w/ $configFilePath
```

## Motivation and Context
Having two sets of code for working with Configuration files is not needed. This allows any further enhancement to the Configuration processing code to be utilized across the entire code base. 

## Tests performed
For Docker - OpenXDMoD Install / Upgrade:
- Unit Tests
- Component Tests
- Integration Tests
- UI Tests
- REG_TEST_ALL=1 Regression Tests

Manual Testing:
- Log into the Internal Dashboard
- Ensure that the Summary Page displays correctly
- Ensure that the Log Tab displays correctly

## Types of changes

- [X] Code Refactor / Consolidation

## Checklist:


background_scripts/update_check.php
```diff
@@ -32,9 +31,14 @@ function main()
...
-    $config = Config::factory();
+    $configFile = new \Configuration\XdmodConfiguration(
+        'update_check.json',
+        CONFIG_DIR,
+        $logger
+    );
+    $configFile->initialize();
 
-    $updateConfig = $config['update_check'];
+    $updateConfig = json_decode($configFile->toJson(), true);
```
smgallo: Since we're using this quite a bit, does it make sense to add this directly to `Configuration`?

```
/** -----------------------------------------------------------------------------------------
 * Return an associative array based on the parsed and translated Configuration.
 *
 * @return array An associative array generated from the Configuration object.
 * ------------------------------------------------------------------------------------------
 */

public function toAssocArray()
{
    return json_decode(json_encode($this->transformedConfig), true);
}  // toJson()
```

classes/Configuration/Configuration.php
```diff
@@ -373,6 +373,8 @@ private function parse($force = false)
         $parsed = $jsonFile->parse();
         if ($parsed === false) {
             $parsed = new stdClass();
+        } elseif(count($jsonFile) > 1) {
```
smgallo: Lets add a comment to the top of this conditional explaining that this supports 2 formats of JSON (a single object and an array).

classes/Configuration/Configuration.php
```diff
@@ -413,7 +415,17 @@ private function transform()
         // any properties that are references to other variables will remain references
 
         $tmp = unserialize(serialize($this->parsedConfig));
-        $this->transformedConfig = $this->processKeyTransformers($tmp);
+        if (is_array($tmp)) {
```
smgallo: Same here, add a comment explaining that we are preserving the array.

classes/Configuration/XdmodConfiguration.php
```diff
@@ -67,7 +67,17 @@ protected function postMergeTasks()
      */
     protected function processExtends()
     {
-        $this->handleExtendsFor($this->transformedConfig);
+        if (is_array($this->transformedConfig)) {
```
smgallo: Same here with the comment...

classes/DB/FilterListBuilder.php
```diff
@@ -281,8 +292,23 @@ private function checkDimensionForRoles(GroupBy $groupBy)
...
+                    'local_config_dir' => implode(
+                        DIRECTORY_SEPARATOR,
+                        array(CONFIG_DIR, 'roles.d')
+                    )
+                )
+            );
+            $configFile->initialize();
+
+            $config = json_decode($configFile->toJson(), true);
+            $roles = $config['roles'];
```
smgallo: The `Configuration` object supports `$configFile->roles`, is there a reason to convert to an array here and elsewhere?

background_scripts/update_check.php
ryanrath: Certainly wouldn't hurt. I'll add that in. Also, on a somewhat related note, what do you think about having `Configuration::initialize` return `$this`?

classes/DB/FilterListBuilder.php
ryanrath: Just because the existing code expects the configuration data to be associative arrays and it seemed easier to convert the new configuration to what the existing code expects then to try and update all the code that utilized the old configuration code to utilize objects.

background_scripts/update_check.php
smgallo: Returning `$this` should be OK.

ryanrath: Yeah, after thinking about it a bit, I don't think the `Configuration::initialize` really buys us anything. I'll get that `toAssocArray` up in just a few though.

PR #807, opened 2019-02-20
Improve resiliency of ETLv2 manage tables

smgallo:
## Description

When comparing table definitions generated from ETL configuration files to those stored in the MySQL information_schema, MySQL normalizes several items such as column and index names to lowercase.  In the case where these values were capitalized in the ETL configuration files, this would cause an endless and unnecessary loop of `ALTER TABLE` statements to be executed during each ETL execution.

As per https://dev.mysql.com/doc/refman/5.5/en/identifier-case-sensitivity.html
```
The case sensitivity of the underlying operating system plays a part in the case sensitivity of database, table, and trigger names. However, column, index, stored routine, and event names are not case sensitive on any platform, nor are column aliases.
```

To aid in identifying items that have changed, logging and error checking was improved. If a table element definition and the current state of the database was found to differ the exact reason in the configuration is logged:

```
2019-02-19 20:38:03 [debug] Discover table 'modw.job_tasks'
2019-02-19 20:38:03 [debug] Column local_job_id_raw: values for "type" differ ("int(11)" != "int(11) unsigned")
2019-02-19 20:38:03 [debug] Column resource_id: values for "type" differ ("int(11)" != "int(11) unsigned")
2019-02-19 20:38:03 [debug] Column person_id: values for "type" differ ("int(11)" != "int(11) unsigned")
2019-02-19 20:38:03 [debug] Index uniq: values for "columns" differ ("organization_origin_id,federation_instance_id" != organization_origin_id,federation_blade_id")
```

**NOTE:** Since variables/macros can be used to create dynamic column names and column names are normalized to lowercase as stored in the MySQL information schema, it was necessary to make the variable names themselves case-insensitive (e.g., ${MYVAR} and ${myvar} are treated as the same variable). Note that since variable substitution occurs after the `DbModel` has been built, the value of those variables must also be lowercase to mitigate an needless `ALTER TABLE` statements.

## Motivation and Context

Stop continual `ALTER TABLE` statements when they are not truly necessary.

## Tests performed

Added tests.

```
$ ../xdmod-qa/travis/build.sh -s
Running only style tests
X11 forwarding request failed on channel 0
Comparing HEAD to upstream/xdmod8.1 (c1e4b1fcd9870ef68322d036f82c7aa7c7f9b21e)
travis_fold:start:syntax
Running syntax tests...
travis_fold:end:syntax
Syntax tests succeeded!

travis_fold:start:style
Running style tests...
travis_fold:end:style
Style tests succeeded!

travis_fold:start:extra
Running extra tests...
travis_fold:end:extra
Extra tests succeeded!
```
## Types of changes

## Checklist:


classes/ETL/DbModel/Trigger.php
```diff
@@ -77,6 +78,13 @@ protected function filterAndVerifyValue($property, $value)
                         sprintf("%s name must be a string, '%s' given", $property, gettype($value))
                     );
                 }
+
+                // Normalize property values to lowercase to match MySQL behavior
+                if ( in_array($property, array('time', 'event')) ) {
+                    $value = strtoupper($value);
+                } elseif ( 'body' == $property && 0 !== stripos($value, "BEGIN") ) {
```
ryanrath: Quick question, is the `'body' == $property ...` portion of this `elseif` needed? It looks like, based on the `switch` statement above that it will always be `'body'` at this point.

ryanrath: *facepalm* fall through w/ the other `case`s. It won't always be `body`

smgallo: It could be `table`, which is case sensitive.

PR #817, opened 2019-02-25
Updates to  to support Shredder

ryanrath:
## Description
The  class requires the results of  as an array of arrays
regardless of how many resources there are. This change adds a new option to
that lets instantiators specify if they want the results of  to be wrapped in an array
if it would not normally be.

## Motivation and Context
Not broken is good.

## Tests performed
Tests need to be / will be added. It's being submitted in this form to get the info up.

## Types of changes

## Checklist:


tests/artifacts/xdmod-test-artifacts/xdmod/configuration/input/to_assoc_array.json
```diff
@@ -0,0 +1,34 @@
...
+    }
+  ],
+  [
+    {
+      "base_file": "resources",
+      "options": [],
+      "expected": "resources"
+    }
+  ]
+]
```
plessbd: should have a new line, not sure why travis is not catching it...

tests/artifacts/xdmod-test-artifacts/xdmod/configuration/output/resources.json
```diff
@@ -0,0 +1 @@
+[{"resource":"frearson","resource_type_id":1,"name":"Frearson"},{"resource":"mortorq","resource_type_id":1,"name":"Mortorq"},{"resource":"phillips","resource_type_id":1,"name":"Phillips"},{"resource":"pozidriv","resource_type_id":1,"name":"Posidriv"},{"resource":"robertson","resource_type_id":1,"name":"Robertson"},{"resource":"openstack","resource_type_id":5,"name":"OpenStack"}]
```
plessbd: Since this is new I would like to see it pretty printed (but not a stickler for it)
It also needs a new line...

tests/artifacts/xdmod-test-artifacts/xdmod/configuration/input/to_assoc_array.json
ryanrath: yeah, I specifically looked for this and was like, welp it's not erroring *shrugs*. I'll go in and add the newlines where they should be

plessbd:
These files are the same we should be able to combine them:
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/output/resources.json
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/output/resources_force_array_return.json


Technically the following are the same, but one is input the other is expected output, not sure if we should / could easily combine them

So are these:
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/input/single_resource.json
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/output/single_resource_force_array_return.json

And these
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/input/resources.json
tests/artifacts/xdmod-test-artifacts/xdmod/configuration/output/resources.json

smgallo:
Lets keep input/output in separate directories even if they're the same unless we can reliably symlink them.

classes/OpenXdmod/Migration/Version350To450/ConfigFilesMigration.php
```diff
@@ -28,7 +28,11 @@ public function execute()
         // in a separate file to allow changes over time.
         $configFile = new XdmodConfiguration(
             'resources.json',
-            CONFIG_DIR
+            CONFIG_DIR,
```
jpwhite4: Do not edit this file.

plessbd:
I was specifically thinking that if input and output should be the same it should be as simple as

```
loaded = json_decode(file_read_contents(blah))

loaded == transform(loaded)
```

smgallo:
@plessbd yep, that would be ok

classes/OpenXdmod/Migration/Version350To450/ConfigFilesMigration.php
ryanrath: Along that same lines, does this file need to exist in the current repo? Or any of the old Migrations for that matter?

ryanrath: Aaaaaand it's been removed it looks like. Guess that answers that. I'll see about removing it from the PR.

plessbd:
as an FYI to anyone else looking at this, I verified the changes (not the tests) on live machines.

PR #824, opened 2019-02-26
Remove unused javascript files

jpwhite4:
These files are not used anywhere. In fact, a quick review of the contents suggests to me that this was never actually used - they appear to be work-in-progress stuff that never made it to production.


jpwhite4:
Of particular note is the `login.js/php` file. This contains multiple calls to the jQuery `$()` function. The `$` function does not exist in XDMoD.

PR #826, opened 2019-02-27
More fixing for datetime changes

plessbd:
Found some more places where the generic and openstack datetimes needed to be fixed.

Whitespace changes from `(16, 3)` to `(16,3)` is required so that the ETL does not do an alter table

Travis will have to be overridden.

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_common/event_reconstructed.json
+++ b/configuration/etl/etl_tables.d/cloud_common/event_reconstructed.json
@@ -15,7 +15,7 @@
             },
             {
                 "name": "start_time",
-                "type": "decimal(16, 6)",
+                "type": "decimal(16,6)",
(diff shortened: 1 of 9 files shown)


configuration/etl/etl_tables.d/cloud_common/event_reconstructed.json
```diff
@@ -15,7 +15,7 @@
             },
             {
                 "name": "start_time",
-                "type": "decimal(16, 6)",
+                "type": "decimal(16,6)",
```
chakrabortyr: No spurious whitespace changes.

smgallo: Necessary to be compatible with mysql-normalized types.

PR #834, opened 2019-03-01
Fix regression in getJobDataSet introduced in #733

jpwhite4:
This change was incorrectly ommited in #733.

Tests will be added to the xdmod-supremm repo to check this code path


jpwhite4:
Note that the test is in https://github.com/ubccr/xdmod-supremm/pull/147

PR #835, opened 2019-03-01
Add node_modules to the RPM.

jpwhite4:
node_modules dependencies are now packaged in the RPM. There will be a corresponding update to the supremm RPM to account for this.

This change includes a %pre hook in the rpm that saves the (non-rpm managed) node_modules directory that could be present on an upgrade. This %pre hook should be removed by the next release. I have included a defensive check to ensure that the save code only runs when the version of the rpm is 8.1.0


jpwhite4:
Need to update the included packages list in the documentation

PR #836, opened 2019-03-05
update OpenStackCloudStagingUserResourceIngestor

plessbd:
to allow for OS having null user names in the events

Tested and verified on federation cloud instances.


plessbd:
This will be replaced by a more comprehensive fix.

PR #838, opened 2019-03-05
Prevent null usernames being added when ingestion cloud data

eiffel777:
This prevents the usernames that are null for a cloud event from being ingested and causing an error when running the cloud ingestor pipelines

## Tests performed
Manually tested in docker and passed existing tests

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_action_defs.d/cloud_openstack/union_user_pi_resource.json
+++ b/configuration/etl/etl_action_defs.d/cloud_openstack/union_user_pi_resource.json
@@ -21,7 +21,7 @@
             }
         ],
         "where": [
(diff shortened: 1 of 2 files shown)


configuration/etl/etl_action_defs.d/cloud_openstack/union_user_pi_resource.json
```diff
@@ -19,6 +19,9 @@
                 "alias": "r",
                 "on": "r.id = raw.resource_id"
             }
+        ],
+        "where": [
+          "raw.user_name IS NOT NULL"
```
plessbd: i believe this should be 4 spaces and not 2...
same with the other file.

PR #840, opened 2019-03-05
Less infinite loops

ryanrath:
## Description
So now that we have a class that handles the merging + extends resolving (
`XdmodConfiguration` ) we no longer need to rely on `Roles::getConfig`.

Usage of this function was replaced in:
  - Tabs.php@58
  - summary3.php@68,75

The one bit of functionality we lose by doing this is that, when a section
cannot be found we no longer attempt to look it up in the `default` acl.

To account for this it was necessary to add a `summary_charts` section to the
`pub` acl. Which in turn required that this data be added to the expected
output in `artifacts/xdmod/acls/output/roles.json`.

~~**Note** Tests will be added shortly~~
**Note** Now that we're removing the problem function. New tests will no longer be needed as existing tests already cover these areas.

## Motivation and Context
Infinite loops bad? 

## Tests performed
Updated existing unit tests to account for latest changes.

## Types of changes

## Checklist:


classes/Models/Services/Tabs.php
```diff
@@ -3,6 +3,8 @@
 namespace Models\Services;
 
 use CCR\DB;
+use Configuration\XdmodConfiguration;
+use mysql_xdevapi\Exception;
```
jpwhite4: Where is this execption used?

configuration/roles.json
```diff
@@ -353,6 +353,156 @@
                     "group_by": "username",
                     "disable": true
                 }
+            ],
+            "summary_charts": [
```
jpwhite4: Please don't modify the summary_charts sections - it you do you will need to create the necessary migration code. I think that the configuration code refactoring work should not be forcing us to change the configuration files.

classes/Models/Services/Tabs.php
ryanrath: It's not, thanks for the catch

configuration/roles.json
ryanrath: *nod* Updated per our convo

PR #846, opened 2019-03-11
Updating linker config to use Configuration

ryanrath:
## Description
Replaced a section of linker.php that read in the `linker.json` config
file and replaced it with the new `XdmodConfiguration` class. This
allows it to have .d files that will be merged in.

**NOTE** This PR is per a convo with @jtpalmer about this causing problems when used in an XSEDE deployment.

## Motivation and Context
We should utilize the new `Configuration` classes where possible so that we can utilize all of the features they provide.

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -415,7 +415,15 @@
     },
     "autoload": {
         "classmap": [
-            "composer.scripts.php"
+            "composer.scripts.php",
+            "classes",
+            "classes/DB",
+            "classes/DB/DBModel",
+            "classes/DB/TimePeriodGenerators",
+            "classes/ExtJS",
+            "classes/Rest",
+            "classes/User",
+            "classes/ReportTemplates"
         ]
     }
 }
--- a/configuration/linker.json
+++ b/configuration/linker.json
@@ -1,22 +1,5 @@
 {
-    "autoloaders": [
-        "vendor/autoload.php"
-    ],
     "include_dirs": [
-        "classes",
-        "classes/DB",
-        "classes/DB/TACCStatsIngestors",
-        "classes/DB/TGcDBIngestors",
-        "classes/DB/POPSIngestors",
-        "classes/DB/XRASIngestors",
-        "classes/DB/Aggregators",
-        "classes/DB/DBModel",
-        "classes/DB/TimePeriodGenerators",
-        "classes/ExtJS",
-        "classes/REST",
-        "classes/User",
-        "classes/ReportTemplates",
-        "classes/AppKernel",
-        "libraries/HighRoller_1.0.5"
+
     ]
 }
--- a/configuration/linker.php
+++ b/configuration/linker.php
@@ -5,15 +5,37 @@ $baseDir = dirname($dir);
 
 require_once($dir . '/constants.php');
 
+require_once("$baseDir/classes/CCR/Json.php");
+require_once("$baseDir/libraries/utilities.php");
+
+// If present, load autoloader configuration from autoloader.json and autoloader.d.
+try {
+    $autoloaderConfigFilePath = implode(
+        DIRECTORY_SEPARATOR,
+        array(
+            CONFIG_DIR,
+            'autoloader.json'
+        )
+    );
+    $autoLoaderConfigFile = CCR\Json::loadFile(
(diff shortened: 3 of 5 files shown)


configuration/linker.php
```diff
@@ -5,35 +5,37 @@
 
 require_once($dir . '/constants.php');
 
-// ---------------------------
-
-// If present, load linker configuration from linker.json and linker.d.
 require_once("$baseDir/classes/CCR/Json.php");
+require_once("$baseDir/libraries/utilities.php");
 
+// If present, load autoloader configuration from autoloader.json and autoloader.d.
```
jtpalmer: Does this actually support `autoloader.d`?

composer.json
```diff
@@ -415,7 +415,15 @@
...
-            "composer.scripts.php"
+            "composer.scripts.php",
+            "classes",
+            "classes/DB",
+            "classes/DB/DBModel",
+            "classes/DB/TimePeriodGenerators",
+            "classes/ExtJS",
+            "classes/Rest",
+            "classes/User",
+            "classes/ReportTemplates"
```
jtpalmer: If you change this file you need to update the lock file.

configuration/linker.php
```diff
@@ -5,35 +5,37 @@
...
+        foreach ($autoLoaderConfigFile['autoloaders'] as $autoloaderPath) {
+            require_once("$baseDir/$autoloaderPath");
+        }
+    }
 } catch (Exception $e) {
     $configDir = $config->getConfigDirPath();
-    echo "Could not find valid \"linker.json\" or \"linker.d\" files in \"$configDir\".\n";
-    echo "Please set up valid linker configuration files and try again.\n";
+    echo "Could not find valid \"autoloader.json\" or \"autoloader.d\" files in \"$configDir\".\n";
+    echo "Please set up valid autoloader configuration files and try again.\n";
```
jtpalmer: In the line before this change `$config` is being used, but is not defined.  That was probably removed in the previous PR.

configuration/linker.json
```diff
@@ -1,22 +1,5 @@
...
-        "classes/DB/XRASIngestors",
-        "classes/DB/Aggregators",
-        "classes/DB/DBModel",
-        "classes/DB/TimePeriodGenerators",
-        "classes/ExtJS",
-        "classes/REST",
-        "classes/User",
-        "classes/ReportTemplates",
-        "classes/AppKernel",
-        "libraries/HighRoller_1.0.5"
```
jtpalmer: Those of these lines that are removed and not added to the `composer.json` will need to be moved into the XSEDE repository `linker.d/xsede.json` file.

configuration/linker.php
```diff
@@ -73,8 +75,6 @@ function xdmodAutoload($className)
 
 // Libraries ---------------------------
 
-require_once($baseDir . '/libraries/utilities.php');
```
jtpalmer: Is there a reason for moving this line?

configuration/linker.json
jtpalmer: ... and the `AppKernel` classes should be in xdmod-appkernels.

configuration/linker.php
ryanrath: yup, it's utilized in `Configuration`

configuration/linker.json
ryanrath: yup, have these entries in their respective module files now. As soon as this PR is ready to go I'll submit those as well.

configuration/linker.php
ryanrath: *nod* these lines will probably be removed in the next iteration of this PR.

composer.json
ryanrath: Yup, incoming in the next iteration of this PR.

configuration/linker.php
ryanrath: Nope, just a copy/paste/replace error. But this won't be present in the next iteration of this PR.

ryanrath:
I'm getting some funky errors on this branch for some reason. I'm going to go ahead and close it, create a new branch from current xdmod8.1, apply changes and then it'll be back.

PR #849, opened 2019-03-12
Addressing problem found by @jtpalmer

ryanrath:
**NOTE:** The Travis tests will fail because we changed `composer.json` but there is no accompanying change to `composer.lock`. This is expected. 

## Description
The problem is that when these entries were only in `composer.json` the
classpath was determined prior to run time, aka. before any other modules are
installed. So, if a module has code installed to any of these paths they will
not be found at round time. By moving them back into `linker.json` we resolve
this problem.

## Motivation and Context
runtime exceptions bad

## Tests performed
All automated tests on OpenXDMoD docker ( upgrade / install ) 

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -416,14 +416,7 @@
     "autoload": {
         "classmap": [
             "composer.scripts.php",
-            "classes",
-            "classes/DB",
-            "classes/DB/DBModel",
-            "classes/DB/TimePeriodGenerators",
-            "classes/ExtJS",
-            "classes/Rest",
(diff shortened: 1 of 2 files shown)


ryanrath:
I've created a new PR https://github.com/ubccr/xdmod-qa/pull/4 that gives us more control over the `composer.json` and how it determines if `composer.lock` should be updated. Once that PR is merged then we can go ahead with this one.

PR #851, opened 2019-03-13
Add charts reports portlets for center director

mkzia:
## Description

1. Add REST API that returns both charts and reports associated with a user
2. Add a reports and charts portlet that lists the charts and reports for a center director

## Motivation and Context

To make it easier for novice center directors to fetch their charts and reports.  

## Tests performed

I manually tested the changes. I will add unit tests for the REST API later. 

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ X] My code follows the code style of this project as found in the **CONTRIBUTING** document.

### DIFF
--- a/classes/XDReportManager.php
+++ b/classes/XDReportManager.php
@@ -475,7 +475,8 @@ class XDReportManager
                 r.format,
                 r.schedule,
                 r.delivery,
-                COUNT(rc.chart_id) AS chart_count
+                COUNT(rc.chart_id) AS chart_count,
+                r.last_modified
             FROM Reports r
             LEFT JOIN ReportCharts rc ON rc.report_id = r.report_id
             WHERE r.user_id = :user_id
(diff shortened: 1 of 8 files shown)


classes/XDReportManager.php
```diff
@@ -507,7 +509,8 @@ public function fetchReportTable()
                 'report_format'   => $entry['format'],
                 'report_schedule' => $entry['schedule'],
                 'report_delivery' => $entry['delivery'],
-                'chart_count'     => $entry['chart_count']
+                'chart_count'     => $entry['chart_count'],
+                'last_modified' => strtotime($entry['last_modified'])
```
jpwhite4: It is better to use UNIX_TIMESTAMP(r.last_modified) in the SQL query. If you do the conversion in php then there will be conversion problems when daylight savings begins/ends and or if the webserver and database are set to different timezones.

PR #866, opened 2019-03-21
Change staging table to use 1 as unknown id instead of -1

eiffel777:
For columns in the cloud staging tables related to the changes in #861 change the COALESCE statements to use 1 instead of -1.

## Tests performed
Ran all tests in docker

## Types of changes

## Checklist:


jpwhite4:
Did the pre-populated table data change too?

jpwhite4:
In the docs/cloud.md is says to run
```
php /usr/share/xdmod/tools/etl/etl_overseer.php -p jobs-common -p jobs-cloud-common -p ingest-resources
```
But I think that the jobs-cloud-common must be run after the ingest-resources otherwise the 'unknown' rows will not be populated properly.

jpwhite4:
Need to update the comment for account_id column in configuration/etl/etl_tables.d/cloud_common/account.json

jpwhite4:
Should not set the default person_id to 1 in configuration/etl/etl_tables.d/cloud_common/instance.json

jpwhite4:
The comment on the modw_cloud.region.region_id column states
```
"Auto-increment relative to resource_id. We do not have unknown assets."
```
yet this table is populated with an 'unknown' row by the initialize_unknown_dimension_values.json. Which is correct?

PR #867, opened 2019-03-22
Run usage charts regression tests

jtpalmer:
## Description

Run the usage chart regression tests.  Update expected data.

## Motivation and Context

These tests were not being run at all.  Only one PHP class (or directory) may be specified in the `phpunit` command line options.

These tests were removed in #620 (https://github.com/ubccr/xdmod/pull/620/files#diff-9e90ea356a9e4f18da418e5292f11fc9L38) and were thought to have been re-added in #652 (https://github.com/ubccr/xdmod/pull/652/files#diff-9e90ea356a9e4f18da418e5292f11fc9R38).  Unfortunately, since the PHP classes were specified individually instead of the entire directory, the tests in the first file were executed and not those in the second file.

The chart export must have also been changed while the tests were not running, so the expected data needed to be updated.

## Tests performed

Ran regression tests in Docker with `REG_TEST_ALL=1`.

## Types of changes

## Checklist:


open_xdmod/modules/xdmod/regression_tests/runtests.sh
```diff
@@ -60,7 +62,7 @@ else
     $phpunit $PUB lib/Controllers/UsageExplorerCloudTest.php & pubpid=$!
 
     EXIT_STATUS=0
-    for pid in $usrpid $pipid $cdpid $cspid $pubpid;
+    for pid in $usrpid $pipid $cdpid $cspid $pubpid $ucpubpid;
```
plessbd: I think we also need another PR to fix the pids for the cloud tests not being the same as the pid for the non cloud tests and to make sure the cloud tests dont overwrite the output from the non cloud tests...

We should probably rewrite this whole file to be better...i take the blame for the initial ugliness

jtpalmer: I'll be doing that in my storage regression tests changes.

PR #870, opened 2019-03-22
allow for timing to be recorded and fix up runtests

plessbd:
Add the ability to record timings of the regression tests and update runtests to be a tiny bit smarter @jtpalmer if this makes it more difficult for you to refactor as part of your storage work let me know and I will kill it.


jtpalmer:
I'm currently working on a lot of changes to these files, but it shouldn't make anything more or less difficult for me.  I'm not familiar with how this timing data is used so if there is some documentation about that please let me know.

PR #873, opened 2019-03-25
Add Federated to About

plessbd:
Adds an about page for the federated module.

This page will also list the instances in a federation or the hub of the federation.

Screen Shot 2019-03-25 at 9 25 14 AM
Screen Shot 2019-03-25 at 9 21 02 AM
Screen Shot 2019-03-25 at 9 18 02 AM


smgallo:
We should change the text from "Imagine" because it implies that this is not real.

html/about/federated.php
```diff
@@ -0,0 +1,113 @@
+<?php
+namespace Xdmod\About;
+
+require_once __DIR__ . '/../../configuration/linker.php';
+use CCR\DB;
+?>
+<h1>Federated Open XDMoD</h1>
+<p>
+    A collection of independent XDMoD instances, each monitoring its own set of computing resources.
```
jpwhite4: Federated XDMoD is a module that allows multiple different XDMoD installations to have their data automatically duplicated and displayed together in a central location. This central XDMoD instance is known as a hub. This hub is able to view the data from all of the other instances.  The individual instances are independent of each other. Each instance may comprise very different underlying resources, configurations, and aggregation levels.

html/about/federated.php
```diff
@@ -0,0 +1,115 @@
+<?php
+namespace Xdmod\About;
+
+require_once __DIR__ . '/../../configuration/linker.php';
+use CCR\DB;
+?>
+<h1>Federated Open XDMoD</h1>
+<p>
+    Federated XDMoD is a module that allows multiple different XDMoD installations to have their data automatically duplicated and displayed together in a central location.
```
xdmod: How about this:

Federated XDMoD supports the collection and aggregation of data from a number of fully-functional
and individually managed XDMoD instances into a single federated instance of XDMoD icapable of
displaying federation-wide metrics. Each participating institution deploys an XDMoD instance through
which local data will be collected and also shipped to a central Federation Hub where it is
aggregated to provide a federation-wide view of the data.  Data particular to an individual center
is available from the Hub by applying filters and drill-downs.

PR #878, opened 2019-03-27
Update Cloud roles file fix roles file comparison

plessbd:
Made it so roles.d/cloud.json is updated properly
Fixed comparison to be based on the object not string


configuration/roles.d/cloud.json
```diff
@@ -26,8 +26,8 @@
                     "group_by": "vm_size_memory"
                 },
                 {
-                  "realm": "Cloud",
-                  "group_by": "submission_venue"
+                    "realm": "Cloud",
```
jpwhite4: What exactly is the configuration/roles.d/cloud.json file used for? It does not appear to be packaged in the RPM.

classes/OpenXdmod/Migration/Version800To810/ConfigFilesMigration.php
```diff
@@ -58,6 +58,81 @@ public function addCloudRolesGroupBy()
             foreach($cloudRolesFile['+roles'] as $key => $unused) {
                 $cloudRolesFile['+roles'][$key]['+query_descripters'][] = array('realm' => 'Cloud', 'group_by' => 'person');
                 $cloudRolesFile['+roles'][$key]['+query_descripters'][] = array('realm' => 'Cloud', 'group_by' => 'username');
+                if ($key === '+default') {
+                    $summary_charts = json_decode(
```
jpwhite4: Don't hardcode this stuff instead pull the contents directly from the template file.

configuration/roles.d/cloud.json
plessbd: SHHH, I have no idea...
@eiffel777 @chakrabortyr can you elaborate, is it safe for me to delete this file?

chakrabortyr: It should nominally be safe to do so, I believe everything is pulled from the templates folder now. However, we should test this to be sure.

smgallo:
I agree with @jpwhite4 and my preference is to follow a similar paradigm as system installs where if the file has been changed, we place a config.json.new file in the directory (it won't be picked up by the Configuration scanner) and let the user sort it out. Migrating configuration files should not be an area where we expend any significant effort.

plessbd:
I agree.

Do we currently have a way of knowing if they edited the file? Since all we have is the "current" file.

I would say some of these files should have an "enabled" key so that the file doesn't have to have objects deleted, just enabled or disabled, then we would be better able to do some of this.

classes/OpenXdmod/Setup/ResourcesSetup.php
```diff
@@ -195,8 +197,8 @@ private function rolesFileMatches($realm)
             return false;
         }
 
-        return file_get_contents(TEMPLATE_DIR . '/roles.d/' . $realm . '.json')
-            === file_get_contents($rolesFile);
+        return JSON::loadFile(TEMPLATE_DIR . '/roles.d/' . $realm . '.json', false)
+            === JSON::loadFile($rolesFile, false);
```
jtpalmer: Two different object instances will never be `===` (https://www.php.net/manual/en/language.oop5.object-comparison.php).

PR #881, opened 2019-04-02
Processor Buckets documentation update

plessbd:
The documentation for Processor Buckets was pointing to a file no longer used.  Updated the documentation and removed the unused file.


docs/configuration.md
```diff
@@ -47,13 +47,13 @@ don't want to use this process and would prefer to manually create the
...
+
+This step will run immediately after you have setup the database that XDMoD will
+be using and does not require any additional input. It is responsible for creating
 and populating the tables required by the Acl framework.
-      
-If your XDMoD Installation requires modifications to the acl tables 
-(etc/etl/etl_tables.d/acls/xdmod/<table>.json) then running this step again or 
+
+If your XDMoD Installation requires modifications to the acl tables
+(etc/etl/etl_tables.d/acls/xdmod/<table>.json) then running this step again or
```
jtpalmer: I thought we weren't including unrelated white space changes in pull request.  Are we allowing that now?

plessbd: unrelated I think is ok, but "just" whitespace changes is not.

This was an editor auto fix that I had not even noticed.  I can back it out if we want

smgallo: Typically I'd say do this in a separate "cleanup" commit. I'm OK with leaving it here since it actually does clean up unnecessary whitespace.

plessbd: as jeff pointed out we also have

https://github.com/ubccr/xdmod/blob/xdmod8.1/.editorconfig#L22-L23

```
[*.md]
trim_trailing_whitespace = false
```

So first I need to get my editor honoring that thing, and second, figure out why we have that and if it actually matters...

PR #841, opened 2019-03-06
No longer truncate aggregate tables

chakrabortyr:
## Description
NOTE: Still a bit of a WIP, just want the Shippable tests.

## Motivation and Context
We don't want to be constantly truncating + reingesting potentially millions of rows as cloud usage scales.

## Tests performed
Made sure existing tests passed + added new set of anonymised data. This anonymised data is then validated in a post ingest script. I manually did some additional validation with the following queries I couldn't think of a holistic way to operationalise:

This is an old instance with new events in the post ingest. Ran this before and after post_ingest_test.sh to ensure that it got updated with a new end_event/last_modified time:
`SELECT * FROM modw_cloud.cloud_events_transient where instance = 'DESIRED_ID';`

Then ran this to ensure that we only had two rows where the last_modified timestamp had changed:
`SELECT * FROM modw_cloud.cloud_events_transient order by end_time DESC, last_modified DESC;`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project as found in the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #882, opened 2019-04-02
Improve storage aggregation

jtpalmer:
## Description

Add use of last modified start date to storage realm aggregation.

## Motivation and Context

These changes allow aggregation of new storage data without aggregating all storage data.  The changes follow the same pattern as the cloud realm (see #841).

## Tests performed

Added new test to check the number of periods aggregated after ingesting data for a single day (after data for multiple days has already been ingested and aggregated).

## Types of changes

## Checklist:


open_xdmod/modules/xdmod/regression_tests/post_ingest_test.sh
```diff
@@ -34,3 +34,19 @@ then
...
+last_modified_start_date=$(date +'%F %T')
+sudo -u xdmod xdmod-ingestor --datatype storage
+agg_output=$(mktemp --tmpdir storage-aggregation-XXXXXXXX)
+sudo -u xdmod xdmod-ingestor --aggregate=storage --last-modified-start-date "$last_modified_start_date" | tee $agg_output
+for unit in day month quarter year; do
+    if ! grep -q "unit: $unit, periods: 1," $agg_output; then
+        echo Did not aggregate 1 period of storage data for unit $unit
+        exit 1
+    fi
+done
```
chakrabortyr: More of a genuine question than a comment, are there any other edge cases we'd need to test for in the upgrade path?

jtpalmer: There are probably edge cases we aren't testing, but this test is just intended to check one specific case.

PR #894, opened 2019-04-17
Update SAML ocs to be a bit more clear

plessbd:
Clarifying the SAML docs a bit.

### DIFF
--- a/docs/simpleSAMLphp.md
+++ b/docs/simpleSAMLphp.md
@@ -126,7 +126,14 @@ $config = array(
 );
 ```
 
-Use the above example to match the response from the IdP, the indexes under `authproc[40]` point to indexes expected by Open XDMoD.
+Use the above example to match the response from the IdP, the indexes (left side of `=>`) under `authproc[40]` point to values (right side of `=>`) expected by Open XDMoD.
+These indexes need to be what SAML receives from the IdP not the ones known from LDAP or other sources.
+
+These indexes can be viewed using the `Test authentication sources` page built into simplesamlphp located at
+`https://<hostname>/simplesaml/module.php/core/authenticate.php` then clicking on `xdmod-sp` after which you will be redirect to the configured IdP login page.
+After you have successfully logged in to the IdP you should be redirected back to a `SAML 2.0 SP Demo Example` page.  This page will show you the attributes passed back by the IdP.
+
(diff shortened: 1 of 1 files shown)


docs/simpleSAMLphp.md
```diff
@@ -126,7 +126,17 @@ $config = array(
 );
 ```
 
-Use the above example to match the response from the IdP, the indexes under `authproc[40]` point to indexes expected by Open XDMoD.
+Use the above example to match the response from the IdP, the indexes (left side of `=>`) under `authproc[40]` point to keys (right side of `=>`) expected by Open XDMoD.
+
+These indexes need to be what SAML receives from the IdP not the ones known from LDAP or other sources.
+
+These can be viewed by using the Test authentication sources page built into simplesamlphp located at
```
jpwhite4: Perhaps put "test authentication sources" in quotes or in fixed font. Also check capitalization of the word Test.

docs/simpleSAMLphp.md
```diff
@@ -126,7 +126,17 @@ $config = array(
...
+`https://<hostname>/simplesaml/module.php/core/authenticate.php` then clicking on `xdmod-sp` and it should redirect you to the configured IdP login page.
+After loggin in you should then be redirected back to a `SAML 2.0 SP Demo Example` page.
```
jpwhite4: Is loggin a word?

docs/simpleSAMLphp.md
```diff
@@ -126,7 +126,17 @@ $config = array(
 );
 ```
 
-Use the above example to match the response from the IdP, the indexes under `authproc[40]` point to indexes expected by Open XDMoD.
+Use the above example to match the response from the IdP, the indexes (left side of `=>`) under `authproc[40]` point to keys (right side of `=>`) expected by Open XDMoD.
+
+These indexes need to be what SAML receives from the IdP not the ones known from LDAP or other sources.
+
```
jpwhite4: No new paragraph here.

docs/simpleSAMLphp.md
plessbd: in some places I am pretty sure we use it as loggin

docs/simpleSAMLphp.md
plessbd: Will add back-ticks the page title case when displayed in the browser matches the case here.

PR #896, opened 2019-04-18
moving tests

plessbd:
`tests` is a wonderful directory for things such as regression, integration, ui, and unit tests to be located.

so I moved them there

`automated_tests` is a dumb name (thanks @plessbd) these have been renamed to `ui`
no need to have `_tests` in the folder names anymore since these are now properly located in a test directory

`tests/ci`
-  `.secrets.json` isnt a secret to anyone so created that under here so it doesnt ahve to be copied around
-  the scripts needed for ci are now here instead of under integration_tests/scripts


plessbd:
I am going to break this up more (again...)

plessbd:
replaced by #897

PR #897, opened 2019-04-19
Relocate Tests

plessbd:
The first commit is just a bunch of `git mv` folders.
```bash
git mv open_xdmod/modules/xdmod/regression_tests tests/regression
git mv open_xdmod/modules/xdmod/automated_tests/ tests/ui
git mv open_xdmod/modules/xdmod/component_tests/ tests/component
git mv open_xdmod/modules/xdmod/tests/ tests/unit
git mv open_xdmod/modules/xdmod/integration_tests/ tests/integration
git mv tests/xdmod-test-artifacts/xdmod tests/artifacts/xdmod
git mv tests/artifacts/xdmod/automated_tests tests/artifacts/xdmod/ui
```
The second is where the actual changing of files takes place.

The third commit is cleaning up the `artifacts/xdmod/` json files
```bash
for file in `find . -type f -name '*.json'`; do echo $file && jq . -- "$file" >| /tmp/cleaned.json && \mv -f /tmp/cleaned.json "$file"; done
```

fourth commit fixes one last travis error.

`tests` is a wonderful directory for things such as regression, integration, ui, and unit tests to be located.

so I moved them there

`automated_tests` is a dumb name (thanks @plessbd) these have been renamed to `ui`
no need to have `_tests` in the folder names anymore since these are now properly located in a test directory

`tests/ci`
-  `.secrets.json` isnt a secret to anyone so created that under here so it doesnt ahve to be copied around
-  the scripts needed for ci are now here instead of under integration_tests/scripts


tests/component/lib/Utilities/LogAndThrowExceptionTest.php
```diff
@@ -45,7 +45,7 @@ public function testNoExceptionCode()
     public function testExceptionCode()
     {
         $msg = "Code = 10";
-        $excpectedMsg = (string) $this->loggable . " $msg";
+        $expectedMsg = (string) $this->loggable . " $msg";
```
jpwhite4: How did this test pass before?

tests/integration/lib/Controllers/BaseUserAdminTest.php
```diff
@@ -210,7 +210,7 @@ protected function createUser(array $options)
         $institution = isset($options['institution']) ? $options['institution'] : -1;
         $firstName = isset($options['first_name']) ? $options['first_name'] : 'Test';
         $lastName = isset($options['last_name']) ? $options['last_name'] : 'User';
-        $emailAddress = isset($option['email_address']) ? $options['email_address'] : $username . self::DEFAULT_EMAIL_ADDRESS_SUFFIX;
+        $emailAddress = isset($options['email_address']) ? $options['email_address'] : $username . self::DEFAULT_EMAIL_ADDRESS_SUFFIX;
```
jpwhite4: How did this test pass before?

tests/ui/wdio.conf.js
```diff
@@ -218,7 +218,7 @@ exports.config = {
             if (webdriverHelpers.hasOwnProperty(extension) && typeof webdriverHelpers[extension] === 'function') {
                 browser.addCommand(extension, webdriverHelpers[extension].bind(browser));
             } else {
-                console.warning('Not Adding: ', extension, typeof webdriverHelpers[extension]);
+                process.stderr.write('Not Adding: ' + extension + typeof webdriverHelpers[extension]);
```
jpwhite4: Note that this code will produce different output to the previous code.

tests/ci/README.md
```diff
@@ -0,0 +1,50 @@
...
+docker build -t centos7:xdmod-base -f Dockerfile.centos7:xdmod-base .
+```
+
+## Cache and Config Files
+We use some default configuration and cache files to set some defaults for
+development and speed up the build times of Docker.
+If you look in `assets/copy-caches.sh` you will see all of the files that will be put into place.
+
+### vimrc
+This is just some defaults for VIM when useing the docker for development
```
jpwhite4: s/useing/using/

tests/component/lib/Utilities/LogAndThrowExceptionTest.php
plessbd: Magic?

tests/integration/lib/Controllers/BaseUserAdminTest.php
plessbd: Magic?

tests/ui/wdio.conf.js
plessbd: yes, however, this is for the most part only for "developers" when running this and should rearely if ever be hit.

PR #899, opened 2019-04-24
Update to 8.5 and 8.1 based shippable

plessbd:
Update to a new version and new docker base build


open_xdmod/modules/xdmod/integration_tests/email-subject-test.sh
```diff
@@ -5,6 +5,13 @@ curl "${host}controllers/mailer.php" --compressed -H 'X-Requested-With: XMLHttpR
 
 sed -i -- 's/subject_prefix = "SEND BEN MONEY"/subject_prefix = ""/' /etc/xdmod/portal_settings.ini
 
+# sometimes it takes a second to send the emails
+queueLength=`postqueue -p | tail -n1 | awk '{print $5}'`
+if [ "$queueLength" != ""  ]; then
```
jtpalmer: Please change this to a while loop.  There's no guarantee that the queue will be emptied after 1 second.

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
```diff
@@ -39,7 +39,7 @@ then
     sudo -u xdmod xdmod-ingestor --aggregate=storage --last-modified-start-date "$last_modified_start_date"
     sudo -u xdmod xdmod-import-csv -t names -i $REF_DIR/names.csv
     sudo -u xdmod xdmod-ingestor
-    php /root/bin/createusers.php
+    php $BASEDIR/../../../../../tests/ci/scripts/create_xdmod_users.php
     # This will ensure that the users created in `/root/bin/createusers.php`
```
jpwhite4: Need to update this comment. Also is it stil necessary to run xdmod.acls-import pipeline with the new create_xdmod_users script? If so then why is it not done automatically in the create_xdmod_users script? If not then remove.

open_xdmod/modules/xdmod/integration_tests/email-subject-test.sh
```diff
@@ -5,6 +5,16 @@ curl "${host}controllers/mailer.php" --compressed -H 'X-Requested-With: XMLHttpR
 
 sed -i -- 's/subject_prefix = "SEND BEN MONEY"/subject_prefix = ""/' /etc/xdmod/portal_settings.ini
 
+# sometimes it takes a second to send the emails
+queueLength=`postqueue -p | tail -n1 | awk '{print $5}'`
+count=0
+while [[ "$queueLength" != ""  && $count < 5 ]]; do
```
jtpalmer: If you're going to add a count then you should also exit with failure if the max count is reached and output a message explaining that the test failed due to the count (i.e. that the queue never emptied) and not because the email subject wasn't found (because the email was never received).

Also, `-lt` should be used in place of `<` for integer comparisons.  (It doesn't cause a problem when you're using `5`, but if you changed that to `10` it wouldn't work as expected.

```suggestion
while [[ "$queueLength" != ""  && $count -lt 5 ]]; do
```

shippable.yml
```diff
@@ -3,14 +3,14 @@ env:
     - XDMOD_TEST_MODE=fresh_install
     - XDMOD_TEST_MODE=upgrade
 build:
-    cache: true
+    cache: false
```
jpwhite4: Why the change to the cache setting? Is this leftover debug stuff?

plessbd: yes left over from debug

open_xdmod/modules/xdmod/integration_tests/scripts/bootstrap.sh
plessbd: @ryanrath I am not about this,  do you know for sure.

All I know is the "new" script is the one @smgallo created that had a bit more customization to it.  I will look into this.

PR #907, opened 2019-05-02
Version 8.1.1 updates

plessbd:
Updates for version 8.1.1

### DIFF
--- a/html/about/release_notes/xdmod.html
+++ b/html/about/release_notes/xdmod.html
@@ -2,6 +2,15 @@
 
 <p>Below is a list of Open XDMoD releases with major features and bug fixes listed.</p>
 
+<h2> Version 8.1.1 (2019-05-02) </h2>
+<h3 class="bug-fixes">Bug Fixes</h3>
+<ul>
+    <li>General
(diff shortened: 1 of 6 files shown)


jtpalmer:
Update the version in `configuration/portal_settings.ini`.

html/about/release_notes/xdmod.html
```diff
@@ -2,6 +2,15 @@ <h1>Open XDMoD Release Notes</h1>
 
 <p>Below is a list of Open XDMoD releases with major features and bug fixes listed.</p>
 
+<h2> Version 8.1.1 (2019-05-02) </h2>
+<h3 class="bug-fixes">Bug Fixes</h3>
```
jtpalmer: Is `class="bug-fixes"` correct?  Should the `id`s be removed from the sections for 8.1.0?

plessbd: I am not 100% sure yet.  I put it as a class since you cant have duplicate id's. I am good with removing them, I am not sure why steve put them in as id's to be honest.

PR #910, opened 2019-05-02
Merge xdmod81 into xdmod85

jtpalmer:


jpwhite4:
Quick question: What should we call the migration classes are they still ./classes/OpenXdmod/Migration/Version810To850/DatabasesMigration.php or do we have to add a new Version811To850 migration?

jtpalmer:
They need to be renamed `Version811To850`. I'll change them now.

PR #911, opened 2019-05-02
Fix Query Descriptor Visibility

ryanrath:
**NOTE: I apologize in advance for the large number of files in this PR. On the plus side, the vast majority of them are test artifact updates.**

## Description
Updated to allow query descriptors to control visibility

**acl-config**
Previously we only allowed statistics to determine if an `acl_group_by` record
was visible or not. But our config files seem to indicate that certain query
descriptors should be able to control their ability to be shown as well. To
support both of these use cases the `processQueryDescriptors` function in
`acl-config` has been updated to take into account the `hide` property defined
at the query descriptor level and the `visible` property defined at the
statistic level. The logic used to determine the final visibility of an
`acl_group_by` record is as follows.

- By default all entries are shown
- If `visible` is defined then we use it's value.
- Else, if `hide` is defined then we use it's value.

Also, there was a minor refactor with the removal of the `$enabled` variable.
The `$disable` variable that it was based on was never going to be null, so it's
use could be replaced w/ just `!$disable`.

**Acls**
There was some minor cleanup done because it was determined that we always want
to have the `visible` column included, not just when a `$statisticName` is
provided.

This allowed us to move the `$descripter->setShowMenu(...)` from out of the
`if(isset($statisticName)) { ... }` guard so that all query descriptors
generated by this function will have their `showMenu` property populated.

These two sets of changes work together to allow us to control our UI state as
we expect to.

## Motivation and Context
Being able to control the visibility of the group bys / statistics via ACLs is expected. This was specifically a problem on an XSEDE install due to the more complicated visibility requirements. In OpenXDMoD there are no group by's / statistics that are hidden via the ACLs so this was not caught there. 

## Tests performed
Manual Testing was performed:
  - Navigate to an XSEDE Dev install, for each of the following users perform the steps below: pub, usr, cc
  - Select the `Usage` Tab. 
  - Ensure that the realms / group by's that are shown match user is allowed to see ( `etc/roles.d/pub.json#query_descripters`
      - Users and which entries should be hidden: 
          - *pub*: Accounts
          - *usr*: Accounts, Allocations
          - *cc*: Accounts, Allocations

## Types of changes

## Checklist:


tests/artifacts/xdmod/acls/output/get_disabled_menus-centerdirector.json
```diff
@@ -0,0 +1,3 @@
+[
```
plessbd: is there a way we could combine all these files that are just empty arrays?  Do they even really need to be files?

tests/artifacts/xdmod/etlv2/dataendpoint/input/directory_scanner/euca_2017-06-01.json
```diff
@@ -1,120 +1,4 @@
...
-      "create_time": "2017-03-21T16:57:45.330Z",
-      "user": "riveraj",
-      "id": "vol-dae393e0",
-      "size": "10"
-    }
-  ],
-  "private_ip": "172.17.47.126",
-  "root_type": "ebs"
-}
+{"node_controller": null, "public_ip": null, "account": "000048934329", "event_type": "STATE_REPORT", "event_time": "2017-05-16T03:55:04Z", "instance_type": {"name": "c1.medium", "cpu": "4", "memory": "16384", "disk": "40", "networkInterfaces": "2"}, "image_type": "emi-521695e8", "instance_id": "i-cb13943e", "record_type": "ADMINISTRATIVE", "block_devices": [{"account": "big", "attach_time": "2017-04-19T13:47:38.609Z", "backing": "ebs", "create_time": "2017-04-19T13:47:38.550Z", "user": "tyearke", "id": "vol-6a9b5bc2", "size": "40"}], "private_ip": null, "root_type": "ebs"}
```
plessbd: is this reverting my cleanup because the test actually compares a string or is there a reason this isnt pretty printed?

smgallo: Pretty-printing broke the input data. This is sample Euca data and is expected to be JSON objects one per line separated by a newline.

tests/artifacts/xdmod/acls/output/get_disabled_menus-centerdirector.json
ryanrath: Is there a way? Absolutely. But it does mean that we already have per user expected test results. i.e. They're the same expected results now, but will they always be the same? Do they need to be files? It would be nice / it keeps the test logic simple.

plessbd: In the regression tests a "reference" is used as the base and it only creates the other files if they are different from the reference.  Just something to keep in mind in the future.

ryanrath: Do the regression tests provide this functionality for other test types?

plessbd: not that I did...
Though that could be a good "base" thing that all the tests could use...

PR #920, opened 2019-05-07
Updated SQL construction for `tablesExist`

ryanrath:
## Description
So it turns out that PHP prepared statement variable binding doesn't play well
with SQL `IN` statements. Just moving things around a bit so that instead of
supplying the `IN` clause w/ a bind parameter we instead just populate it in the
base query string.

This PR is related to: https://ubccr.freshdesk.com/a/tickets/12470

## Motivation and Context
It's nice to have a `recover` mode that actually functions

## Tests performed
Manual Tests performed on a local docker instance. 

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -2127,12 +2127,6 @@ function verifyDatabase()
 function tablesExist(iDatabase $db, array $tables, $schema, $all = true)
 {
     $handle = $db->handle();
-    $query = <<<SQL
-SELECT *
-FROM information_schema.tables
-WHERE table_schema = :table_schema
-AND table_name in (:table_list);
-SQL;
 
     $tables = array_reduce(
         $tables,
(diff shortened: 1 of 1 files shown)


bin/acl-config
```diff
@@ -2127,12 +2127,6 @@ function verifyDatabase()
 function tablesExist(iDatabase $db, array $tables, $schema, $all = true)
```
chakrabortyr: Could we potentially readapt this to work with the tableExists function in the MySQLHelper class? This could be moved over there as well in theory.

ryanrath: It certainly could be moved. If we do then I'd like to see a separate consolidation PR for that though. I say *if* because I ultimately see `acl-config` going away, to be replaced w/ some ETLv2 pipelines.

PR #926, opened 2019-05-10
Less stringent DataWarehouse Descripter tests

plessbd:
We don't really care "what" comes back.

As we shouldnt have to change the tests for verbiage updates, a PR updating those will contain the discussions on how it should say, the tests just verify that something comes back.

We care that a list of realms with dimensions and metrics comes back and that each dimension and metric has the expected information.


tests/artifacts/xdmod/schema/dw_descripter.spec.json
```diff
@@ -0,0 +1,126 @@
...
+            "title": "realm"
+        },
+        "dimensions": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "object",
+                "$ref": "#/definitions/dimension"
+            },
+            "minProperties": 1,
+            "title": "Dimesnions"
```
ryanrath: s/Dimesnions/Dimensions

tests/integration/lib/Controllers/UserAdminTest.php
```diff
@@ -421,33 +417,46 @@ public function testGetDwDescripters(array $user)
...
-            $this->getTestFiles()->getFile('user_admin', $expectedFileName, 'output')
+        $schemaObject = JSON::loadFile(
+            $this->getTestFiles()->getFile('schema', 'dw_descripter.spec', ''),
+            false
         );
-
-        $this->assertEquals($expected, $actual, "[$username] Get Data Warehouse Descripters - Expected:\n\n" . json_encode($expected) . "\n\nReceived:\n\n" . json_encode($actual));
-
+        $validator = new Validator();
+        $validator->check(json_decode(json_encode($actual)), $schemaObject);
```
ryanrath: encode/decode? Are there problems w/ just using `$schemaObject`?

plessbd: it has to be an object instead of an associative array

PR #931, opened 2019-05-20
Improve DirectorySanner to properly handle first file being empty

smgallo:
## Description

The DirectoryScanner data endpoint now properly handles the case where the first file scanned is empty or otherwise contains no valid records. Previously, the first record (which is used to determine field names) was set even if the first file had no data and this caused an error.

## Motivation and Context

Bugfix

## Tests performed

Added new test, manual testing in docker.

## Types of changes

## Checklist:


classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -98,7 +98,7 @@ protected function _execute()
         // If there are no records we can bail out. Otherwise we may not even be able to infer the
         // source field names.
 
-        if ( 0 == $this->sourceEndpoint->count() ) {
+        if ( false === $firstRecord || 0 == $this->sourceEndpoint->count() ) {
```
plessbd: should this be ===

tests/component/lib/ETL/IngestorTest.php
```diff
@@ -154,10 +182,16 @@ private function executeOverseerAction($action, $localOptions = "")
...
         // Note that tests are run in the directory where the PHP class is defined.
         $overseer = realpath(BASE_DIR . '/tools/etl/etl_overseer.php');
-        $configFile = realpath(BASE_DIR . '/tests/artifacts/xdmod/etlv2/configuration/input/xdmod_etl_config_8.0.0.json');
-        $options = sprintf('-c %s -a %s %s -v warning', $configFile, $action, $localOptions);
-        $command = sprintf('%s %s', $overseer, $options);
+        $configFile = realpath(BASE_DIR . self::TEST_INPUT_DIR . '/xdmod_etl_config_8.0.0.json');
+        $options = sprintf('-c %s -a %s %s', $configFile, $action, $localOptions);
+
+        // Add a verbosity flag if the local options do not already contain one
+        if ( "" == $localOptions || false === strpos($localOptions, '-v') ) {
```
plessbd: should this be ===

classes/ETL/Ingestor/StructuredFileIngestor.php
smgallo: It is already `===`

tests/component/lib/ETL/IngestorTest.php
smgallo: Do you mean `"" == $localOptions`? What does that add?

ryanrath: Just that it would fail if `$localOptions` was ever anything other than a `string`. Don't know if that's a concern or not though.

smgallo: Since we have complete control over the tests and the default is a string if not specified I think we're fine with a `==`.

ryanrath: Sounds good to me

smgallo:
If we do need to identify a specific record, that would be handled by the ingestor. The DirectoryScanner is just an endpoint for reading the files.

ryanrath:
*facepalm* I may have missed that those are test artifacts... Sorry 'bout that. All good :)

PR #943, opened 2019-05-23
Novice User Interface - Center Report Card: Support

ryanrath:
## Description
The changes made in this PR are in support of the `novice_appkernels` branch of the `appkernels` module. Specifically: 
  - `classes/Rest/Controllers/SummaryControllerProvider.php`: 
    - While testing I ran into problems w/ the `$queryConfig->featured` property not being present. Which caused exceptions to be thrown. So I just added a quick guard to make sure that the property was present prior to inspecting its value. 
  - `classes/XDUser.php`:
    - The `getResources` function is utilized in the main `novice_appkernels` PR in the `appkernels` module, but as it made sense to include it in `XDUser.php` the change was made here. 
  - `html/gui/js/Viewer.js`:
    - Eslint flagged this file for some reason so I just went through and reformatted it according to our current rules. 
  - `html/index.php`:
    - These files are required by components included in the `appkernels` module which were moved from the internal dashboard to the main interface. 

## Motivation and Context
To get OpenXDMoD ready to support the `CenterReportCardPortlet` as well as host the `AppKernelPerformanceMap` from the internal dashboard. 

## Tests performed
Manual Testing was performed in conjunction w/ the App Kernels `novice_appkernels` branch installed. These changes are live on my dev port [9006]. 

## Types of changes

## Checklist:


html/gui/js/Viewer.js
```diff
@@ -10,479 +10,429 @@
...
+            border: false,
+            activeTab: 0,
+            region: 'center',
+            enableTabScroll: true,
+            defaults: {
+                tabCls: 'tab-strip'
+            },
+
+            listeners: {
+                // eslint-disable-next-line no-shadow
```
plessbd: part of me wants to just fix this instead of disabling the linter...

html/gui/js/Viewer.js
```diff
@@ -10,479 +10,429 @@
...
+                    var token = null;
+                    if (!hasActiveTab && hasTab) {
+                        CCR.xdmod.ui.activeTab = tab;
+
+                        XDMoD.TrackEvent('Tab Change', tab.title);
+
+                        token = tabPanel.id + CCR.xdmod.ui.tokenDelimiter + tab.id;
+                        Ext.History.add(token);
+                        // eslint-disable-next-line no-mixed-operators
+                    } else if (hasActiveTab && hasTab && tab.id !== CCR.xdmod.ui.activeTab.id ||
```
plessbd: part of me wants to just fix this instead of disabling the linter...

html/gui/js/Viewer.js
```diff
@@ -10,479 +10,429 @@
...
+                // Ext.menu.MenuMgr.hideAll();
+                var tokenType = typeof token;
+                var parts = tokenType === 'string' ? CCR.tokenize(token) : token;
+                if (!parts) {
+                    return;
+                }
 
-      if (token) {
+                var root = CCR.exists(parts.root) ? parts.root : 'main_tab_panel';
+                // eslint-disable-next-line no-shadow
```
plessbd: part of me wants to just fix this instead of disabling the linter...

html/gui/js/Viewer.js
```diff
@@ -10,479 +10,429 @@
...
+                }
+            }); // viewStore.on('load',...
+        }, this, {
+            single: true
+        }); // mainPanel.on('render',…
+
+        CCR.xdmod.ui.Viewer.viewerInstance = this;
+    }, // initComponent
+    getParameterByName: function (name, source) {
+        // eslint-disable-next-line no-param-reassign,no-useless-escape
```
plessbd: part of me wants to just fix this instead of disabling the linter...

html/internal_dashboard/index.php
```diff
@@ -189,10 +189,8 @@
   <script type="text/javascript" src="js/UsersSummary/Store.js"></script>
   <script type="text/javascript" src="js/UsersSummary/Portlet.js"></script>
 
-  <link rel="stylesheet" type="text/css" href="../gui/lib/extjs/examples/ux/css/ColumnHeaderGroup.css"/>
```
plessbd: why are these removed?  Is this actually related to this change?  Not sure we should remove it as part of this if it isnt related, it should be its own cleanup PR

classes/Rest/Controllers/SummaryControllerProvider.php
```diff
@@ -115,7 +114,7 @@ public function getPortlets(Request $request, Application $app)
 
                     $queryConfig = json_decode($query['config']);
 
-                    if (!$queryConfig->featured) {
+                    if (isset($queryConfig->featured) && !$queryConfig->featured) {
```
jpwhite4: The code should only include charts that have featured set to true. Otherwise they should be skipped.

```suggestion
                    if (!isset($queryConfig->featured) || !$queryConfig->featured) {
```

classes/XDUser.php
```diff
@@ -2606,4 +2606,48 @@ public function isSticky()
     {
         return $this->sticky;
     }
+
+    /**
+     * Retrieves the resources that this user has access to.
```
jpwhite4: I don't think this function belongs in the XDUser class. It also circumvents the ACL tables - which surely is the authoritative source of user access information. Why not build on the existing  getDimensionValues functionality in the Warehouse Controller? This already has the access control code and use the GroupBy classes to acces the data in the datawarehouse.

ryanrath: \*nod\* It does circumvent the ACL tables. I'll look at the `getDimensionValues` function for a re-write. As an explanation, the results of this function are used in constructing a `\AppKernel\PerformanceMap` object which utilizes the ACL tables to restrict the final output by the resources the requesting user has access to. But those changes are in the AppKernels module PR and I didn't reference it specifically. Thanks again for the catch, I'll have an update incoming shortly.

PR #946, opened 2019-05-24
Allow search to be run without extra parameters

jpwhite4:
## Description

update the job viewer search endpoint so that it is permissible to run a search with no extra
filters.

## Motivation and Context
This change is needed for the novice user JobPortal. This will run a filterless search which will
return all jobs that the user has access to.

## Tests performed
Added a load of extra tests to make sure the input data are validated and added a test for the no parameter case.

### DIFF
--- a/tests/integration/lib/Rest/JobViewerTest.php
+++ b/tests/integration/lib/Rest/JobViewerTest.php
@@ -219,21 +219,113 @@ class JobViewerTest extends \PHPUnit_Framework_TestCase
         $this->xdmodhelper->logout();
     }
 
-    public function testInvalidJobSearchMissingParams() {
+    public function missingParamsProvider() {
 
-        $searchparams = array(
-            'realm' => 'Jobs',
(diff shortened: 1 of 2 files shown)


jpwhite4:
Also added a minor optimization where the non-priv query adds a limit to reduce the amount of data returned.

tests/integration/lib/Rest/JobViewerTest.php
```diff
@@ -219,21 +219,113 @@ public function testInvalidJobSearchJson() {
...
+            true
+        );
+
+        return $tests;
+    }
+
+    /**
+     * @dataProvider missingParamsProvider
+     */
+    public function testInvalidJobSearchMissingParams($xdmodhelper, $searchparams, $isfinal) {
```
plessbd: I am not sure but could you set `$isFinal = false` and then only pass in `true` on the last one?
Or should we be using `tearDown()` or `tearDownAfterClass()` to do the logout?

PR #947, opened 2019-05-28
Add recent charts and reports porlet

mkzia:
## Description
This PR adds the recent charts and report portlet. The following files were changed:
  - `classes/Rest/Controllers/SummaryControllerProvider.php` was updated to add a new REST endpoint `recentchartsreports`. This endpoint returns a user's recent charts and reports in JSON format. 
  - `configuration/etl/etl_tables.d/xdb/reports.json`  was updated to modify the `Reports` table to include the last modified date.
  - `configuration/etl/etl.d/xdmod-migration-8_1_2-8_5_0.json` was added to add the last modified field to the `Reports` table when migrating. 
  - `classes/XDReportManager.php` was updated to return the newly added`last_modified` field when `fetchReportTable` is called. 
  - ` configuration/assets.json` was updated to include the this portlet
  - `html/gui/js/modules/ReportGenerator.js` was updated to add a new listener called `load_report`. This listener is used to fire an event that selects the report clicked in the recent charts and reports portlet. 
  - `html/gui/js/report_builder/ReportsOverview.js` was updated to add a new listener called `load_report` that opens the report selected by the `load_report` listener in `html/gui/js/modules/ReportGenerator.js`. 
  - `html/gui/js/modules/summary/RecentChartsReportsPortlet.js` was added to render the recent charts and reports portlet. 

## Motivation and Context
To make it easier for selected users to quickly access their recent charts and reports. 

## Tests performed
Manually tested the functionality. Have plans to add integrations tests. 

## Types of changes

## Checklist:


html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,138 @@
...
+
+XDMoD.Modules.SummaryPortlets.RecentChartsReportsPortlet = Ext.extend(Ext.ux.Portlet, {
+
+    layout: 'fit',
+    autoScroll: true,
+    title: 'Recent Charts and Reports',
+    width: 1000,
+
+    initComponent: function () {
+        var aspectRatio = 0.8;
```
jpwhite4: please make the aspect ratio 11/17 to match the rest of the charts.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,138 @@
...
+                        } else if (r.data.type === 'Report') {
+                            CCR.xdmod.ui.reportGenerator.fireEvent('load_report', r.data.report_id);
+                        }
+                    }
+                }
+            })
+        });
+        this.height = this.width * aspectRatio;
+        this.items = [this.chartReportGrid];
+        this.chartReportStore.reload();
```
jpwhite4: Why the reload() call?

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,138 @@
...
+        this.chartReportStore.sort('ts', 'DESC');
```
jpwhite4: Why the explicit sort call? What happens if you put the sortInfo in the store's configuration?

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,138 @@
...
+        XDMoD.Modules.SummaryPortlets.RecentChartsReportsPortlet.superclass.initComponent.apply(this, arguments);
+    },
+    listeners: {
+        duration_change: function (timeframe) {
+            this.chartReportStore.reload();
+            this.chartReportStore.sort('ts', 'DESC');
+            this.chartReportGrid.getView().refresh();
```
jpwhite4: Is this view refresh needed? Doesn't the store load trigger the view refresh automatically?

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
mkzia: Changed

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
mkzia: Removed reload and put load after store definition.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
mkzia: Put sort config in store.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
mkzia: Removed reload.

mkzia: Removed listener since it is redundant.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,137 @@
...
+    width: 1000,
+
+    initComponent: function () {
+        var aspectRatio = 11 / 17;
+
+        this.chartReportStore = new Ext.data.JsonStore({
+            // store configs
+            autoDestroy: true,
+            url: XDMoD.REST.url + '/summary/recentchartsreports',
+            storeId: 'chartReportStore',
```
jpwhite4: Suggest not setting a storeId since this causes  the store to be registered with the store manager and you do not use (not need to use) the store manager.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
```diff
@@ -0,0 +1,137 @@
...
+                    }
+                }
+            ],
+            sortInfo: {
+                field: 'ts',
+                direction: 'DESC'
+            }
+
+        });
+        this.chartReportStore.load();
```
jpwhite4: FYI. you can set autoLoad: true in the properties and then you don't need the explicit load() call.

mkzia: Explicit load has been replaced with autoLoad property.

html/gui/js/modules/summary/RecentChartsReportsPortlet.js
mkzia: StoreId has been removed.

PR #951, opened 2019-05-29
Fixed loading the organization abbrevation from organization.json

pphysch:
## Description
Fixed the web portal always loading the organization's full name as its abbrevation, rather than the abbrevation provided by the admin during setup. Both strings are stored in the `organization.json` config file.

## Motivation and Context
XDMoD allows the administrator to enter an abbrevation for their organization name, which may otherwise be lengthy ("CHPC" vs. "Center for High Performance Computing"). However, XDMoD has not been using the abbrevation for anything (except in `/xdmod/classes/XDReportManager.php`), perhaps because it has not been loaded correctly. This change fixes only the incorrect loading but does not change how the name/abbrev variables are used.

In future PRs it would be nice to change some instances of ORGANIZATION_NAME to ORGANIZATION_NAME_ABBREV in the web portal. For example, `/xdmod/classes/DataWarehouse/Query/Jobs/Statistics/UtilizationStatistic.php` includes 6 instances of the full name in its description paragraph, and the usage of the full name in the main statistics menu may lead to cropping on smaller monitors (especially when nesting queries). Most of these could be changed to the shorter abbrevation without losing any semantic clarity.

## Types of changes

## Checklist:


jpwhite4:
The `ORGANIZATION_NAME_ABBREV` is used to set the javascript variable `CCR.xdmod.org_abbrev` which in turn is used in the report generator to set the `chart_drill_details`. It is not obvious to me that this change will not have any side effects such as breaking the report generator for existing saved charts (that were saved in previous versions of the software). Have you done any testing / validation of the report generator with this change?

pphysch:
Thanks for your response.

I created a fresh XDMoD instance and a couple reports and verified they had empty `chart_drill_details` values in `moddb.ReportCharts`. With my name and abbrev as "Center for High Performance Computing" and "CHPC" respectively, I am able to see "Center for High Performance Computing" (i.e. `ORGANIZATION_NAME_ABBREV`) rendered as the subtitle in the report Preview, PDF, and Word Docs.

Then I applied this change live to the instance and the report Preview, PDF and Doc are now showing "CHPC" as the subtitle after the old reports are newly generated, as expected, with no visible errors.
[before.pdf](https://github.com/ubccr/xdmod/files/3238562/before.pdf)
[after.pdf](https://github.com/ubccr/xdmod/files/3238561/after.pdf)

As far as I understand, this is the only current usage of `ORGANIZATION_NAME_ABBREV`; the subtitle for reports that have empty `chart_drill_details`. It's not clear to me that the the abbrevation is even appropriate here over the full name. On both PDF and Word Document, the "subtitle" font size is much smaller than the word title and there is plenty of room for the full organization name, which is otherwise not present on the report unless included in one of the fields. 

Therefore, I would further suggest that the existing usage of `ORGANIZATION_NAME_ABBREV` in the `classes/XDReportManager.php` be changed to `ORGANIZATION_NAME`, as both variables are already using identical string values (hence this PR). With this additional change there would be no change whatsoever to report content after a merge, so existing downloaded reports would have the same subtitle content as newly generated reports. This has the positive side effect of removing all current references to `ORGANIZATION_NAME_ABBREV` from the source so it may be easier to decide where to go with the organization abbrevation string. If this makes sense, I can add another commit to this PR and test it.

PR #952, opened 2019-06-06
Enable APCu cache for Configuration objects

smgallo:
## Description

The GroupBy refactor will see data that is currently stored in PHP code moved to configuration files parsed via the `Configuration` classes. While these classes provide additional functionality such as references, include files, and comments, there is a performance overhead for providing these features. This work is an effort to reduce that overhead as well as improve overall performance.

Parsed `Configuration` objects are now cached using APCu to improve performance. [APCu](https://www.php.net/manual/en/intro.apcu.php) is a user-space in-memory cache for PHP that uses shared memory to cache data between requests. This cache is not persistent and is cleared on a restart of the web server but for our purposes this is OK as the cache will be reloaded automatically as calls are made. Note that on the first call, performance will be the same as the local configuration object cache as files are parsed and the cache is seeded but on subsequent calls `Configuration` objects will be pulled from the cache with only `Configuration::factory()` and `Configuration::__construct()` being called.

Changes to global or local configuration files are automatically detected by scanning the global and local configuration files and comparing the maximum modification time of all files to that of the object currently in the cache. If configuration files have been modified, they are re-parsed and the stale configuration object is updated in the cache.

Additional work was also performed:
- Comments were updated and brought in line with current XDMoD standards
- An `iConfiguration` interface is now provided that all `Configuration` object code must implement. This specifies the public methods that are available.
- A bug was corrected where timestamp columns specifying `CURRENT_TIMESTAMP` as a default value would see that value quoted when it should not be.

## Motivation and Context

Performance improvements prior to GroupBy refactoring.

## Tests performed

Implemented new tests.

## Types of changes

## Checklist:


plessbd:
need to update documentation: https://github.com/ubccr/xdmod/blob/xdmod8.5/docs/software-requirements.md

smgallo:
Updated docs as per @plessbd

classes/Configuration/Configuration.php
```diff
@@ -481,66 +564,97 @@ public function initialize($force = false)
...
+            // Only process .json files
+
+            $pos = strrpos(strtolower($file), '.json');
+            if ( false === $pos || (strlen($file) - $pos) != 5 ) {
+                continue;
+            }
+
+            $fullpath = $this->localConfigDir . '/' . $file;
+            $this->localConfigFiles[] = $fullpath;
+            $fileStats = stat($fullpath);
```
jpwhite4: Should check for a `false` return code from `stat()` here and output some sort or useful error message.

smgallo:
@jpwhite4 requested changes made.

PR #954, opened 2019-06-10
Fixed detecting when an SSO User is being edited

ryanrath:
## Description
Due to a previous change, the initialization of `cached_user_type` was moved to
inside the `else` clause. Which makes it a little hard when it's being used in
the `if` clause. I've updated the code so that it correctly detects if an SSO

## Motivation and Context
It's nice when the UI behaves as we expect

## Tests performed
Manual Tests were performed, i.e. edited an SSO User and made sure that the User Type was displayed as expected. 

## Types of changes

## Checklist:


jpwhite4:
The  cached_user_type code still looks whiffy to me. Line 669 does a triple equals check on it but it is set to a String and compared to an Int constant. Also it appears to be not initialized correctly. in the case where the user is an SSO user

ryanrath:
@jpwhite4 All set with simplifying that `if/else` case. I also was able to figure out / fix that infinite save loop problem that you showed me a bit ago.

PR #955, opened 2019-06-11
add ids to cloud summary charts

plessbd:
When going to the summary page and getting charts with data, the cloud summary charts do not have dataseries id's
This doesn't seem to cause any noticeable errors, but does create php warnings.

This fixes that.


classes/OpenXdmod/Migration/Version812To850/ConfigFilesMigration.php
```diff
@@ -15,6 +16,25 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
...
+                            if(!isset($data['data_series']['data'][$dsKey]['id'])) {
+                                $cloudRolesFile['+roles']['+default']['+summary_charts'][$key]['data_series']['data'][$dsKey]['id'] = $dataId;
+                                $dataId += 0.00000000000001;
+                            }
+
+                        }
+                    }
+                }
+            }
+        }
```
jpwhite4: Where is the code that saves the new file?

plessbd: you and your actually saving...

templates/roles.d/cloud.json
```diff
@@ -49,6 +49,7 @@
                             },
                             "group_by": "resource",
                             "has_std_err": false,
+                            "id": 0.00000000000010,
```
jpwhite4: Why the choice of 1.0e-13? It would be more readable to use 1, 2, 3, ...

jpwhite4: Also you have managed to pick two id values that will not work with the code!
```javascript
var i = 0.00000000000010;
var j = 0.00000000000011;

var test = function (x, y) {

    if (Math.abs(x - y) < 1e-14) {
        console.log('same');
    } else {
        console.log('different');
    }
};

test(i, j);
```

classes/OpenXdmod/Migration/Version812To850/ConfigFilesMigration.php
```diff
@@ -15,6 +16,26 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
...
+        $cloudRolesFilePath = CONFIG_DIR . '/roles.d/cloud.json';
+        if (file_exists($cloudRolesFilePath)) {
+            $cloudRolesFile = Json::loadFile($cloudRolesFilePath);
+            if (isset($cloudRolesFile['+roles']['+default']['+summary_charts'])) {
+                foreach($cloudRolesFile['+roles']['+default']['+summary_charts'] as $key => $data) {
+                    if(isset($data['data_series']['data'])){
+                        $dataId = 1;
+                        foreach($data['data_series']['data'] as $dsKey => $dsData) {
+                            if(!isset($data['data_series']['data'][$dsKey]['id'])) {
+                                $cloudRolesFile['+roles']['+default']['+summary_charts'][$key]['data_series']['data'][$dsKey]['id'] = $dataId;
```
smgallo:
```suggestion
                                $cloudRolesFile['+roles']['+default']['+summary_charts'][$key]['data_series']['data'][$dsKey]['id'] = $dataId++;
```

classes/OpenXdmod/Migration/Version812To850/ConfigFilesMigration.php
```diff
@@ -15,6 +16,26 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
...
+                                $dataId++;
```
smgallo:
```suggestion

```

PR #961, opened 2019-06-12
Update the  "Per Job" statistics to support aggregate view mode.

jpwhite4:
## Description
Previously there were several statistics that were disabled from display in "Aggregate" mode and would only show data in "Timeseries" plot mode. They also had a problem where the sort order of the data in timeseries mode was incorrect when sorting by value.

This fixes the statistics so they produce the correct answer in aggregate mode and have the correct sort order in timeseries mode.


jpwhite4:
Note that the `usesTimePeriodforAggregate()` functionality can be removed once the statstics are updated.

plessbd:
notes from @smgallo on this

### usesTimePeriodTablesForAggregate()

Indicates if a statistic can use tables where data is rolled up by time period to perform the calculation for an aggregate chart. In these cases, if data is compared between the sum of all timeseries data points and the aggregate value for the same time period, the values are not equal. If an aggregate view is requested and any of these statistics is being displayed either the chart will be automatically switched to timeseries mode for a chart with a single metric or a message indicating that the statistic cannot be used in aggregate form will be displayed for a chart displaying multiple statistics.

Returns `false` in the following classes:
- Open XDMoD
    - classes/DataWarehouse/Query/Jobs/Statistics/AverageCPUHoursStatistic.php
    - classes/DataWarehouse/Query/Jobs/Statistics/AverageNodeHoursStatistic.php
    - classes/DataWarehouse/Query/Jobs/Statistics/AverageProcessorCountStatistic.php
    - classes/DataWarehouse/Query/Jobs/Statistics/NormalizedAverageProcessorCountStatistic.php
- XSEDE
    - classes/DataWarehouse/Query/Jobs/Statistics/AverageNUStatistic.php
    - classes/DataWarehouse/Query/Jobs/Statistics/AverageSUStatistic.php

PR #971, opened 2019-06-28
Adding New User Help Tour and functionality to reset if a user has seen a tour or not

eiffel777:
This PR adds the New User guided tour. When a user logs in a check is made to see if they have viewed the New User tour or not. If they have not, a message box is shown and asks them if they would like to see the New User tour. Along with the option to view or decline the tour a checkbox is also present in the message box asking if they would no longer like to see the 'Do you want to see the New User tour' message box when they log in. 

An item is placed in the Help Menu so that anyone can see the help tour at any time no matter if they have seen it before or not. 

In the User Management pane in the internal dashboard another option has also been placed in the Actions drop down so that the attribute storing if a person has seen the New User Tour can be reset to false. This value is set to true once a person gets to the last tip in the help tour.

## Motivation and Context
Help new users become more familiar with the user interface and functionality of XDMoD

## Tests performed
Manually tested in docker

## Types of changes

## Checklist:


html/controllers/user_admin/reset_user_tour_viewed.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
jpwhite4: I though that the controller interface code was deprecated and all new code should use the rest stack instead.

classes/Rest/Controllers/SummaryControllerProvider.php
```diff
@@ -191,4 +194,35 @@ public function resetLayout(Request $request, Application $app)
...
         ));
     }
+
+    /**
+     * Set value for if a user should view the help tour or not
+     */
+    public function setViewedUserTour(Request $request, Application $app)
+    {
+        $user = $this->authorize($request);
+        $content = json_decode($this->getStringParam($request, 'data', true), true);
```
jpwhite4: Need to properly sanitize use input before storing in the database. This code has a potential code injection vulnerability since the user could put arbitrary content in the 'data' object.

html/gui/js/modules/NoviceUser.js
```diff
@@ -118,6 +118,56 @@ Ext.extend(XDMoD.Module.Summary, XDMoD.PortalModule, {
...
+
+                    if (CCR.xdmod.publicUser !== true) {
+                        var userTour = self.createNewUserTour();
+                        var conn = new Ext.data.Connection();
+
+                        conn.request({
+                            url: XDMoD.REST.url + '/summary/viewedUserTour',
+                            params: {
+                                data: Ext.encode({
+                                    uid: CCR.xdmod.ui.mappedPID,
```
jpwhite4: What is the reason for putting the uid and (transient) token in the data object?

classes/Rest/Controllers/SummaryControllerProvider.php
jpwhite4: Why the complex data encoding? You can simply have a single parameter true / false that would meet the requirements for this end point

html/controllers/user_admin/reset_user_tour_viewed.php
```diff
@@ -0,0 +1,24 @@
...
+    \xd_response\presentError('invalid_id_specified');
+};
+
+$user = XDUser::getUserByID($_POST['uid']);
+if ($user === null) {
+    \xd_response\presentError("user_does_not_exist");
+}
+
+$storage = new \UserStorage($user, 'viewed_user_tour');
+$storage->upsert(0, ['uid' => $_POST['uid'], 'viewedTour' => $_POST['viewed_user_tour']]);
```
jpwhite4: Why store the uid in this data structure?

html/controllers/user_admin/reset_user_tour_viewed.php
ryanrath: ^ 100% yes please.

classes/Rest/Controllers/AdminControllerProvider.php
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        $storage = new \UserStorage($selected_user, 'viewed_user_tour');
+        $storage->upsert(0, ['viewedTour' => $viewedTour]);
+
+        return $app->json(
+            array(
+                'success' => true,
+                'total' => 1,
+                'message' => 'This user will be now be propmted to view the New User Tour the next time they visit XDMoD'
```
jpwhite4: s/propmed/prompted/

classes/Rest/Controllers/AdminControllerProvider.php
```diff
@@ -0,0 +1,53 @@
...
+     * @param Request $request
+     * @param Application $app
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     * @throws \Exception
+     */
+    public function resetUserTourViewed(Request $request, Application $app)
+    {
+        $this->authorize($request, array('mgr'));
+        $viewedTour = $this->getIntParam($request, 'viewedTour', true);
+        $selected_user = XDUser::getUserByID($this->getIntParam($request, 'uid', true));
```
jpwhite4: Should check if `$selected_user === null` and throw a badrequest if so.

PR #988, opened 2019-07-26
Add a view chart json button for developers

jpwhite4:
This adds a button to the chart context menu that pops up a window that displays the relevant json config for the chart. This is only available to user accounts with the developer flag.

The purpose of this feature is to make it easy for system admins to add new charts to the summary page. They simply create the chart in the metric explorer then click on show json to get the text that can be cut and pasted into the configuration file.


jtpalmer:
Please update `docs/howto-summary-charts.md` with the relevant instructions.

jtpalmer:
After discussion with @jpwhite4 it was determined that the documentation I mentioned does not need to be updated at this time.

PR #993, opened 2019-07-26
Set default chart configuration

jpwhite4:
Update the default setting for the novice user summary page to match the original design. Also add etlv2 variables for the person_id and person_name to allow this information to be used in chart filter configuration settings.

### DIFF
--- a/configuration/roles.d/summary.json
+++ b/configuration/roles.d/summary.json
@@ -1,104 +1,510 @@
 {
-  "+roles": {
-    "+cd": {
-      "summary_portlets": [
-        {
-          "name": "Chart Thumbnails",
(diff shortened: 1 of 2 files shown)


configuration/roles.d/summary.json
```diff
@@ -1,104 +1,510 @@
...
-        {
-          "name": "Chart Thumbnails",
-          "type": "ReportThumbnailsPortlet",
-          "region": "top"
+        "+cs": {
+            "summary_portlets": [
+                {
+                    "name": "Chart Thumbnails",
+                    "type": "ReportThumbnailsPortlet",
+                    "config": {
```
plessbd: in other files you are ordering keys alphabetically.  Why not here?

PR #1007, opened 2019-08-20
Increase size of system username column

plessbd:
Cloud usernames can be whole email addresses, up the system username to be able to support this


smgallo:
The userid columns for the raw openstack events are 64 characters. Do we want to increase those as well?

plessbd:
`user_id` is an actual GUID so we don't need to change that update `user_name`

smgallo:
Lets add a comment into the definition so it's clear while we are making changes to that file.

PR #1017, opened 2019-08-30
update bootstrap and gitignore to be more specific

plessbd:
I need to pay more attention...

This fixes a few issues introduced in previous test updates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Numerous always-ignore extensions
-#*.diff
+*.diff
 *.err
 *.orig
 *.log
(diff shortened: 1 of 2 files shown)


jtpalmer:
Do you know why tests were still passing while these paths were incorrect?

plessbd:
yes :)

plessbd:
These diffs only do things when NOT testing the jobs realm.  Since we are *currently* always testing the jobs realms these diffs do not get applied.

jtpalmer:
Where are the instructions to test individual realms?

plessbd:
I will work on some documentation updates when we start building this out more, but in the sort term you set an environment variable `XDMOD_REALMS` equal to a comma separated list of currently `jobs`, `storage`, and/or `cloud`

```bash
if [ -z $XDMOD_REALMS ]; then
    export XDMOD_REALMS=jobs,storage,cloud
fi
```

PR #1021, opened 2019-09-05
Update tour construction

plessbd:
Move the Tour to be part of `CCR.js` instead of the UserDashboard.

### DIFF
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -229,7 +229,7 @@ XDMoD.GlobalToolbar.Contact = function () {
 // -------------------------------------------------
 
 XDMoD.createTour = function () {
-    if (typeof XDMoD.tour !== 'undefined') {
+    if (XDMoD.tour) {
         return;
     }
     // TODO: have this pull each of the 'UI' modules for their descriptions
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -228,6 +228,133 @@ XDMoD.GlobalToolbar.Contact = function () {
 
 // -------------------------------------------------
 
+XDMoD.createTour = function () {
+    if (typeof XDMoD.tour !== 'undefined') {
+        return;
+    }
+    // TODO: have this pull each of the 'UI' modules for their descriptions
+    // instead of hard coding these.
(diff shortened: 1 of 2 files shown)


html/gui/js/modules/NoviceUser.js
```diff
@@ -135,13 +133,17 @@ Ext.extend(XDMoD.Module.Summary, XDMoD.PortalModule, {
                                     Ext.Msg.show({
                                         cls: 'new-user-tour-dialog-container',
                                         title: 'New User Tour',
```
jtpalmer:
```suggestion
                                         title: 'XDMoD Tour',
```

html/gui/js/modules/NoviceUser.js
```diff
@@ -132,7 +132,7 @@ Ext.extend(XDMoD.Module.Summary, XDMoD.PortalModule, {
                                 if (serverResp.data.length === 0 || !serverResp.data[0].viewedTour) {
                                     Ext.Msg.show({
                                         cls: 'new-user-tour-dialog-container',
-                                        title: 'New User Tour',
+                                         title: 'XDMoD Tour',
```
jtpalmer: Now there's an extra space.

jtpalmer:
```suggestion
                                        title: 'XDMoD Tour',
```

html/gui/js/CCR.js
```diff
@@ -228,6 +228,133 @@ XDMoD.GlobalToolbar.Contact = function () {
 
 // -------------------------------------------------
 
+XDMoD.createTour = function () {
+    if (typeof XDMoD.tour !== 'undefined') {
```
jtpalmer: Why not
```js
if (XDMoD.tour !== undefined) {
```
or even
```js
if (XDMoD.tour) {
```

jtpalmer: Since the previous code (https://github.com/ubccr/xdmod/pull/1021/files#diff-fdf8fdd7e1b5ed7fe7177759254f4351L213) was using this logic I don't see a reason to change it.

```suggestion
    if (XDMoD.tour) {
```

plessbd: im an old programmer, no reason other than that...

PR #1035, opened 2019-09-11
Fix chart time display when the system timezone has a positive offset from UTC

jpwhite4:
Fix issue with the chart tooltip displaying the wrong date when
the system timezone is a negative offset from UTC. That is countries
located east of the prime meridian.

See https://help.xdmod.org/a/tickets/14165


plessbd:
timezones to the east of UTC have Positive offsets https://en.wikipedia.org/wiki/List_of_UTC_time_offsets

PR #1036, opened 2019-09-11
Add help window for Dashboard Components

jpwhite4:
This change also updates all components to inherit from the appropriate base
class.

A dashboard component author should add a 'help' property that contains
the definition of the help window title and contents.


jpwhite4:
The actual help content for the components will be in a future pull request

PR #1047, opened 2019-09-17
Add ability to hide realm from the metric catalog

jpwhite4:
This change prevents the information for a realm being shown in the Usage and Metric Explorer tabs. The reason for this change is that we do not want the JobEfficiency realm to be shown in the Usage and Metric Explorer tabs since it only exists for the purpose of the Job Efficiency dashboard.

There will be a pull request in the xdmod-supremm repo that sets the 'show_in_catalog' flag to false and a corresponding test that confirms that this works.

### DIFF
--- a/classes/DataWarehouse.php
+++ b/classes/DataWarehouse.php
@@ -357,7 +357,8 @@ class DataWarehouse
      * Get a mapping of categories to realms.
      *
      * If a realm does not explicitly declare a category, its name is used
-     * as the category.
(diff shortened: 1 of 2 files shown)


plessbd:
is this meant to go into 8.5?

smgallo:
I will add a configuration option `show_in_catalog` and an associated `showInMetricCatalog()` method for both Realms and Statistics in the new GroupBy code.

PR #1087, opened 2019-10-02
Remove unnecessary test artifacts

jpwhite4:


plessbd:
you should probably redo this off of 8.6. Other than that, it looks good, compared the test numbers at least and it looks like you didnt miss any of them.

Part of me wonders if you could even resuse the arrays in the jobs regression tests so we dont have to duplicate this in two places...

PR #1088, opened 2019-10-02
Make the Shredder ignore empty lines

kcgthb:
When the job accounting file contains empty lines, the shredder fails.

`xdmod-slurm-helper` fails when a job name contains a carriage-return character. The generated `sacct` output contains a blank line that makes `xdmod-slurm-helper` fail.

Here's an example:
```
$ sacct -j 51180322 -p -o jobid,jobname%32
JobID|JobName|
51180322|f20__1__e3__d55__t6__p1__rs0.5__rr165__c0.25
|
51180322.batch|batch|
51180322.extern|extern|
```
Note how the job name sends the last pipe delimiter to the next line. Because of this, the temporary file that `sacct` output is sent to contains a blank line. And `xdmod-slurm-helper` fails with the following error:
```
2019-10-02 11:58:14 [debug] Shredding line ''
2019-10-02 11:58:14 [info] Rolling back database transaction
2019-10-02 11:58:18 [critical] Failed to shred line 165993 of file
/tmp/sacct-log-sherlock-8syyVH "": Malformed Slurm sacct line: ''
(stacktrace: #0 /usr/bin/xdmod-slurm-helper(177):
OpenXdmod\Shredder->shredFile('/tmp/sacct-log-...')
#1 /usr/bin/xdmod-slurm-helper(21): main()
#2 {main})
2019-10-02 11:58:18 [critical] Malformed Slurm sacct line: ''
(stacktrace: #0 /usr/share/xdmod/classes/OpenXdmod/Shredder.php(405):
OpenXdmod\Shredder\Slurm->shredLine('')
#1 /usr/bin/xdmod-slurm-helper(177):
OpenXdmod\Shredder->shredFile('/tmp/sacct-log-...')
#2 /usr/bin/xdmod-slurm-helper(21): main()
#3 {main})
```

## Description
Check that the line is not empty before processing it. Skip empty lines

## Motivation and Context
https://help.xdmod.org/support/tickets/22127

## Tests performed
Run `xdmod-slurm-helper` on job records with job names containing carriage returns. The job records are correctly processed.

## Types of changes

## Checklist:


bin/xdmod-update-resource-specs
```diff
@@ -161,7 +161,7 @@ function main()
         DIRECTORY_SEPARATOR,
         array(
             CONFIG_DIR,
-            'resource-specs.json'
+            'resource_specs.json'
```
plessbd: this file is unrelated to this change.  please remove it from this PR

kcgthb: Oops, sorry about that. Fixed.

plessbd:
In the mean time you should be able to skip using `xdmod-slurm-helper` and use the documentation located at https://open.xdmod.org/8.1/resource-manager-slurm.html to export to a file and use `xdmod-shredder`.

plessbd:
I said might ;)  I am not the main developer of the shredders, but I thought we had some code in there that would at least catch this, but I could be wrong...

classes/OpenXdmod/Shredder.php
```diff
@@ -401,23 +401,27 @@ public function shredFile($file)
...
-                            'file'        => $file,
-                            'line_number' => $lineNumber,
-                            'line'        => $line,
-                        ));
-                        $duplicateCount++;
-                        continue;
-                    } else {
-                        throw $e;
+                // Skip empty lines
+                if ($line !== '') {
```
jpwhite4: if the line is skipped then the record count should not be updated.

classes/OpenXdmod/Shredder.php
```diff
@@ -401,23 +401,27 @@ public function shredFile($file)
...
+                                'message'     => $msg,
+                                'file'        => $file,
+                                'line_number' => $lineNumber,
+                                'line'        => $line,
+                            ));
+                            $duplicateCount++;
+                            continue;
+                        } else {
+                            throw $e;
+                        }
```
jtpalmer: I'd also like to see some debug logging in this change.  You can also simplify the change and address @jpwhite4 's comment with something like this:

```diff
- $recordCount++;

+ if ($line === '') {
+     $this->logger->debug([
+         'message'     => 'Skipping blank line',
+         'file'        => $file,
+         'line_number' => $lineNumber
+     ));
+     continue;
+ }
+ 
+ $recordCount++;
```

kcgthb: Ah yes, good suggestions! Let me try this.

kcgthb:
Implemented @jtpalmer's and @jpwhite4's suggestions.

plessbd:
switched back to the 8.5 branch so tests can hopefully pass since we dont have the new branch completely upgraded

kcgthb:
Well, thinking about it a little more, this will only solve the case where the job name contains `\n` but nothing after.
If a user submits a job like this:
```
$ srun -J "$(echo test$'\n'break)" sleep 1
```
the job name will be split over 2 lines in the `sacct` output, and the shredder will try to parse a line that contains just `break` and fail.

So I guess special characters should be escaped earlier, directly at the `sacct` level.

PR #1091, opened 2019-10-07
WIP: Fix Chart Settings

chakrabortyr:
WIP: Don't review yet, just want Shippable to run my tests.

## Description
Removes spurious JSON.encode/decode calls for chart settings in the Usage/Metric Explorer Tabs

## Motivation and Context
Already perfectly valid JSON, encoding/decoding is redundant. 
 
## Types of changes

## Checklist:


classes/DataWarehouse/Access/Usage.php
```diff
@@ -93,18 +93,17 @@ private function getSummaryCharts(XDUser $user) {
                 if (!$statsClass->isVisible()) {
                     continue;
                 }
-                $statUsageChartSettings = $usageChartSettings;
```
plessbd: why did this get removed?  shouldnt this change just be removing the json encode on line 117

chakrabortyr: looks like that variable was only ever assigned the value of $usageChartSettings and given a few additional variables. So it seemed redundant, ostensibly at least.

tests/artifacts/xdmod/schema/get-menus.spec.json
```diff
@@ -25,10 +25,10 @@
           "type":"string"
         },
         "defaultChartSettings": {
-          "type":"string"
```
plessbd: i would like to spec to be updated to have the object be validated for its expected contents as well

classes/DataWarehouse/Access/Usage.php
plessbd: i want to say it had to do with the different settings for the stat vs the groupby but i will need to talk with @ryanrath about why we added that one to be sure.

chakrabortyr:
This is now in #1134

PR #1006, opened 2019-08-15
Updated to add resource_types and resource_type_realms

ryanrath:
## Description
Added Resource Types -> Supported Realms relations that we then use to filter what data `acl-config` processes.

## Motivation and Context
We wanted to be able to only process, in `acl-config` the realms that are currently enabled. To accomplish this we associated resource_types w/ the appropriate realms. Then we select all currently installed resources and their types and retrieve a distinct list of the realms for those types. 

There was also a few changes made to support Trevor's realm separation work ( Making it so that XDMoD doesn't need to have the Jobs realm installed to function properly. ). 

## Tests performed
- All automated tests for: Fresh Install & Upgrade
- Manually tested on xdmod-dev
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project as found in the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #1103, opened 2019-10-10
Ensure resourcetypes table is populated correctly.

jpwhite4:
- Revert the change to the primary key value for the 'UNK' resource
  type. This is now the same value (0) as previous s/w versions.
- Ensure the 'UNK' resource is added to the various resource type tables
  in the jobs_staging pipeline.
- Revert the addition of the ingest-resource-types pipeline when
  shredding jobs. This was added in #1006 possibly as a work around for
  the fact that the staging pipeline had not been modfied appropriately.
- Add a test to make sure this does not happen again.

### DIFF
--- a/configuration/etl/etl.d/staging.json
+++ b/configuration/etl/etl.d/staging.json
@@ -99,7 +99,6 @@
             "name": "ResourceTypesStagingUnknown",
             "description": "Populates the resource types table w/ the Unknown resource",
             "namespace": "ETL\\Maintenance",
-            "truncate_destination": false,
             "class": "ExecuteSql",
             "options_class": "MaintenanceOptions",
             "sql_file_list": [
--- a/configuration/etl/etl.d/staging.json
+++ b/configuration/etl/etl.d/staging.json
@@ -96,6 +96,17 @@
             }
         },
         {
+            "name": "ResourceTypesStagingUnknown",
+            "description": "Populates the resource types table w/ the Unknown resource",
+            "namespace": "ETL\\Maintenance",
+            "truncate_destination": false,
+            "class": "ExecuteSql",
+            "options_class": "MaintenanceOptions",
+            "sql_file_list": [
+                "cloud_openstack/unknown_resource_type.sql"
+            ]
+        },
+        {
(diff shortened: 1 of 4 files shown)


configuration/etl/etl.d/staging.json
```diff
@@ -95,6 +95,17 @@
                 }
             }
         },
+        {
+            "name": "ResourceTypesStagingUnknown",
+            "description": "Populates the resource types table w/ the Unknown resource",
+            "namespace": "ETL\\Maintenance",
+            "truncate_destination": false,
```
smgallo: `truncate_destination` doesn't have any effect here because there is no table that is being operated on.

jpwhite4:
In XDMoD 7.5 and earlier the resourcetype table contained the following:
```
MariaDB [modw]> select * from resourcetype;
+----+-----------------------+---------+
| id | description           | abbrev  |
+----+-----------------------+---------+
|  0 | Unknown               | UNK     |
|  1 | Default Resource Type | Default |
+----+-----------------------+---------+
```

in XDMoD 8.0 and 8.1 it contained:
```
MariaDB [modw]> select * from resourcetype;
+----+----------------------------+--------+
| id | description                | abbrev |
+----+----------------------------+--------+
|  1 | High-performance computing | HPC    |
|  2 | High-throughput computing  | HTC    |
|  3 | Data-intensive computing   | DIC    |
|  4 | Grid of resources          | Grid   |
|  5 | Cloud resource             | Cloud  |
|  6 | Visualization system       | Vis    |
|  7 | Virtual Machine system     | VM     |
|  8 | Tape storage resource      | Tape   |
|  9 | Disk storage resource      | Disk   |
| 10 | Storage grid               | StGrid |
| 11 | User Support               | US     |
+----+----------------------------+--------+
```

so the question is whether we need an UNK type in the table? @ryanrath or @plessbd were there specific changes made that required the UNK type to be present?

plessbd:
There were not.  It was added because it was in the file.  And looking
through time it looks like unknown was added by me as part of the migration
to etlv2 probably as a copy paste error or part of testing

PR #1124, opened 2019-10-22
Revert changes to roadmap in #1031 now that trello fixed their…

plessbd:
puts the roadmap back to the original version since it now works again.

### DIFF
--- a/classes/OpenXdmod/Migration/Version850To900/ConfigFilesMigration.php
+++ b/classes/OpenXdmod/Migration/Version850To900/ConfigFilesMigration.php
@@ -14,6 +14,15 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
     public function execute()
     {
         $this->assertPortalSettingsIsWritable();
-        $this->writePortalSettingsFile();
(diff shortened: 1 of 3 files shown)


plessbd:
Note that is will fail until #1123 is merged.

classes/OpenXdmod/Migration/Version850To860/ConfigFilesMigration.php
```diff
@@ -14,6 +14,15 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
...
-        $this->writePortalSettingsFile();
+        $data = parse_ini_file($this->portalSettingsPath, true);
+        $roadmapUrl = 'https://trello.com/b/mdFESh6j.html';
+        if('https://trello.com/embed/board?id=mdFESh6j' !== $data['roadmap']['url']){
+            $roadmapUrl = $data['roadmap']['url'];
+        }
+        $this->writePortalSettingsFile(
+            array_merge(
+                ['roadmap_url' => $roadmapUrl]
+            )
```
jtpalmer: You don't need to merge a single array.

PR #1131, opened 2019-10-25
Proposed Fix for the underlying json config parsing problems

jpwhite4:
## Description
The `XDMoDConfiguration` class has problems parsing some of the xdmod configuration files. There have been a couple of attempts to workaround the problem notably f989888b0 and 7c497ac4c
but both of these did not fix the underlying design problem. 

The underlying problem is that the `JsonFile.php` parser is designed to be used
to parse structured record-based json data. It includes support for parsing hybid json-like files
that contain multiple json blobs and it has code that validates the files based on a very
limited set of allowable content structure. The XDMoD configuration files are not record-based
json, they do not contain multiple json blobs and they do not conform to the content structure
that the JsonFile class is designed to handle.

My proposed solution is to stop using the record-based parser for the config data.

This change also backs out the now unneeded workarounds and updates the test artifacts appropriately.

### DIFF
--- a/classes/Configuration/Configuration.php
+++ b/classes/Configuration/Configuration.php
@@ -170,14 +170,6 @@ class Configuration extends Loggable implements iConfiguration
     protected $initialized = false;
 
     /**
-     * Will force the results of `parse` to be an array.
-     *
-     * @var boolean
-     */
-
-    protected $forceArrayReturn = false;
-
-    /**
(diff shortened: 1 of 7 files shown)


ryanrath:
@jpwhite4 quick question before approving this one. Was this the PR where you were talking about redoing the inheritance of `JsonConfigFile` class? Or was that another one?

PR #1134, opened 2019-10-28
Fix Chart Settings

chakrabortyr:
## Description
Removes spurious JSON.encode/decode calls for chart settings in the Usage/Metric Explorer Tabs

## Motivation and Context
Already perfectly valid JSON, encoding/decoding is redundant. 
 
## Tests performed

Ran the tests. Also spun up a new Docker and verified that I could still view/modify a saved chart before and after this change.

## Types of changes

## Checklist:


html/gui/js/modules/Usage.js
```diff
@@ -1454,7 +1454,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                             chartContainer.setHeight(CCR.xdmod.ui.thumbHeight * chartThumbScale);
                             chartContainer.mask('Loading...');
 
-                            var chart_params = Ext.apply({}, Ext.util.JSON.decode(r.get('chart_settings')), menuParams);
+                            var chart_params = Ext.apply({}, r.get('chart_settings'), menuParams);
```
jpwhite4: What is the impact of this change on any saved charts in the report generator?

chakrabortyr: The chart settings appear to be saved in moddb as RESTful query parameters, which are then encoded as JSON and sent to the frontend. So we should be okay.

jpwhite4: Also need to check the app kernels module is not impacted by this change.

jpwhite4:
Don't forget that you have to update all the places in the frontend code that processes the chart setting data. This includes but is not limited to:
- The `reloadChartFunc` in Usage.js
- The `fromJSON` function in ChartToolbar.js
- The `getChartParameters` function in Usage.js
- The chart store load callback in Usage.js

html/gui/js/modules/Usage.js
```diff
@@ -1925,11 +1925,11 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                         if (n.attributes.supportsAggregate === false) {
                             // do not overide chart settings for nodes that correspond to statistics that must be
                             // shown in timeseries mode (note for future: there should be no statistics like this in XDMoD).
-                            var decoded = JSON.parse(n.attributes.chartSettings);
+                            var decoded = n.attributes.chartSettings;
```
jpwhite4: Hmm. Javascript objects are assigned by reference whereas the original code assigned by value. In the original code the value of n.attributes.chartSettings was not modified. However it is now bein modified in this code. What is the impact of this change?

html/gui/js/modules/Usage.js
```diff
@@ -2659,12 +2659,14 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                         'Chart: ' + chartStore.getAt(0).get('title') + ', Params: ' + chartStore.getAt(0).get('params_title')
                     );
 
-                    var serverChartSettings = chartStore.getAt(0).get('chart_settings').replace(/`/g, '"');
-                    if (serverChartSettings === '') {
-                        serverChartSettings = n.attributes.defaultChartSettings;
+                    var serverChartSettings = Ext.apply({}, chartStore.getAt(0).get('chart_settings'));
+
+                    if (!serverChartSettings) {
+                        serverChartSettings = Ext.apply({}, n.attributes.defaultChartSettings);
```
plessbd: why did this change from pass by reference to pass by value?

.gitignore
```diff
@@ -4,6 +4,7 @@
 *.orig
 *.log
 *.rej
+*.sarif
```
plessbd: this is not related to this change please remove it.

html/gui/js/modules/Usage.js
```diff
@@ -1083,7 +1083,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
             var newStartDate = self.getDurationSelector().getStartDate().format('Y-m-d');
             var newEndDate = self.getDurationSelector().getEndDate().format('Y-m-d');
             var aggregationUnit = self.getDurationSelector().getAggregationUnit();
-            var chartSettings = JSON.decode ? JSON.decode(n.attributes.chartSettings) : JSON.parse ? JSON.parse(n.attributes.chartSettings): {};
+            var chartSettings = Ext.apply({}, n.attributes.chartSettings);
```
jpwhite4: I thought that Ext.apply was a shallow copy. Is this safe to do here?

html/gui/js/modules/Usage.js
```diff
@@ -1925,11 +1925,11 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                         if (n.attributes.supportsAggregate === false) {
                             // do not overide chart settings for nodes that correspond to statistics that must be
                             // shown in timeseries mode (note for future: there should be no statistics like this in XDMoD).
-                            var decoded = JSON.parse(n.attributes.chartSettings);
+                            var decoded = Ext.apply({}, n.attributes.chartSettings);
```
jpwhite4: There is no need to create an object copy here since this code is supposed to modify the node itself.
```js
if (n.attributes.supportsAggregate === false) {
    n.attributes.dataset_type = 'timeseries';
    n.attributes.display_type = 'line';
    n.attributes.swap_xy = false;
}
```

html/gui/js/modules/Usage.js
```diff
@@ -2659,12 +2659,14 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                         'Chart: ' + chartStore.getAt(0).get('title') + ', Params: ' + chartStore.getAt(0).get('params_title')
                     );
 
-                    var serverChartSettings = chartStore.getAt(0).get('chart_settings').replace(/`/g, '"');
-                    if (serverChartSettings === '') {
-                        serverChartSettings = n.attributes.defaultChartSettings;
+                    var serverChartSettings = Ext.apply({}, chartStore.getAt(0).get('chart_settings'));
+
+                    if (!serverChartSettings) {
```
jpwhite4: This won't work

html/gui/js/modules/Usage.js
chakrabortyr: Yes, we don't have nested objects in the chart settings.

jpwhite4:
The controllers are planned to go away in 9.5 so there is no point in changing them now.

chakrabortyr:
How anticlimactic

PR #1141, opened 2019-10-31
Do not json_encode included file

plessbd:
this is a fix that was added to `feature/groupby-refactor` as it was discovered during it.
However, this is a fix unrelated to that work.


jtpalmer:
See #1049

PR #1142, opened 2019-10-31
Allow JSON references to overwrite keys

plessbd:
added to `feature/groupby-refactor` but not exclusively related
see #1078 for more details


jtpalmer:
See #1078

PR #1145, opened 2019-11-01
Make shippable failure logging better

plessbd:
on failures the logs sometimes get truncated this prioritizes the logs that get output so that we can hopefully get better logging.

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -31,7 +31,7 @@ build:
         - ./tests/ui/runtests.sh --headless --log-junit `pwd`/shippable/testresults --sso
         - ./vendor/phpunit/phpunit/phpunit -c ./tests/integration/phpunit.xml.dist --testsuite sso --log-junit `pwd`/shippable/testresults/xdmod-sso-integration.xml
         - mv ./configuration/portal_settings.ini.old ./configuration/portal_settings.ini
-        - if [ -s /var/log/xdmod/apache-error.log ]; then cat /var/log/xdmod/apache-error.log; false; fi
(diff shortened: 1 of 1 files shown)


shippable.yml
```diff
@@ -33,7 +33,10 @@ build:
         - mv ./configuration/portal_settings.ini.old ./configuration/portal_settings.ini
         - if [ -s /var/log/xdmod/apache-error.log ]; then cat /var/log/xdmod/apache-error.log; false; fi
```
jtpalmer:
```suggestion
        - [ ! -s /var/log/xdmod/apache-error.log ]
```

shippable.yml
```diff
@@ -31,9 +31,12 @@ build:
         - ./tests/ui/runtests.sh --headless --log-junit `pwd`/shippable/testresults --sso
         - ./vendor/phpunit/phpunit/phpunit -c ./tests/integration/phpunit.xml.dist --testsuite sso --log-junit `pwd`/shippable/testresults/xdmod-sso-integration.xml
         - mv ./configuration/portal_settings.ini.old ./configuration/portal_settings.ini
-        - if [ -s /var/log/xdmod/apache-error.log ]; then cat /var/log/xdmod/apache-error.log; false; fi
+        - [ ! -s /var/log/xdmod/apache-error.log ]
```
jtpalmer: Shippable does not like my previous suggestion, but this should work:
```suggestion
        - test ! -s /var/log/xdmod/apache-error.log
```

PR #1151, opened 2019-11-05
Updates to Cloud\GroupByDomain and test artifacts

ryanrath:
## Description
- Updated the `GroupByDomain` class so that it correctly joins up to other tables
by both domain and resource, this was causing incorrect data to be shown.

- Updated the existing cloud data being ingested to include a `domain` property.
This now ensures that when viewing Cloud data by domain this information is
actually returned.

- With `GroupByDomain` being updated we needed to regenerate all of the test
artifacts for it.

## Motivation and Context
When it says Group By Domain, it should group by domain. 

## Tests performed
all automated tests.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Cloud/GroupBys/GroupByDomain.php
+++ b/classes/DataWarehouse/Query/Cloud/GroupBys/GroupByDomain.php
@@ -103,15 +103,7 @@ class GroupByDomain extends GroupBy
         $query->addGroup($domainIdField);
 
         $fkField = new TableField($dataTable, 'domain_id');
-
-        // the cloud_fact_by_* tables are *special* so they went with `host_resource_id` instead of `resource_id` for
-        // their `resource_id` fk field. We need to account for that here.
-        $resourceFkFieldName = 'resource_id';
-        if (strpos($dataTable->getName(), 'cloudfact_by') !== false) {
-            $resourceFkFieldName = 'host_resource_id';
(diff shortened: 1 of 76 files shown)


classes/DataWarehouse/Query/Cloud/GroupBys/GroupByDomain.php
```diff
@@ -85,27 +85,37 @@ public function __construct()
...
 
         $query->addGroup($domainIdField);
 
         $fkField = new TableField($dataTable, 'domain_id');
-        $resourceFkField = new TableField($this->table, 'resource_id');
+
+        // the cloud_fact_by_* tables are *special* so they went with `host_resource_id` instead of `resource_id` for
+        // their `resource_id` fk field. We need to account for that here.
+        $resourceFkFieldName = 'resource_id';
+        if (strpos($dataTable->getName(), 'cloudfact_by') !== false) {
```
jpwhite4: Is this if statement necessary? This code only runs in the cloud realm so will always use the cloudfact tables?

ryanrath: Yeah, I had added this while debugging an issue where the FilterListBuilder was bombing due to it's fk field being `resource_id` not `host_resource_id`. But I can't reproduce that error case with the code as it is now.

PR #1153, opened 2019-11-06
WIP: Fix edge case when Realm gets removed

plessbd:
There is an edge case that prevents the data warehouse export from working when a realm has been removed.

this resolves that only by putting that the realm is invalid as the name.

the ability to resubmit the request will fail as the realm is invalid, deleting the request will work as will downloading the file if it is currently available.

### DIFF
--- a/html/gui/js/modules/DataExport.js
+++ b/html/gui/js/modules/DataExport.js
@@ -358,7 +358,11 @@ XDMoD.Module.DataExport.RequestsGrid = Ext.extend(Ext.grid.GridPanel, {
                     dataIndex: 'realm_id',
                     scope: this,
                     renderer: function (value) {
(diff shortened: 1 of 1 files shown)


jtpalmer:
I don't think it's a good idea to display the realm as "Unsupported Realm".  It would be preferable to display the actual realm name or not display the row at all.

plessbd:
Yeah, i wasnt 100% sure how we wanted to handle it.
And really this is how I am making it seem less broken in a specific use case at the moment and i didnt want to lose this.

jpwhite4:
I don't think XDMoD supports removing realms. How does one go about removing a realm?

plessbd:
this specifically happened because I upgraded xdmod and removed supremm

This was completely an edge case of my own design.

PR #1158, opened 2019-11-07
Improvements validation of untrusted input.

jpwhite4:
This pull request has several small updates to improve handling of untrusted user supplied data. Since these are all fixes for the same class of problem I have put them into the same pull request but kept the individual changes as separate commits.

1. Improve input validation for get_rawdata endpoint.

   The get_rawdata metric explorer endpoint had some rather  obfuscuted code that checked for valid   data in the start and    limit variables. This code was difficult to understand and    it also triggered false positives in the code analysis software.
    There was also a minor inconsistency if the start parameter    was missing from the request. The row data were returned    as if the start parameter was 0 but the totalAvailable metric    was not returned.
    This update changes the code to use the standard php filter    API to ensure sanitized input and now has consistent behavior    for the totalAvailable metric.
    
    Also added tests.

1. Suppress output of known-invalid untrusted user data in pseudo login. This code appears to be here for debug purposes.

1. Update the password reset page to validate the reset identifier
    
    Ensure the validated reset identifier is passed back to the browser
    rather than the unmodified user-supplied version.

1. Ensure user id is correctly validated.
    
    Update the XDUser class to use the known valid id from the database
    rather than a user-controlled value.

### DIFF
--- a/tests/integration/lib/Controllers/MetricExplorerTest.php
+++ b/tests/integration/lib/Controllers/MetricExplorerTest.php
@@ -112,7 +112,7 @@ class MetricExplorerTest extends BaseTest
         $this->assertEquals(401, $response[1]['http_code']);
     }
 
-    function rawDataProvider()
+    public function rawDataProvider()
     {
         $params = array (
             'show_title' => 'y',
--- a/tests/integration/lib/Controllers/MetricExplorerTest.php
+++ b/tests/integration/lib/Controllers/MetricExplorerTest.php
@@ -160,7 +160,7 @@ class MetricExplorerTest extends BaseTest
                     'total' => 0,
                 ),
                 'z_index' => 0,
-                'visibility' => NULL,
+                'visibility' => null,
                 'enabled' => true,
                 'metric' => 'job_count',
                 'realm' => 'Jobs',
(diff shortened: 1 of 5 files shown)


tests/integration/lib/Controllers/MetricExplorerTest.php
```diff
@@ -112,6 +112,133 @@ public function testGetDimensionFilters()
...
+                'trend_line' => false,
+                'line_type' => 'Solid',
+                'line_width' => 2,
+                'shadow' => false,
+                'filters' => array (
+                    'data' => array (),
+                    'total' => 0,
+                ),
+                'z_index' => 0,
+                'visibility' => NULL,
```
jtpalmer:
```suggestion
                'visibility' => null,
```

tests/integration/lib/Controllers/MetricExplorerTest.php
```diff
@@ -112,6 +112,133 @@ public function testGetDimensionFilters()
         $this->assertEquals(401, $response[1]['http_code']);
     }
 
+    function rawDataProvider()
```
jtpalmer:
```suggestion
    public function rawDataProvider()
```

PR #1164, opened 2019-11-13
Change ACL error message

jtpalmer:
## Description

Change `acl-config` error message to add additional details.

## Motivation and Context

Specifies the command to run so you don't need to look it up and provides a URL so you don't need to search for it.

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -427,7 +427,8 @@ SQL;
         $log->err('Backup tables detected! ');
         $log->err('Unable to continue!');
         $log->err('This may indicate a failed previous run.');
-        $log->err('Please see the XDMoD Troubleshooting documentation to resolve.');
+        $log->err('Run "acl-config --recover" to recover using backup tables.');
+        $log->err('See https://open.xdmod.org/9.0/faq.html#why-do-i-see-backup-tables-detected-when-running-the-acl-config-script for more information.');
         exit(1);
     } elseif (!$backupTablesExist) {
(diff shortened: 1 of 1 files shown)


bin/acl-config
```diff
@@ -427,7 +427,8 @@ SQL;
         $log->err('Backup tables detected! ');
         $log->err('Unable to continue!');
         $log->err('This may indicate a failed previous run.');
-        $log->err('Please see the XDMoD Troubleshooting documentation to resolve.');
+        $log->err('Run "acl-config --recover" to recover using backup tables.');
+        $log->err('See https://open.xdmod.org/9.0/faq.html#why-do-i-see-backup-tables-detected-when-running-the-acl-config-script for more information.');
```
plessbd: is there any way to have this link without the version number in it?  from my very basic testing i couldnt get it to redirect with the anchor but didnt know if you looked into it

I dont like the idea of having to update this everytime. but i also dont like my suggestion very much either...
```suggestion
        $log->err('See the XDMoD FAQ (https://open.xdmod.org/faq.html)'); 
        $log->err('section: Why do I see “Backup tables detected!” when running the acl-config script?');
        $log->err('for more information.');
```

jtpalmer: I couldn't find a URL that would redirect properly either.  I'd prefer to not include the version number, but unless we can get the redirect working this is good enough for me.

PR #1167, opened 2019-11-21
Remove hardcoded database ids.

jpwhite4:
The rest endpoint that generated the dashboard report used a hardcoded
template identifier for each known role. This severly limits the
ease of adding templates for new roles and also is a problem with the
XSEDE instance where the template ids do not match the open source one
since there are more templates in the database.

This changes the code to lookup the template based on the template name
and uses the one appropriate to the user's role. Note that we already
prescribed the name for the dashboard templates to be "Dashboard Tab
Report".

This new code deliberately omits the fallthrough case where an unknown
role would get report template 3. Instead it returns a not-found error.
Note that this scenario cannot happen with the default configuration  of
the open source software. If a site chooses to modify to configuration
to allow a role to be able to see the report then they must also specify
the report to be viewed.

The enumerateTemplates code has an order by clause added to it. This
will ensure that the results from the query are returned in a
deterministic way. The code always uses the first template in the list
if multiple match the query.

**This needs to go into 8.6 to allow us to create reports for the various 
XSEDE-specific roles.**


classes/Rest/Controllers/DashboardControllerProvider.php
```diff
@@ -11,6 +11,7 @@
 
 use Models\Services\Acls;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use DataWarehouse\Query\Exceptions\NotFoundException;
```
jtpalmer: I think it would be more appropriate to use `Symfony\Component\HttpKernel\Exception\NotFoundHttpException`.  The documentation for `DataWarehouse\Query\Exceptions\NotFoundException` refers to unknown group by, but that might just be out of date.  Using the `HttpKernel` exception makes it clear that it is intended to be handled by the REST framework.

classes/Rest/Controllers/DashboardControllerProvider.php
```diff
@@ -276,7 +270,12 @@ public function getRoleReport(Request $request, Application $app)
                 }
             }
             if (is_null($userReport)){
-                $template = $rm->retrieveReportTemplate($user, $template_id);
+                $availTemplates = $rm->enumerateReportTemplates(array($role), 'Dashboard Tab Report');
+                if (empty($availTemplates)) {
+                    throw new NotFoundException();
```
jtpalmer: I'm not sure if it would get used anywhere, but passing a message wouldn't hurt.

PR #1169, opened 2019-11-26
Add report template that supports variable substitution.

jpwhite4:
This report template supports variable substitution for any
users acl parameters. The parameter IDs and names are available.

This template will be used in the XSEDE 8.6 release, but belongs in the code since it does not contain any XSEDE-specific  stuff.

See the corresponding XSEDE pull request for example of how it is used.

### DIFF
--- a/classes/ReportTemplates/VariableReportTemplate.php
+++ b/classes/ReportTemplates/VariableReportTemplate.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace ReportTemplates;
+
+use Models\Services\Parameters;
+use DataWarehouse\Access\MetricExplorer;
+use XDReportManager;
+
+/**
+ * Generates a report that includes macro expansion of variables defined in the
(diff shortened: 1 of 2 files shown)


classes/ReportTemplates/VariableReportTemplate.php
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Generates a report that includes macro expansion of variables defined in the
+ * template. The substitution variables are generated from the users acls.
+ * The internal identifier for a dimension has an _id postfix and the human
+ * readable string has a _name postfix. For example, to get the service provider
+ * name for a center staff user you would include the string ${provider_name} in
+ * the template and ${provider_id} gives the numerical identifier for the provider.
+ */
+class VariableReportTemplate extends \ReportTemplates\aReportTemplate
```
plessbd: Something like `parameterized` might be a bit better than `Variable`

PR #1171, opened 2019-11-27
Fix foreign key constraint on moddb report_template_acls

jpwhite4:
Also fix the ETLv2 code to correctly handle modifications to foreign
keys.

### DIFF
--- a/classes/ETL/DbModel/ForeignKeyConstraint.php
+++ b/classes/ETL/DbModel/ForeignKeyConstraint.php
@@ -188,15 +188,15 @@ class ForeignKeyConstraint extends NamedEntity implements iEntity
         }
 
         if ($this->on_delete != $cmp->on_delete
-            && !in_array($this->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
-            && !in_array($cmp->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
+            && !(in_array($this->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
+                 && in_array($cmp->on_delete, array(null, 'RESTRICT', 'NO ACTION')))
         ) {
             return -2;
         }
 
         if ($this->on_update != $cmp->on_update
-            && !in_array($this->on_update, array(null, 'RESTRICT', 'NO ACTION'))
-            && !in_array($cmp->on_update, array(null, 'RESTRICT', 'NO ACTION'))
+            && !(in_array($this->on_update, array(null, 'RESTRICT', 'NO ACTION'))
+                 && in_array($cmp->on_update, array(null, 'RESTRICT', 'NO ACTION')))
         ) {
             return -3;
         }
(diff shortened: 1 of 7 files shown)


classes/ETL/DbModel/ForeignKeyConstraint.php
```diff
@@ -188,15 +188,15 @@ public function compare(iEntity $cmp)
         }
 
         if ($this->on_delete != $cmp->on_delete
-            && !in_array($this->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
-            && !in_array($cmp->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
+            && !(in_array($this->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
+                 && in_array($cmp->on_delete, array(null, 'RESTRICT', 'NO ACTION')))
```
jtpalmer: This is equivalent to the before the change.

```
p := in_array($this->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
q := in_array($cmp->on_delete, array(null, 'RESTRICT', 'NO ACTION'))
r := $this->on_delete != $cmp->on_delete
```


| p | q | r | !p | !q | !p && !q | p && q | !(p && q) | r && !p && !q | r && !(p && q) |
| - | - | - | -- | -- | -------- | ------ | --------- | ------------- | -------------- |
| T | T | X | F  | F  | F        | T      | F         | F             | F              |
| T | F | F | F  | T  | F        | F      | T         | F             | F              |
| F | T | F | T  | F  | F        | F      | T         | F             | F              |
| F | F | Y | T  | T  | T        | F      | T         | Y             | Y              |

(Pardon my abuse of logic notation.)

The only cases where the result is different are when `p != q`, but if one of `$this->on_delete` and `$cmp->on_delete` are in the set `array(null, 'RESTRICT', 'NO ACTION')` (and the other isn't) then they cannot be equal (which is already guaranteed since `$this->on_delete != $cmp->on_delete` must also be true).

The logic is convoluted, so if you think this is more readable then you can change it.

jpwhite4: The code is most definitely different. The bug was that CASCADE was not being correctly diffed against NO ACTION. See code below for test case:

```php
<?php
  
function orig_diff($left, $right)
{
    $return = false;
    if ($left != $right &&
        !in_array($left, array(null, 'RESTRICT', 'NO ACTION')) &&
        !in_array($right, array(null, 'RESTRICT', 'NO ACTION'))) {
        $return = true;
    }
    return $return;
}

function new_diff($left, $right)
{
    $return = false;
    if ($left != $right &&
        !(in_array($left, array(null, 'RESTRICT', 'NO ACTION')) &&
        in_array($right, array(null, 'RESTRICT', 'NO ACTION')))) {
        $return = true;
    }
    return $return;
}


var_dump(orig_diff('CASCADE', 'NO ACTION'));
var_dump(new_diff('CASCADE', 'NO ACTION'));
```

jtpalmer: Looks like I was reading it as `$this->on_delete == $cmp->on_delete`.  You should add a test since this case isn't covered.

jpwhite4: Yep I'll be adding some more unit tests to this pull request since the coverage is a bit sparse.

PR #1172, opened 2019-12-02
Adding transactions / lock files to acl-config

ryanrath:
## Description
- We recently ran into an issue with multiple instances of `acl-config` running
  at the same time. While `acl-config` is meant to be indempotent ( as long as
  the underlying configuration files are not changed ), this was not guaranteed
  when multiple instances of `acl-config` are run at the same time. To resolve
  this issue, `acl-config` now utilizes a lock file to ensure that one and only
  one `acl-config` instance is running at a time.
- This same issue brought to light the lack of transactions in `acl-config`.
  This PR wraps all of `acl-config`s db actions in a transaction. This makes
  `acl-config` more robust when something like `ctrl+c` is pressed while
  running. In this case, we capture that a signal has been sent and attempt to
  perform a rollback to preserve db state.

## Motivation and Context
Making `acl-config`s behavior more robust is a good thing

## Tests performed
- All manual tests performed
- Manual testing of `acl-config` while running, including killing it mid run and spinning up multiple copies to make sure that the lockfile handling works correctly. 

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -36,6 +36,21 @@ $logLevel = Log::ERR;
 $recover = false;
 
 try {
+    $lockFile = '/var/tmp/acl-config.lock';
+    $lockFileHandle = @fopen($lockFile, 'w');
+
+    if ($lockFileHandle === false) {
+        $lastError = error_get_last();
+        fwrite(STDERR, sprintf("Current User: %s\n", get_current_user()));
+        fwrite(STDERR, sprintf("Failed to open lock file \"%s\": %s", $lockFile, print_r($lastError, true)));
+        exit(1);
+    }
+
+    if (!@flock($lockFileHandle, LOCK_EX | LOCK_NB)) {
+        fwrite(STDERR, "acl-config not running due to another process holding the lock.\n");
+        exit(1);
+    }
+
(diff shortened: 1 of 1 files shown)


bin/acl-config
```diff
@@ -79,21 +94,55 @@ try {
...
+    }
+    $log->debug("*** Database Verification Passed!");
+
+    $db = DB::factory('database');
+
+    // Begin by making sure that we have an active transaction running. If something funky happens then we need to make
+    // sure we can roll back any changes that were made.
+    $db->beginTransaction();
+
+    $sigHandler = function ($sig) use (&$db) {
```
jpwhite4: There is no need for this code. The transaction does not need to be rolled back unless it has been committed.

jpwhite4:
This code is still calling DROP table. This command cannot be run in a transaction.

ryanrath:
Closing this PR per conversation w/ @jpwhite4. I'll be starting a new PR that aims to cleanup `acl-config` so that the only thing it does is deal with synchronizing the contents of the DB w/ the contents of the configuration files. I'll be merging in the lock file enhancement with that PR.

PR #1173, opened 2019-12-02
Fix race condition in test harness

jpwhite4:
The unit tests were not passing in my docker. The reason was because there is no guarantee that the process has exited by the time the `proc_get_status()` function is called. So the test would fail if the process was still in the running state when the status function was executed.

This change removes the `proc_get_status()` call and instead uses the return value from the proc_close which is guaranteed to give a yes/no answer.

### DIFF
--- a/classes/ETL/DataEndpoint/Filter/ExternalProcess.php
+++ b/classes/ETL/DataEndpoint/Filter/ExternalProcess.php
@@ -123,6 +123,7 @@ class ExternalProcess extends \php_user_filter
             // Close the process's input file so it can finish up
 
             @fclose($this->pipes[0]);
+            $this->pipes[0] = null;
 
             // Process all data left on the pipe from the process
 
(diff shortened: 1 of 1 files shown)


classes/ETL/DataEndpoint/Filter/ExternalProcess.php
```diff
@@ -220,30 +221,26 @@ public function onCreate()
...
-        // While the process is running the exit status is shown as -1
+        $exitStatus = proc_close($this->filterResource);
+        fclose($this->tmpResource);
 
-        if ( ! in_array($status['exitcode'], array(-1, 0)) ) {
-            $errorMessage = 'Error executing external filter process: ';
-            while ( ! feof($this->pipes[2]) ) {
-                $errorMessage .= fgets($this->pipes[2]);
-            }
+        if ($exitStatus !== 0) {
```
jtpalmer: I don't know how accurate this is, but the [proc_close documentation](https://www.php.net/proc_close) states that:

> Returns the termination status of the process that was run. In case of an error then -1 is returned.

I don't know if it's possible that a value other than 0 or -1 is returned, but [elsewhere](https://github.com/ubccr/xdmod/blob/xdmod9.0/bin/xdmod-slurm-helper#L375) I've written code that explicitly compares against -1.

classes/ETL/DataEndpoint/Filter/ExternalProcess.php
```diff
@@ -220,30 +221,26 @@ public function onCreate()
...
+        }
+        fclose($this->pipes[1]);
 
-        $status = proc_get_status($this->filterResource);
+        // Read stderr stream to EOF before closing.
+        $errorOutput = '';
+        while (($buffer = fgets($this->pipes[2])) !== false) {
+            $errorOutput .= $buffer;
+        }
+        fclose($this->pipes[2]);
```
jtpalmer: Is it possible that anything additional could be written to STDOUT and STDERR between here and the `proc_close` below?  I'm curious if this can be put after the `proc_close`, but I've never tried that.

classes/ETL/DataEndpoint/Filter/ExternalProcess.php
jpwhite4: exit status can be a range of values not just 0 or -1. The unit tests have a 127 and a 1 return values (this can be seen in the logs).

classes/ETL/DataEndpoint/Filter/ExternalProcess.php
jpwhite4: The docs state that you should close all open file handles before the `proc_close` to avoid deadlocks. Since STDERR is read until EOF (or error) then you shouldn't get any more data out of stderr. Any data sitting in the stdout buffer will get thrown away, but this should be handled by the code starting on line 121.

If you add a filter that writes an infinite amount of data to stderr after stdin is closed then the proc_close() call will never execute. Don't do this.

PR #1180, opened 2019-12-10
Update organization affiliation change email

jtpalmer:
## Description

Change the text of the organization affiliation change user notification email, the contact email address used in the message, and the reply-to address.

## Motivation and Context

I'm not sure what is the best contact email to use in this email, but it's definitely not the "sender" email, which in our case is a "do-not-reply" email address.  I think the contact page email makes sense, but there is also an argument for using the technical support email address.  We use the same email address for both so it doesn't make any difference to us.

I've also removed a blank line, and replaced "@" with "at" since using "@" is an unnecessary abbreviation.

## Tests performed

No tests.

## Types of changes

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -94,7 +94,7 @@ Dear %s,
 
 This email is to notify you that XDMoD has detected a change in your organization affiliation. We
 have taken steps to ensure that this is accurately reflected in our systems. If you have any questions
-or concerns please contact us @ %s.
+or concerns please contact us at %s.
 
 Thank You,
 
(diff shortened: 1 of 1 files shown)


classes/XDUser.php
```diff
@@ -2522,7 +2522,7 @@ public function synchronizeOrganization()
                         'body' => sprintf(
                             self::USER_NOTIFICATION_EMAIL,
                             $this->getFormalName(),
-                            \xd_utilities\getConfiguration('mailer', 'sender_email')
+                            \xd_utilities\getConfiguration('general', 'contact_page_recipient')
```
jpwhite4: suggest setting the replyAddress to \xd_utilities\getConfiguration('general', 'contact_page_recipient') so that if the user replys to the email it goes to the correct place.

jtpalmer: @jpwhite4 I've added the replyTo.  Should the body of the email even contain the address given this change?

jpwhite4: I'd leave the email address in the body too. It doesn't hurt to provide redundant info.

PR #1183, opened 2019-12-11
Removing cs,cd restrictions on getResources

ryanrath:
## Description
When this function is used in an XSEDE installation it's not just CS/CD that need to use this
function. Per a convo w/ @jpwhite4, this function does not need to be filtered. There will be an
accompanying PR in the AppKernels repo (https://github.com/ubccr/xdmod-appkernels/pull/74).

## Motivation and Context
To allow PO's to utilize the Center Report Card Dashboard component

## Tests performed
Open XDMoD: all automated fresh install / upgrade 

## Types of changes

- [] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/classes/XDUser.php
+++ b/classes/XDUser.php
@@ -2627,10 +2627,6 @@ SQL;
     public function getResources($resourceNames = array())
     {
         // We need to make sure that this function is only called for Center [Director|Staff]
-        if ( ! ( $this->hasAcl(ROLE_ID_CENTER_DIRECTOR) ||  $this->hasAcl(ROLE_ID_CENTER_STAFF) ) ) {
-            throw new Exception('Unable to complete action. User is not authorized.');
-        }
-
         $db = DB::factory('database');
 
         $query = <<<SQL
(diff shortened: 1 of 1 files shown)


classes/XDUser.php
```diff
@@ -2627,10 +2627,6 @@ public function isSticky()
     public function getResources($resourceNames = array())
     {
         // We need to make sure that this function is only called for Center [Director|Staff]
-        if ( ! ( $this->hasAcl(ROLE_ID_CENTER_DIRECTOR) ||  $this->hasAcl(ROLE_ID_CENTER_STAFF) ) ) {
```
jpwhite4: The comment should also be removed in this pull request since it is not longer accurate.

jpwhite4: Also the function documentation needs to be updated too to remove the line
```
 @throws Exception if the user this function is called for is not a Center [Director|Staff]
```

ryanrath: Thanks for the catch, updated in latest commit

PR #1188, opened 2019-12-12
Cleanup / make acl-config more resilient

ryanrath:
## Description
- `acl-config`:
  - Added Lock File protection to protect against multiple copies of `acl-config` running simultaneously 
  - Added a signal handler that should ensure the lock file is unlocked when `acl-config` is ctrl-c'd etc. 
  - We no longer need to run `acls-xdmod-management` multiple times within `acl-config`, we now run 
    it once after the backup tables are handled.
  - We've refactored the way that `acl-config` works, previously we wholesale Truncated / re- 
    populated the "managed" tables ( any table that we can completely recreate from available 
    configuration files ) and then re-populated the "non-managed" tables i.e. `user_acls` and 
    `user_acl_group_by_parameters`. This method did not allow us to utilize db transactions as both 
    `TRUNCATE` and DDL ( drop / create / update table ) force an implicit commit on execution. This 
    left us open to having the database in an inconsistent state. Now, instead of truncating / populating 
    the tables directly we instead create / populate a set of `staging` tables from the configuration files 
    first. This leaves the production tables in tact in the event some problem arises. Then, wrapped in 
    a db transaction, we DELETE FROM && INSERT INTO each production table from it's associated 
    staging table. 

## Motivation and Context
We have run into a number of problems lately that we believe to be caused by multiple copies of `acl-config` running at the same time and or interrupted runs of `acl-config`. It's hoped that these changes will resolve these issues. 

## Tests performed
- All automated tests in an OpenXDMoD docker: upgrade / fresh install
- Manual testing: 
  - Inserting an Exception at random times to make sure that existing functionality is not compromised with an incomplete execution. Also, to make sure that we can still utilize the lock file. 
  - Ctrl+C out of an execution at various times to make sure that we don't bork things with an incomplete execution. 

## Types of changes

## Checklist:

### DIFF
--- a/bin/acl-config
+++ b/bin/acl-config
@@ -142,6 +142,7 @@ try {
 
 /**
  *
+ * @throws Exception
  **/
 function main()
 {
@@ -584,55 +585,6 @@ SQL;
 }
 
 /**
- * Attempt to drop the tables managed by this script.
- *
- * @param iDatabase $db
- * @throws Exception if a problem occurs while executing sql.
- */
-function wipeManagedTables(iDatabase $db)
-{
-    global $dryRun, $log;
-
-    $log->info("Wiping Managed Tables...");
-
-    $tables = array(
-        'user_acl_group_by_parameters',
-        'user_acls',
-        'acl_tabs',
-        'tabs',
-        'acl_hierarchies',
-        'acl_group_bys',
-        'report_template_acls',
-        'acls',
-        'acl_types',
-        'group_bys',
-        'statistics',
-        'hierarchies',
-        'realms',
-        'modules',
-        'resource_type_realms'
-    );
-
-    foreach($tables as $table) {
-        $query = <<<SQL
-DROP TABLE $table;
-SQL;
-        if ($dryRun) {
-            $log->info("Would have dropped: $table");
-            continue;
-        }
-
-        $log->debug("Dropping Table\n$query");
-
-        $db->execute($query);
-
-        $log->debug("Table dropped");
-    }
-
-    $log->info("Managed Tables Wiped!");
-}
-
-/**
  * Re-populate the user related tables
  * @param iDatabase $db the database to utilize while re-populating the tables.
  * @param array $tables the tables to be re-populated.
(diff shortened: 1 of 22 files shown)


bin/acl-config
```diff
@@ -80,22 +111,42 @@ try {
         )
     );
 
+    // A handler function to make sure we cleanup the lock file, even if the process is killed.
+    $sigHandler = function ($sig) use ($lockFileHandle, $lockFile) {
+        @flock($lockFileHandle, LOCK_UN);
+        @fclose($lockFileHandle);
+        @unlink($lockFile);
```
jpwhite4: Is there a race condition here? You could unlink the file after a different process has locked it.

bin/acl-config
```diff
@@ -256,12 +304,12 @@ SQL;
      * their way back into the system.
      */
     $userAclRepopulate = <<<SQL
-    INSERT INTO user_acls (user_id, acl_id)
+    INSERT INTO user_acls_staging (user_id, acl_id)
```
jpwhite4: Where is user_acls_staging table created?

jpwhite4:
Why do you create _bkup tables? Surely they are not necessary if you are using staging tables.

bin/acl-config
```diff
@@ -35,7 +35,38 @@ $dryRun = false;
...
+    'acls',
+    'acl_group_bys',
+    'acl_hierarchies',
+    'tabs',
+    'acl_tabs',
+    'user_acls',
+    'user_acl_group_by_parameters'
+);
+
 try {
```
plessbd: looks like we are doing lockfiles in many different ways could we reuse this implementation?
`classes/ETL/LockFile.php`

bin/acl-config
ryanrath: It's created as a part of `acls-xdmod-management.json#/acls-xdmod-management/AclsStaginTableManagement`

bin/acl-config
ryanrath: Hrmm... maybe? it does have a couple of very ETL specific bits to it, but it may work out.

PR #1189, opened 2019-12-12
Only try to load metric explorer charts.

jpwhite4:
This mitigates the UI lockup when trying to click on a Usage chart from
a report thumbnail (asana issue
https://app.asana.com/0/159049597309611/1153563882786297)

A complete fix will be to actually support all types of charts that can
be shown (Usage, Custom Queries, App Kernel . etc. ). But this at least
ensures that the UI does not lockup.

### DIFF
--- a/html/gui/js/modules/dashboard/ReportThumbnailsComponent.js
+++ b/html/gui/js/modules/dashboard/ReportThumbnailsComponent.js
@@ -99,6 +99,13 @@ XDMoD.Module.Dashboard.ReportThumbnailsComponent = Ext.extend(Ext.Panel, {
             listeners: {
                 click: {
                     fn: function (dataView, index, node, e) {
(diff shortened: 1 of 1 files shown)


html/gui/js/modules/dashboard/ReportThumbnailsComponent.js
```diff
@@ -99,6 +99,13 @@ XDMoD.Module.Dashboard.ReportThumbnailsComponent = Ext.extend(Ext.Panel, {
             listeners: {
                 click: {
                     fn: function (dataView, index, node, e) {
+                        var config = JSON.parse(JSON.stringify(dataView.store.data.items[index].json.chart_id));
```
plessbd: i do not like this at all...

jpwhite4: Niether do I. In fact it appears that the store is not being setup and used according to best practice. Probably will leave this for 8.6 and refactor for 9.0

PR #1191, opened 2019-12-13
Fix pie chart datalabel formatting.

jpwhite4:
The word wrap algorithm for pie chart datalabels would truncate long
lines before word wrapping. The truncation algorithm did not take into
account word boundaries but the wrapping one did. This meant that, even
though the max length is set to the equivalent of two lines of text,
sometimes the result would be on three lines (for a total of four lines
when the values are included). The highchart datalabel layout engine
only leaves enough room for 1 line of text so the charts would be quite
ugly. Furthermore no indication of truncation was given.

For example the text: "Computer and Information Science and Engineering"
would get mangled to:

```
Computer and
Information Science
an
```

The new code runs the word wrap first. If the string spans more than two
lines then ellipses are added to the end of the second line to indicate
that data has been chopped. The padding between datalabels has also been
increased to prevent overlap.

These changes only impact pie charts, but will alter all pie charts in
XDMoD.

Old:

image

New:

image

Old:

image

New:

image


plessbd:
you are going to need to update the usagechart artifacts

jpwhite4:
The test artifacts have been updated. Note that only pie charts have changed and only those pie chart that have datalabels (either show_aggregate_labels=y or show_error_labels=y

html/gui/js/HighChartWrapper.js
```diff
@@ -121,11 +121,12 @@ XDMoD.utils.createChart = function (chartOptions, extraHandlers) {
                             return this.y.toString();
                         }
                         if (this.series.type === 'pie') {
-                            var x = this.point.name;
-                            if (this.point.name.length > settings.maxL + 3) {
-                                x = this.point.name.substring(0, settings.maxL - 3);
+                            var lines = this.point.name.wordWrap(settings.wrap, '\t').split('\t');
+                            if (lines.length > 2) {
```
jtpalmer: If mostly just curious, but did you find and reasoning behind the calculations of `wrap` and `maxL`.  It appears that a single long line was allowable (if there are ten or fewer data elements).  Would we ever want to have that behavior?  And would we ever want to make the number of lines configurable?

PR #1192, opened 2019-12-13
Groupby refactor

plessbd:
update groupbys and their configurations


classes/Realm/iGroupBy.php
```diff
@@ -0,0 +1,362 @@
...
+     *    filter with the (possibly multiple) values.
+     *
+     * @param array $request The HTTP request
+     *
+     * @return array An arrray of label strings, typically of the form "label = string"
+     */
+
+    public function generateQueryParameterLabelsFromRequest(array $request);
+
+    // are these called with all arguments?
```
jtpalmer: Well, are they?

classes/Realm/iGroupBy.php
```diff
@@ -0,0 +1,362 @@
...
+     * @return string
+     */
+
+    public function getDefaultDatasetType();
+
+    /**
+     * @param string $dataset_type The type of dataset that we are querying (e.g., timeseries or
+     *   aggregate).
+     *
+     * @return string
```
jtpalmer: What are valid display types?

classes/User/Elements/QueryDescripter.php
```diff
@@ -2,12 +2,17 @@
...
 
+use \DataWarehouse\Query\AggregateQuery;
+use \DataWarehouse\Query\TimeseriesQuery;
+
 class QueryDescripter
 {
 
     /**
-     * @var string
+     * @var Realm The Realm that we are querying.
```
jtpalmer:
```suggestion
     * @var \Realm\Realm The Realm that we are querying.
```

plessbd:
@jtpalmer a lot of the code was done as part of a different branch and a lot of details were in the commits from that

https://github.com/ubccr/xdmod/compare/feature/groupby-refactor

I will look through to find what I can.

bin/acl-config
```diff
@@ -207,16 +212,14 @@ function main()
...
-                    'statistics' => json_decode(json_encode($statistics), true)
-                );
-            }
+        foreach ( $realmObjects as $realmId => $realmObj ) {
+            $groupBys = $realmObj->getGroupByIds();
+            $statistics = $realmObj->getStatisticIds();
+            $results[$module]['realms'][$realmId] = array(
+                'display_name' => $realmObj->getName(),
+                'group_bys' => json_decode(json_encode($groupBys), true),
+                'statistics' => json_decode(json_encode($statistics), true)
```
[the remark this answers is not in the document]
plessbd: @ryanrath would be best to answer this as this predates the refactor.

bin/acl-config
```diff
@@ -1648,7 +1639,7 @@ function createRealm(iDatabase $db, $moduleName, $realmDisplay)
 SQL;
 
     $params = array(
-        ':realm_name' => $realmName,
+        ':realm_name' => $realmId,
```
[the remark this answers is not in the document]
plessbd: @smgallo changed this I think to differentiate between the name that is used as an id and the display

bin/acl-config
```diff
@@ -1735,19 +1726,18 @@ SQL;
     }
     $totalInserted = 0;
 
-    foreach ($groupBys as $groupBy) {
-        $groupByName = $groupBy['name'];
+    foreach ($groupBys as $groupById) {
 
         $parameters = array(
-            ':group_by_name' => $groupByName,
+            ':group_by_name' => $groupById,
```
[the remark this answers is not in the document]
plessbd: same as previous I think this was to help differentiate

bin/acl-config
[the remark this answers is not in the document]
smgallo: Correct. I used the suffixes `Id` and `Name` to differentiate between internal identifiers and and human-readable names.

classes/Realm/Realm.php
```diff
@@ -0,0 +1,815 @@
...
+                        $orderA = ( isset($a->order) ? $a->order : 0 );
+                        $orderB = ( isset($b->order) ? $b->order : 0 );
+
+                        if ( $orderA < $orderB ) {
+                            return -1;
+                        } elseif ( $orderA > $orderB ) {
+                            return 1;
+                        } else {
+                            return 0;
+                        }
```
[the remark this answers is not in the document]
plessbd: I didnt that was in until PHP 7

classes/Realm/iGroupBy.php
```diff
@@ -0,0 +1,362 @@
...
+    public static function factory($shortName, \stdClass $config, Realm $realm, Logger $logger = null);
+
+    /**
+     * @return Realm The realm that this GroupBy is associated with.
+     */
+
+    public function getRealm();
+
+    /**
+     * Note: Was getName()
```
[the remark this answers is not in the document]
smgallo: This was an internal note during the refactor, it can be removed.

classes/Realm/iGroupBy.php
smgallo: Another refactor note, this can be removed.

bin/acl-config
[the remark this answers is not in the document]
smgallo: There was a lot of confusion in the code between names and what I'd consider to be internal identifiers. Another pass through to clean some of this up is probably a good idea.

bin/acl-config
```diff
@@ -1813,20 +1803,18 @@ SQL;
...
     }
     $totalInserted = 0;
-    foreach ($statistics as $statistic) {
-
-        $statisticName = $statistic['name'];
+    foreach ($statistics as $statisticId) {
 
         $parameters = array(
-            ':statistic_name' => $statisticName,
+            ':statistic_name' => $statisticId,
```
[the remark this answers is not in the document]
plessbd: @smgallo changed this I think to differentiate between the name that is used as an id and the display

classes/DB/Loggable.php
```diff
@@ -1,18 +0,0 @@
...
-class Loggable
-{
-	protected $_logger = null;
-    public function setLogger(Log $logger)
-    {
-        $this->_logger = $logger;
-    }
-} //Ingestor
-
-?>
```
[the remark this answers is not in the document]
smgallo: It was part of a general code cleanup that happened before the group by. Must have gotten left out of the original cleanup.

classes/DataWarehouse/Data/SimpleDataset.php
```diff
@@ -341,21 +331,20 @@ public function export($export_title = 'title')
 
         foreach ($group_bys as $group_by) {
             $headers[]
-                = $group_by->getName() === 'none'
+                = $group_by->getId() === 'none'
```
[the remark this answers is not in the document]
smgallo: If I remember correctly, it is used as a special case.

classes/Realm/Realm.php
```diff
@@ -0,0 +1,815 @@
+<?php
+/**
+ * A Realm is a collection of data, descriptive attributes, and statistics.
+ *
+ * ToDo: Use APCu to store a cache of realm objects and possibly group by and statistic objects.
+ */
+
+namespace Realm;
```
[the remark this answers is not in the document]
smgallo: Up to you guys as to how you want to organize it. I felt that since Realms are central to so much of XDMoD that it was important enough to warrant a top-level namespace.

classes/Realm/Statistic.php
```diff
@@ -0,0 +1,494 @@
...
+        return $this->showInCatalog;
+    }
+
+    /**
+     * @see iStatistic::__toString()
+     */
+
+    public function __toString()
+    {
+        return sprintf("%s->Statistic(id=%s, formula=%s)", $this->realm, $this->id, $this->formula);
```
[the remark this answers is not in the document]
smgallo: I believe only for logging and debugging.

classes/DB/FilterListBuilder.php
```diff
@@ -325,31 +332,38 @@ private function createDimensionQuery(Query $realmQuery, GroupBy $groupBy)
...
         // table and column names,
-        $dimensionName = $groupBy->getName();
+        $dimensionId = $groupBy->getId();
         $dimensionQuery = $this->createDimensionQuery($realmQuery, $groupBy);
         $dimensionQueryTables = $dimensionQuery->getSelectTables();
         $dimensionQueryFields = $dimensionQuery->getSelectFields();
         $dimensionTableStringComponents = explode(' ', $dimensionQueryTables[1]);
         $dimensionTable = $dimensionTableStringComponents[0];
-        preg_match('/\.(\S+)\s/', $dimensionQueryFields['id'], $dimensionColumnMatches);
+        preg_match('/\.(\S+)\s/', $dimensionQueryFields[ sprintf('%s_id', $dimensionId) ], $dimensionColumnMatches);
```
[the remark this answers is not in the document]
plessbd: I agree we should, but I dont think it belongs as part of the refactor.
I will create a "cleanup" task for things that are found in this PR that can be done separately (the deletion of the logger file that isnt used being another)

classes/Realm/Statistic.php
```diff
@@ -0,0 +1,494 @@
...
+                    break;
+                case 'aggregate_formula':
+                    $this->aggregateFormula = trim($value);
+                    break;
+                case 'data_sort_order':
+                    // The sort order is specified in the JSON config file as the string
+                    // representation of a PHP constant so convert it to an integer in order to
+                    // properly use it. See https://php.net/manual/en/function.array-multisort.php
+                    $sortOrder = null;
+                    eval('$sortOrder = $value');
```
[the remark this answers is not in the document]
smgallo: As far as I could tell when it was written. For historical reasons the value is actually a string such as "SORT_REGULAR" or "SORT_NUMERIC" that needs to be converted to the PHP constant.

classes/DB/Loggable.php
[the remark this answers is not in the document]
plessbd: I will create a "cleanup" task that removes this outside of this pull request

classes/Realm/iGroupBy.php
```diff
@@ -0,0 +1,362 @@
...
+    public function getDefaultShowErrorLabels();
+
+    /**
+     * @return string
+     */
+
+    public function getDefaultEnableErrors();
+
+    /**
+     * @return string
```
jtpalmer:
```suggestion
     * @return string "y" or "n".
```

classes/Realm/iGroupBy.php
jtpalmer: I see `h_bar` and `line`, but I'm assuming there's a full list somewhere.

classes/Realm/iRealm.php
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * @return string The name of the module that defined this Realm. The default is "xdmod", the
+     *   core XDMoD module.
+     */
+
+    public function getModuleName();
+
+    /**
+     * @return boolean TRUE if the realm is disabled and should not be visible iat all to the
```
jtpalmer:
```suggestion
     * @return boolean TRUE if the realm is disabled and should not be visible at all to the
```

classes/Realm/iRealm.php
```diff
@@ -0,0 +1,333 @@
...
+     */
+
+    public function getStatisticObject($id);
+
+    /**
+     * @return string The default weighting statistic for this realm. This can be overriden by
+     *   individual statistics.
+     */
+
+    public function getDefaultWeighgtStatName();
```
jtpalmer:
```suggestion
    public function getDefaultWeightStatName();
```

classes/Realm/Realm.php
```diff
@@ -0,0 +1,815 @@
...
+                sprintf("Attempt to instantiate undefined Statistic class %s", $className)
+            );
+        }
+
+        $factory = sprintf('%s::factory', $className);
+        return forward_static_call($factory, $shortName, $this->statisticConfigs->$shortName, $this, $this->logger);
+    }
+
+    /**
+     * @see iRealm::getDefaultWeighgtStatName()
```
jtpalmer:
```suggestion
     * @see iRealm::getDefaultWeightStatName()
```

classes/Realm/Realm.php
```diff
@@ -0,0 +1,815 @@
...
+     */
+
+    public function getDefaultWeighgtStatName()
```
jtpalmer:
```suggestion
    public function getDefaultWeightStatName()
```

classes/DataWarehouse/Query/Jobs/RawData.php
```diff
@@ -12,29 +12,42 @@
...
+        $realmId,
+        $aggregationUnitName,
+        $startDate,
+        $endDate,
+        $groupById = null,
+        $statisticId = null,
+        array $parameters = array(),
+        Log $logger = null
     ) {
+        $realmId = 'Jobs';
```
jtpalmer: Is the `RawData` class going to get the same refactoring as the other group by classes in the near future?  If not in this pull request then why is the `$realmId` a constructor parameter here if it's immediately overridden?

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1248,7 +1264,7 @@ public function processJobSearch(Request $request, Application $app, XDUser $use
             throw new BadRequestException('Invalid search parameters specified in params object');
         } else {
             $QueryClass = "\\DataWarehouse\\Query\\$realm\\RawData";
-            $query = new $QueryClass("day", $startDate, $endDate, null, "", array());
+            $query = new $QueryClass($realm, "day", $startDate, $endDate, null, "", array());
```
jtpalmer: The `$realm` here looks redundant, but I've commented on that separately.

classes/User/Elements/QueryDescripter.php
```diff
@@ -252,79 +240,81 @@ public function getAllQueries(
...
 
+    /**
+     * Generate a list of query filter objects for all group-bys associated with the current realm
+     * based on the request.
+     *
+     * Note: This is currently only used by html/controllers/ui_data/summary3.php
+     *
+     * @param array $request An associative array of GET/POST request variables
+     *
+     * @return array An arrray of \DataWarehouse\Query\Model\Parameter objects
```
jtpalmer:
```suggestion
     * @return array An array of \DataWarehouse\Query\Model\Parameter objects
```

or

```suggestion
     * @return \DataWarehouse\Query\Model\Parameter[]
```

classes/User/Elements/QueryDescripter.php
```diff
@@ -252,79 +240,81 @@ public function getAllQueries(
...
 
+    /**
+     * Generate a list of query filter objects for all group-bys associated with the current realm
+     * based on the request.
+     *
+     * Note: This is currently only used in classes/DataWarehouse/QueryBuilder.php
+     *
+     * @param array $request An associative array of GET/POST request variables
+     *
+     * @return array An arrray of label strings
```
jtpalmer:
```suggestion
     * @return array An array of label strings
```

classes/Realm/iGroupBy.php
jtpalmer: I was in such suspense, but I guess the mystery was solved long ago.

tests/unit/lib/Realm/StatisticTest.php
```diff
@@ -0,0 +1,256 @@
+<?php
+/**
+ * Unit tests for Statistic class.
+ */
+
+namespace UnitTesting\Realm;
+
+use CCR\Log as Logger;
+use \Realm\Realm;
```
jtpalmer:
```suggestion
use Realm\Realm;
```

tests/component/phpunit.xml.dist
```diff
@@ -10,7 +10,6 @@
          convertWarningsToExceptions="true"
          forceCoversAnnotation="false"
          mapTestClassNameToCoveredClassName="false"
-         printerClass="PHPUnit_TextUI_ResultPrinter"
```
jtpalmer: Was this intentionally changed here?  I don't know how this changes the results.

tests/integration/phpunit.xml.dist
```diff
@@ -11,7 +11,6 @@
          convertWarningsToExceptions="true"
          forceCoversAnnotation="false"
          mapTestClassNameToCoveredClassName="false"
-         printerClass="PHPUnit_TextUI_ResultPrinter"
```
jtpalmer: Same question as the other.

tests/component/lib/Export/BatchProcessTest.php
```diff
@@ -88,8 +88,8 @@ public static function setUpBeforeClass()
         list($user) = self::$dbh->query("SELECT id FROM Users WHERE username = 'normaluser'");
 
         // Create records and set one to be expiring.
-        self::$submittedRequestId = self::$queryHandler->createRequestRecord($user['id'], 'jobs', '2019-01-01', '2019-01-31', 'CSV');
-        self::$expiringRequestId = self::$queryHandler->createRequestRecord($user['id'], 'jobs', '2019-01-01', '2019-01-31', 'JSON');
+        self::$submittedRequestId = self::$queryHandler->createRequestRecord($user['id'], 'Jobs', '2019-01-01', '2019-01-31', 'CSV');
+        self::$expiringRequestId = self::$queryHandler->createRequestRecord($user['id'], 'Jobs', '2019-01-01', '2019-01-31', 'JSON');
```
jtpalmer: If this is changing then it's all but guaranteed that you're going to need to add a migration for the data in the `moddb`.`batch_export_requests` table.

classes/Rest/Controllers/WarehouseControllerProvider.php
jtpalmer: Emphasis on "looks", I realize that it is necessary as is.

bin/acl-config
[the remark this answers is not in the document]
smgallo: The configuration files use short string identifiers as the ids, which is what the realms are based on since that is how they are constructed.

classes/Realm/iGroupBy.php
smgallo: Sorry to let you down. I'll try to have a better story next time. :)

classes/DataWarehouse/Access/MetricExplorer.php
```diff
@@ -232,30 +232,29 @@ public function get_data($user)
...
 
                 $query->addGroupBy($data_description->group_by);
                 $query->addStat($data_description->metric);
 
                 if ($data_description->std_err == 1) {
-                    try {
-                        $query->addStat('sem_'.$data_description->metric);
+                    $semStatId = \Realm\Realm::getStandardErrorStatisticFromStatistic(
+                        $data_description->metric
+                    );
```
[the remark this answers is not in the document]
smgallo: There is likely old code that I did not want to touch as part of this pull request. Actually, there was a lot that I _wanted_ to change but was trying not to allow too much scope creep since it was already a large change.

For the definition see https://docs.google.com/document/d/1ZPxofHgDfoaCgraY86GM1hvO9Z0cjyZtMjX7qtMmPxE

Metric: A measurement of a particular quantity associated with a fact such as Total CPU Hours or Wait Time.
Statistic: A function applied to one or more metrics in a subgroup to generate a single value. For example, minimum, maximum, average, standard deviation, etc.

classes/Realm/iRealm.php
smgallo: I don't know that there is but a class constant with a list of acceptable names is a good idea.

classes/DataWarehouse/Query/Jobs/RawData.php
smgallo: The `Query` class was going to be next on my list since it and it's extending classes are a mess. They were intentionally left out of this PR.

bin/acl-config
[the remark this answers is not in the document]
smgallo: If it is no longer needed, this can be changed.

tests/component/phpunit.xml.dist
plessbd: Yes it was, with this printer their the tests that either failed or errored were not printed.  This was making degugging difficult so I removed the printer and the output had those test names printed properly.

classes/Realm/Realm.php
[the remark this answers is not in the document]
jpwhite4: Could also do
```
$a - $b
```
which will return a positive, zero or negative number depending on the values

classes/Realm/iGroupBy.php
plessbd: I assume it is
https://github.com/ubccr/xdmod/blob/18d551a9cb3284938857399429ca699970dacfaf/classes/DataWarehouse/VisualizationBuilder.php#L22-L28

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1488 @@
...
+                    }
+                    break;
+                case 'chart_display_type_map':
+                    break;
+                case 'data_sort_order':
+                    // The sort order is specified in the JSON config file as the string
+                    // representation of a PHP constant so convert it to an integer in order to
+                    // properly use it. See https://php.net/manual/en/function.array-multisort.php
+                    $sortOrder = null;
+                    eval('$sortOrder = $value;');
```
jtpalmer: It's still here!

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1013,21 +1036,37 @@ public function getDimensionName(Request $request, Application $app, $dimensionI
     public function getDimensionValueName(Request $request, Application $app, $dimensionId, $valueId)
     {
         $user = $this->getUserFromRequest($request);
+        $realmName = $request->get('realm');
+        Acls::hasDataAccess($user, $realmName, $dimensionId);
+
         $valueName = MetricExplorer::getDimensionValueName($user, $dimensionId, $valueId);
         $success = !empty($valueName);
```
jtpalmer: This assignment is now "useless" (https://sonarcloud.io/project/issues?id=ubccr_xdmod&issues=AXB9lSAVTP8OQjfieZ4G&open=AXB9lSAVTP8OQjfieZ4G&pullRequest=1192) since it's assigned again before being used.

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1490 @@
...
+     * @param string $formula The SQL column or formula
+     *
+     * @return string The formula with aliases changed to the group by table alias, or with an alias
+     *   added if only a column was specified.
+     *
+     * @throws Exception if the formula had an aliased column name but the table that it referred to
+     *   did not match the group by attribute table.
+     */
+
+    protected function verifyAndReplaceTableAlias($formula, \DataWarehouse\Query\iQuery $query)
```
jtpalmer: `$query` is not documented above.

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1490 @@
...
+    {
+        $formula = $this->realm->getVariableStore()->substitute($formula);
+        if ( $this->isAggregationUnit && 'none' == $this->id ) {
+            return $formula;
+        }
+
+        $matches = array();
+        if ( 0 === preg_match_all('/([a-zA-Z0-9$_]+\.)?([a-zA-Z0-9$_]+\.[a-zA-Z0-9$_]+)/', $formula, $matches, PREG_SET_ORDER) ) {
```
jtpalmer: Are quoted identifiers intentionally disallowed in formulae?

bin/acl-config
```diff
@@ -1507,32 +1512,17 @@ SQL;
...
             $params = array(
                 ':module_name' => $module,
-                ':realm_name' => $realmName,
+                ':realm_name' => $realm,
                 ':acl_name' => $acl,
                 ':group_by_name' => $groupBy,
-                ':statistic_name' => $statisticName,
-                ':visible' => ( $shown ? 1 : 0 ),
+                ':statistic_name' => $statisticId,
+                ':visible' => 1,
```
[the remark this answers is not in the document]
plessbd: @ryanrath I cant remember how this works...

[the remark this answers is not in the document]
spark0r: how Visible works?

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1490 @@
...
+
+    public function isAggregationUnit()
+    {
+        return $this->isAggregationUnit;
+    }
+
+    /*
+     * DEVNOTE: Is there really any reason why we should not always treat these methods as $multi =
+     * true and prefix them by the group by id? It would likely make the query more readable.
+     */
```
jtpalmer: Remove development note.

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1490 @@
...
+        $labelList[] = sprintf('%s = %s', $this->getName(), $labelString);
+
+        $this->logger->debug(
+            sprintf('%s Generated query parameter labels %s', $this, implode('; ', $labelList))
+        );
+
+        return $labelList;
+    }
+
+    // are these called with all arguments?
```
jtpalmer: Remove development note.

classes/Realm/GroupBy.php
```diff
@@ -0,0 +1,1490 @@
...
+    protected $sortOrder = SORT_DESC;
+
+    /**
+     * @var boolean Set to true if this group by should not be utilized.
+     */
+
+    protected $disabled = false;
+
+    /**
+     * @see iRealm::factory()
```
jtpalmer:
```suggestion
     * @see iGroupBy::factory()
```

classes/Realm/GroupBy.php
plessbd: @smgallo Do you remember why you made this specific regular expression?

plessbd: from what @smgallo said, no.
However, I am fine with waiting to change it until we have a reason to.

smgallo: That expression was to check to see if an alias was being used in the attribute value query and replace it with the table name. In the design document under "GroupBy Configuration Options" see Examples > attribute_values_query.

```
An example where multiple tables are joined is GroupBy resource. Note that the `rf` alias in `CONCAT(rf.name, '-', rf.code)` will be replaced by the attribute table when applying a GroupBy to the constructed query.

```javascript
"attribute_values_query": {
    "records": {
        "id": "rf.id",
        "short_name": "rf.code",
        "name": "CONCAT(rf.name, '-', rf.code)",
        "order_id": "id"
    },
    "joins": [
        {
            "name": "resourcefact",
            "alias": "rf"
        },
        {
            "name": "resourcespecs",
            "alias": "rs",
            "on": "rf.id = rs.resource_id"
        }
    ],
    "where": [
        "rs.processors IS NOT NULL"
    ],
    "orderby": [
        "rf.code",
        "rf.name"
    ]
}
```

jtpalmer: There's no need to change it. Everyone just needs to be aware that if you ever use `` rf.`code` `` it won't work as expected. e.g. `` "CONCAT(rf.`name`, '-', rf.`code`)" `` would not match the regular expression.

bin/acl-config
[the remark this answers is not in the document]
jtpalmer: How was this resolved?  Everything is always visble?

classes/Realm/GroupBy.php
```diff
@@ -1079,6 +1072,8 @@ public function applyTo(\DataWarehouse\Query\iQuery $query, $multi_group = false
      *
      * @param string $formula The SQL column or formula
      *
+     * @param Query $query used when using an aggrigation unit
```
jtpalmer:
```suggestion
     * @param Query $query used when using an aggregation unit
```

jtpalmer: It might be more useful to say "Used to determine the table alias when the GroupBy is an aggregation unit."

jtpalmer:
Is the ordering in these charts intentionally changed?

Before:
CPU_Hours__Total__by_Job_Size_2020-02-01_to_2020-02-29_aggregate (1)

After:
CPU_Hours__Total__by_Job_Size_2020-02-01_to_2020-02-29_aggregate

jtpalmer:
"Provider" is listed in the Cloud drill-down, but not in the list of Group Bys. (It doesn't work.)

Usage-Metrics

Usage-Drill-Down

classes/Realm/iStatistic.php
```diff
@@ -0,0 +1,182 @@
...
+
+    public function getUnit();
+
+    /**
+     * In order to provide increased flexibility, the formulas specified by a statistic can make use
+     * of variable substitution. When called, a number of variables set in the query VariableStore
+     * are applied to the formula. The generic Statistic::getFormula() can be overriden as needed to
+     * apply advanced logic.
+     *
+     * @param Query $query The target Query that this formula will be embedded in. Note that the
```
jtpalmer:
```suggestion
     * @param \DataWarehouse\Query\iQuery $query The target Query that this formula will be embedded in. Note that the
```

classes/DB/FilterListBuilder.php
```diff
@@ -311,10 +311,16 @@ private function checkDimensionForRoles(GroupBy $groupBy)
      * @param  GroupBy $groupBy    The GroupBy to construct the Query around.
      * @return Query               A Query constructed around $groupBy.
      */
-    private function createDimensionQuery(Query $realmQuery, GroupBy $groupBy)
+    private function createDimensionQuery(iQuery $realmQuery, GroupBy $groupBy)
```
jtpalmer: Change `@param` above to use `iQuery`.

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
```diff
@@ -15,9 +15,12 @@ public function execute()
     {
         parent::execute();
 
+        $dbh = DB::factory('datawarehouse');
+        $dbh->execute("UPDATE moddb.batch_export_requests SET realm = 'Jobs' WHERE realm = 'jobs'");
+        $dbh->execute("UPDATE moddb.batch_export_requests SET realm = 'SUPREMM' WHERE realm = 'supremm'");
```
jtpalmer: Make sure the corresponding change happens for XSEDE.  There are several more realms supported there.

tests/unit/lib/ETL/DbModel/DbModelTest.php
```diff
@@ -35,7 +35,7 @@ public static function setUpBeforeClass()
         $conf = array(
             'db' => false,
             'mail' => false,
-            'consoleLogLevel' => Log::WARNING
+            'consoleLogLevel' => Log::EMERG
```
jtpalmer: I personally don't think that these tests should be writing anything to the console, but I don't think changing it as part of this pull request serves a purpose.  Was this breaking something?

classes/DB/FilterListBuilder.php
```diff
@@ -73,25 +64,33 @@ public function buildAllLists()
...
      * Build filter lists for the given dimension.
      *
+     * NOTE: This function does not support dimensions with multi-column keys even though the
+     *       GroupBy classes do. It must be refactored in order to support them. -SMG 2019-09-09
+     *
      * @param Query   $realmQuery A query for the realm the dimension is in.
      * @param GroupBy $groupBy    The dimension's GroupBy to build lists for.
      */
-    public function buildDimensionLists(Query $realmQuery, GroupBy $groupBy)
+    public function buildDimensionLists(iQuery $realmQuery, GroupBy $groupBy)
```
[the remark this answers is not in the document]
smgallo: We should use `iGroupBy`. That's the contract and if something needs additional functionality not specified in the interface it's breaking the contract.

classes/DataWarehouse/Query/iQuery.php
```diff
@@ -0,0 +1,134 @@
...
+     * @param string $groupById The short identifier for the GroupBy that this query will use.
+     * @param Log|null $logger A Log instance that will be utilized during processing.
+     * @param string $statisticId A short identifier for an initial Statistic that will be returned
+     *   by this query. Note that 'all' can be used to include all available statistics in the
+     *   specified Realm.
+     * @param array $parameters An array of \DataWarehouse\Query\Model\Parameter objects that will
+     *   be used to add WHERE conditions to the Query.
+     */
+
+     // If there are statistics that should always be added for a specific Realm, specify them in
```
jtpalmer:
```suggestion
    // If there are statistics that should always be added for a specific Realm, specify them in
```

classes/DataWarehouse/Query/iQuery.php
```diff
@@ -0,0 +1,134 @@
...
+        Logger $logger = null
+    );
+
+    /**
+     * Update the values of the internal VariableStore. Since the contents of the Query class are
+     * volitile and may be modified on the fly, this should be called prior to performing variable
+     * substitution. Note that rather than clearing and setting variables, we are using overwrite()
+     * so we do not clear variables that may be set elsewhere.
+     *
+     * @return VariableStore The updated variable store object
```
jtpalmer:
```suggestion
     * @return \ETL\VariableStore The updated variable store object
```

classes/DataWarehouse/Query/iQuery.php
```diff
@@ -0,0 +1,134 @@
...
+     */
+
+    public function updateVariableStore();
+
+    /**
+     * @return VariableStore The current variable store for this query. Note that it is safer to
```
jtpalmer:
```suggestion
     * @return \ETL\VariableStore The current variable store for this query. Note that it is safer to
```

classes/DataWarehouse/Query/iQuery.php
```diff
@@ -0,0 +1,134 @@
...
+     *    access the variable store via updateVariableStore() so that its data is updated.
+     */
+
+    public function getVariableStore();
+
+    /**
+     * @return Realm The object that represents the realm that is query is generating data for.
```
jtpalmer:
```suggestion
     * @return \Realm\Realm The object that represents the realm that is query is generating data for.
```

classes/DataWarehouse/Visualization/HighChartTimeseries2.php
```diff
@@ -169,16 +167,16 @@ public function configure(
 
             // Determine statistic name. In this case use the Aggregate classname to get the stat.
             // (Speculative). August 2015
```
jtpalmer: Is this still speculative?

classes/DataWarehouse/Visualization/HighChartTimeseries2.php
```diff
@@ -169,16 +167,16 @@ public function configure(
 
             // Determine statistic name. In this case use the Aggregate classname to get the stat.
```
jtpalmer: No longer using the aggregate class name.

tests/unit/lib/ETL/DbModel/DbModelTest.php
smgallo: Those are there (and quite useful) for debugging during test writing and verification. It doesn't need to write anything during normal operation.

bin/acl-config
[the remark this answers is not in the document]
plessbd: sorry, this was me testing things out.  I am working on a "better" way to do this.  At another point I think this should actually be removed from the Realm completely and moved to the acl's but that will be for later

plessbd:
> Is the ordering in these charts intentionally changed?
> 
> Before:
> CPU_Hours__Total__by_Job_Size_2020-02-01_to_2020-02-29_aggregate (1)
> 
> After:
> CPU_Hours__Total__by_Job_Size_2020-02-01_to_2020-02-29_aggregate

nope, this is a misconfiguration, I will update this

classes/DB/Aggregator.php
```diff
@@ -9,24 +9,17 @@
  */
 use CCR\Loggable;
 
+use Log as Logger;
+
 class Aggregator extends Loggable
 {
-    public static $__initialized;
```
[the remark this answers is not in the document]
jtpalmer: On closer inspection I don't think that class is being used anymore.  I assume job time series aggregation is handled via ETL, but have not confirmed this.

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
```diff
@@ -15,9 +15,12 @@ public function execute()
...
+
         // NOTE: It's possible that this should be removed after the GroupBy
         // refactoring is merged.
-        $dbh = DB::factory('datawarehouse');
         $dbh->execute('DROP TABLE IF EXISTS modw_filters.Storage_person___username');
         $dbh->execute('DROP TABLE IF EXISTS modw_filters.Storage_pi___username');
         $dbh->execute('DROP TABLE IF EXISTS modw_filters.Storage_username');
```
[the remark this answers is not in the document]
plessbd: I believe it still is, I can remove the comment

classes/DB/Aggregator.php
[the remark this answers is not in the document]
plessbd: I do not believe it is being used I think @smgallo forgot to remove it when he switched things to etlv2

plessbd:
> "Provider" is listed in the Cloud drill-down, but not in the list of Group Bys. (It doesn't work.)
> 
> Usage-Metrics
> 
> Usage-Drill-Down

should be fixed now

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
```diff
@@ -15,9 +15,10 @@ public function execute()
     {
         parent::execute();
 
-        // NOTE: It's possible that this should be removed after the GroupBy
-        // refactoring is merged.
         $dbh = DB::factory('datawarehouse');
+        $dbh->execute("UPDATE moddb.batch_export_requests SET realm = 'Jobs' WHERE realm = 'jobs'");
+        $dbh->execute("UPDATE moddb.batch_export_requests SET realm = 'SUPREMM' WHERE realm = 'supremm'");
+
```
jtpalmer: Lines below (`DROP TABLE` statements) are no longer necessary if #1245 is merged as is.

classes/DataWarehouse/Data/SimpleTimeseriesDataset.php
```diff
@@ -876,15 +842,15 @@ public function getSummarizedColumn(
                             );
 
                             // if we are taking AVG, correct it to 'avg of n others', n is $normalizeBy value
-                            $dataCurrentValue = $dataCurrentValue * ($countTsCurrentValue / $normalizeBy); 
+                            $dataCurrentValue = $dataCurrentValue * ($countTsCurrentValue / $normalizeBy);
```
jtpalmer: Note that #1248 changes this calculation (https://github.com/ubccr/xdmod/pull/1248/files#diff-0244e1ca34cbd2da230ea9789d42d71eR860).

classes/DB/FilterListBuilder.php
```diff
@@ -35,23 +36,23 @@ class FilterListBuilder extends Loggable
      */
     private static $rolesDimensionNames = null;
 
+    /**
+     * @var \Realm\iRealm The Realm that we are currently operating on.
+     */
+
+    private $currentRealm = null;
```
jpwhite4: I think this should not be a private member variable. It seems to be only used as a function local variable in the buildRealmLists() function

classes/DB/FilterListBuilder.php
```diff
@@ -63,25 +64,33 @@ public function buildAllLists()
...
-     * @param Query   $realmQuery A query for the realm the dimension is in.
-     * @param GroupBy $groupBy    The dimension's GroupBy to build lists for.
+     * NOTE: This function does not support dimensions with multi-column keys even though the
+     *       GroupBy classes do. It must be refactored in order to support them. -SMG 2019-09-09
+     *
+     * @param iQuery   $realmQuery A query for the realm the dimension is in.
+     * @param iGroupBy $groupBy    The dimension's GroupBy to build lists for.
      */
-    public function buildDimensionLists(Query $realmQuery, GroupBy $groupBy)
+    public function buildDimensionLists(iQuery $realmQuery, iGroupBy $groupBy)
```
jpwhite4: Seems dangerous having this public since it relies on the private member variable $this->currentRealm being set to a specific value.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -88,32 +90,37 @@ private function getSummaryCharts(XDUser $user) {
...
                     $statUsageChartSettings['swap_xy'] = false;
                 }
-                $errorstat = 'sem_' . $userStatistic;
-                if (in_array($errorstat, array_keys($usageRealmAggregateClass::getRegisteredStatistics())) ) {
+
+                $errorstat = \Realm\Realm::getStandardErrorStatisticFromStatistic(
+                    $userStatistic
+                );
+
+                if (in_array($errorstat, $realm->getStatisticIds()) ) {
```
jpwhite4: In the metric explorer code that does the same thing it runs:
```
if ($query->getRealm()->statisticExists($semStatId)) {
```
Whereas here is runs:
```
in_array($errorstat, $realm->getStatisticIds())
```
why the different function calls that do the same thing?

I note that the metric explorer one is probably faster to run since it doesn't require building a temporary array and then searching through it.

classes/DataWarehouse/Access/Usage.php
```diff
@@ -176,8 +181,12 @@ public function getCharts(XDUser $user, $chartsKey = 'data') {
 
             // Get the request's group by.
             $usageGroupBy = \xd_utilities\array_get($this->request, 'group_by', 'none');
-            $usageGroupByObject = $usageRealmAggregateClass::getGroupBy($usageGroupBy);
-
+            if ( ! $realm->groupByExists($usageGroupBy) ) {
```
jpwhite4: This code does not appear to be necessary since the $realm->getGroupByObject() will throw the UnknownGroupBy exception if the group by does not exist.

html/gui/js/modules/Usage.js
```diff
@@ -977,7 +977,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                             //
                             // Open XDMoD Default: CPU Hours: Total
                             // XDMoD Default: XD SUs Charged: Total
-                            var defaultStatistic = CCR.xdmod.features.xsede ? "total_su" : "total_cpu_hours";
+                            var defaultStatistic = CCR.xdmod.features.xsede ? 'total_su' : 'total_cpu_hours';
```
jpwhite4: These sort of changes make it very difficult to review the code for problems

classes/Rest/Controllers/DashboardControllerProvider.php
```diff
@@ -391,7 +391,14 @@ public function getStatistics(Request $request, Application $app)
         // doesn't exist" error message with something more informative for
         // Open XDMoD users.
         try {
-            $query = new \DataWarehouse\Query\Jobs\Aggregate($aggregationUnit, $startDate, $endDate, 'none', 'all');
+            $query = new \DataWarehouse\Query\AggregateQuery(
```
jpwhite4: Does this new code still throw a PDOException 42S02?

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
```diff
@@ -15,9 +15,10 @@ public function execute()
     {
         parent::execute();
 
-        // NOTE: It's possible that this should be removed after the GroupBy
-        // refactoring is merged.
         $dbh = DB::factory('datawarehouse');
+        $dbh->execute("UPDATE moddb.batch_export_requests SET realm = 'Jobs' WHERE realm = 'jobs'");
```
jpwhite4: If the realm naming convention has changed in the table does this impact and pending batch export requests?

classes/DataWarehouse/Query/Storage/GroupBy.php
```diff
@@ -91,30 +91,6 @@ public function getInfo()
         return $this->info;
```
jpwhite4: Is this file supposed to be here? I thought all of this stuff was gone?

Same for:
classes/DataWarehouse/Query/Cloud/GroupBy.php
classes/DataWarehouse/Query/Storage/GroupBy.php
classes/DataWarehouse/Query/Jobs/GroupBy.php

classes/DataWarehouse/Query/GroupBy.php
classes/DataWarehouse/Query/Statistic.php

etl/js/lib/etl_profile.js
```diff
@@ -567,21 +516,52 @@ var extractandsubst = function(column, item) {
...
-    aggCode = aggCode.replace(/_INFO_/g, column.comments);
-    aggCode = aggCode.replace(/_LABEL_/g, column.label || itemAlias);
-    aggCode = aggCode.replace(/_CATEGORY_/g, column.category || 'unknown');
-    aggCode = aggCode.replace(/\:field_name/g, column.name);
+    var label = column.label;
+    var description = column.comments;
     for( var tagidx in column.dynamictags ) {
-        aggCode = aggCode.replace( new RegExp(":label_"+tagidx, "g"), column.dynamictags[tagidx] );
-        aggCode = aggCode.replace(new RegExp(":Label_" + tagidx, "g"), wordToUpper(column.dynamictags[tagidx]));
+        label = label.replace(':label_' + tagidx, column.dynamictags[tagidx]);
```
jpwhite4: The new code replaces only the first occurrence of the string. The old code replace _all_ occurrences.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -978,20 +994,27 @@ public function getQuickFilters(Request $request, Application $app)
...
-                 ? array(
-                     'success' => $success,
-                     'results' => array(
-                         'name' => $dimensionName
-                     ))
-                 : array(
-                         'success' => false,
-                         'message' => "Unable to find a name for dimension: $dimensionId"
-                 );
+        $realm = $request->get('realm');
```
jpwhite4: We've taked about this before , but I can't remember which code was calling this function. Also is there an integration test for this endpoint?

jpwhite4: Also there are other places in the code that still call MetricExplorer::getDimensionName. do these need to be updated too? If not why not.

jpwhite4: In fact I'm not sure this code should change at all in this pull request.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1013,21 +1036,34 @@ public function getDimensionName(Request $request, Application $app, $dimensionI
...
-                     'success' => $success,
-                     'results' => array(
-                         'name' => $valueName
-                     )
-                 )
-                 : array(
-                     'success' => $success,
-                     'message' => "Unable to find a name for dimesion: $dimensionId | value: $valueId"
-                 );
+        $realmName = $request->get('realm');
```
jpwhite4: Same comments for the getDimensionValue apply here too.

classes/DataWarehouse/Access/Usage.php
smgallo: I agree that they should be consolidated and optimized, but not in this PR. The goal here was to move the functionality to the new classes/infrastructure and take another pass through to make other adjustments.

classes/Rest/Controllers/DashboardControllerProvider.php
plessbd: yes, not that it matters...
as this line https://github.com/ubccr/xdmod/blob/xdmod9.0/html/controllers/ui_data/summary3.php#L54 is the one actually throwing the exception...

tests/integration/lib/Controllers/MetricExplorerTest.php
```diff
@@ -60,7 +60,7 @@ public function testGetDwDescripterNoAuth()
     public function testGetDimensionFilters()
     {
         //TODO: Needs further integration for other realms
-        if (!in_array("jobs", self::$XDMOD_REALMS)) {
+        if (!in_array("Jobs", self::$XDMOD_REALMS)) {
```
jpwhite4: These changes are not necessary any more and I think that is why your shippable tests are failing

tests/artifacts/xdmod/integration/rest/warehouse-export/input/delete-request.json
```diff
@@ -2,7 +2,7 @@
     "Normal user": [
         "usr",
         {
-            "realm": "jobs",
+            "realm": "Jobs",
```
jpwhite4: I really think that we risk causing problems for ourselves by changing the realm names in the external interfaces. Will existing data export requests still work after the change?

Why are the names changing?

plessbd: @smgallo might be able to say why they got init capped  I do not remember why that change happened

classes/Rest/Controllers/WarehouseControllerProvider.php
plessbd: it had to do with the job viewer call to get the metrics on a realm. where it wasnt actually specifying the realm in the call (https://github.com/ubccr/xdmod/pull/1192/commits/ac90dcd37ce07d182f20a85b20ad84cb76c8a2c1).  i will look at removing this

tests/artifacts/xdmod/integration/rest/warehouse-export/input/delete-request.json
smgallo: Trying to remember. The internal realm name (the one we use to refer to them internally) is lower case while the uppercase is the display name. The ACLs did some normalization and lowercased/uppercased (?) the value which caused an issue with programmatic manipulation and I corrected that in the feature branch. This may be what is affecting the tests here.

classes/Rest/Controllers/DashboardControllerProvider.php
plessbd: fixed it in #1258

tests/artifacts/xdmod/integration/rest/warehouse-export/input/delete-request.json
plessbd: it also looks like we at some point made it uppercase as an identifier
https://github.com/ubccr/xdmod/blob/2db78ac0f980f2834f8e6a2e39616db6fce54fab/html/controllers/ui_data/summary3.php#L51

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
plessbd: i do not believe so.

tests/integration/lib/Controllers/UsageExplorerTest.php
```diff
@@ -618,7 +618,7 @@ public function provideFilterIdLookup()
         //TODO: Needs further integration for storage realm
         $realmData = array();
 
-        if (in_array("cloud", self::getRealms())) {
+        if (in_array("Cloud", self::getRealms())) {
```
jpwhite4: This should not change.

tests/integration/lib/Controllers/UserAdminTest.php
```diff
@@ -283,7 +283,7 @@ function ($filter) use ($personId) {
     public function provideTestUsersQuickFilters()
     {
         # Handle special case where there's no PI data (Cloud Realm enabled only)
-        if (self::getRealms() == array("cloud")) {
+        if (self::getRealms() == array("Cloud")) {
```
jpwhite4: This should not change

tests/integration/lib/Controllers/UserAdminTest.php
```diff
@@ -342,7 +342,7 @@ public function testGetMenus(array $user)
         # Check expected file
         $expected = array();
         foreach(self::$XDMOD_REALMS as $realm) {
-            $expectedOutputFile = $this->getTestFiles()->getFile('user_admin', $output, "output/$realm");
+            $expectedOutputFile = $this->getTestFiles()->getFile('user_admin', $output, "output/" . strtolower($realm));
```
jpwhite4: no change here.

tests/integration/lib/Controllers/UserInterfaceTest.php
```diff
@@ -71,7 +71,7 @@ public function testGetMenus(array $options)
         # Check expected file
         $expected = array();
         foreach(self::$XDMOD_REALMS as $realm) {
-            $expectedOutputFile = $this->getTestFiles()->getFile('user_interface', $expectedOutputFileName, "output/$realm");
+            $expectedOutputFile = $this->getTestFiles()->getFile('user_interface', $expectedOutputFileName, "output/" . strtolower($realm));
```
jpwhite4: change === no thanks.

tests/artifacts/xdmod/rest/output/get_statistics-usr.json
```diff
@@ -131,16 +131,16 @@
     ],
     "data": [
         {
-            "id": [
+            "none_id": [
```
jpwhite4: Presumably the change in these expected results is due to an external API change. What are the consequences of this api change? I really strongly think that the best way to reduce the risk of breakage in a refactor task is to keep the external API constant. If we want to change the rest API then we should up the version number and deprecate the old version in a seperate pull request.

configuration/roles.d/cloud.json
```diff
@@ -1,7 +1,8 @@
 {
     "+roles": {
         "+default": {
-            "+query_descripters": [{
```
jpwhite4: Why is this file changed?

tests/artifacts/xdmod/rest/output/get_statistics-usr.json
plessbd: @smgallo might be able to provide more information but i believe this is specifically because of (more in his notes)

### $multi_group in applyTo() (Revisit after initial refactor)

Appears to be used by the following methods in `GroupBy`. These are called when creating `TableField` objects in `GroupBy::applyTo()` and appear to determine if the name of the group by is included in the field alias. Should the group by name **always** be included? This would mean that we could eliminate this parameter.

Used by `TableField::__toString()` and in
- `Query::getDimensionValuesQuery()`: `$groups_str = implode(', ', $this->getGroups())`
- `Query::getQueryString()`: `$data_query .= " group by \n".implode(",\n",$this->getGroups())`

`Query::setSelectFields()` returns the qualified name of all the query fields.
Does the 1st group by need to have multi = false? Query.php has `id`, `name`, `short_name` hard coded so maybe every query class invocation must specify a group by so `setGroupBy()` is called which has `$multi = false`.

smgallo: If I recall correctly, the new infrastructure always qualifies the attributes with the group by name (here group by none).

PR #1193, opened 2019-12-17
enable cors

plessbd:
Enable the ability for an admin to enable cors allowed domains

DO NOT MERGE this is for a specific release and might have its base changed.

### DIFF
--- a/classes/Rest/XdmodApplicationFactory.php
+++ b/classes/Rest/XdmodApplicationFactory.php
@@ -115,7 +115,6 @@ class XdmodApplicationFactory
         // SETUP: the authentication Middleware to be run before the route is.
         $app->before("\Rest\Controllers\BaseControllerProvider::authenticate", Application::EARLY_EVENT);
 
-
         // SETUP: an after middleware that detects the query debug mode and, if true, retrieves
         //        and returns the collected sql queries / params.
         $app->after(function (Request $request, Response $response, Application $app) {
@@ -202,12 +201,17 @@ class XdmodApplicationFactory
                     $origin = $_SERVER['HTTP_ORIGIN'];
                     if (in_array($origin, $allowedCorsDomains)) {
                         // if these headers change we will need to update the `after` above
-                        http_response_code(204);
-                        header('Access-Control-Allow-Origin: '. $origin);
-                        header('Access-Control-Allow-Headers: x-requested-with');
-                        header('Access-Control-Allow-Credentials: true');
-                        header('Vary: Origin');
-                        exit();
+                        return new Response(
+                            '',
+                            204, /* ignored use header `X-Status-Code` */
+                            [
+                                'X-Status-Code' => 204,
+                                'Vary' => 'Origin',
+                                'Access-Control-Allow-Origin' => $origin,
+                                'Access-Control-Allow-Headers' => 'x-requested-with',
+                                'Access-Control-Allow--Credentials' => 'true'
+                            ]
+                        );
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


tests/integration/lib/Rest/WarehouseExportControllerTest.php
```diff
@@ -346,7 +346,7 @@ public function testDeleteRequestErrors()
         // Non-integer ID.
         list($content, $info, $headers) = self::$helpers['usr']->delete('rest/warehouse/export/request/abc');
         $this->assertRegExp('#\bapplication/json\b#', $headers['Content-Type'], 'Content type header');
-        $this->assertEquals(404, $info['http_code'], 'HTTP response code');
+        $this->assertEquals(4, floor($info['http_code'] / 100), 'HTTP response code');
```
chakrabortyr: Why did this change?

jtpalmer: The changes by @plessbd replace all 404 responses with 405.

plessbd:
> Adding a route that matches _anything_ may have unpredictable consequences. The whole team should discuss this before merging.

classes/Rest/XdmodApplicationFactory.php
```diff
@@ -115,9 +115,30 @@ public static function getInstance()
         // SETUP: the authentication Middleware to be run before the route is.
         $app->before("\Rest\Controllers\BaseControllerProvider::authenticate", Application::EARLY_EVENT);
 
+
```
jtpalmer: Please remove unrelated white space change.

classes/Rest/XdmodApplicationFactory.php
```diff
@@ -170,12 +191,32 @@ public static function getInstance()
...
+                    $allowedCorsDomains = explode(',', $corsDomains);
+                    $origin = $_SERVER['HTTP_ORIGIN'];
+                    if (in_array($origin, $allowedCorsDomains)) {
+                        // if these headers change we will need to update the `after` above
+                        http_response_code(204);
+                        header('Access-Control-Allow-Origin: '. $origin);
+                        header('Access-Control-Allow-Headers: x-requested-with');
+                        header('Access-Control-Allow-Credentials: true');
+                        header('Vary: Origin');
+                        exit();
```
jtpalmer: Is there a reason that the `Response` object isn't used here?

plessbd: specifically because I tried to use `Response('',204...` but it didn't work.  So after your prompting I found this little gem


https://silex.symfony.com/doc/1.3/usage.html
> As Silex ensures that the Response status code is set to the most appropriate one depending on the exception, setting the status on the response won't work. If you want to overwrite the status code, set the X-Status-Code header

classes/Rest/XdmodApplicationFactory.php
```diff
@@ -170,12 +190,37 @@ public static function getInstance()
...
+                        // if these headers change we will need to update the `after` above
+                        return new Response(
+                            '',
+                            204, /* ignored use header `X-Status-Code` */
+                            [
+                                'X-Status-Code' => 204,
+                                'Vary' => 'Origin',
+                                'Access-Control-Allow-Origin' => $origin,
+                                'Access-Control-Allow-Headers' => 'x-requested-with',
+                                'Access-Control-Allow--Credentials' => 'true'
```
jtpalmer: Is this a typo? (Different from the one above)
```suggestion
                                'Access-Control-Allow-Credentials' => 'true'
```

classes/Rest/XdmodApplicationFactory.php
```diff
@@ -170,12 +190,37 @@ public static function getInstance()
...
-                );
+            if($code == 405 && strtoupper($_SERVER['REQUEST_METHOD']) === 'OPTIONS'){
+                try {
+                    $corsDomains = \xd_utilities\getConfiguration('cors', 'domains');
+                } catch (\Exception $cors) {
+                    $corsDomains = null;
+                }
+                if (!empty($corsDomains)){
+                    $allowedCorsDomains = explode(',', $corsDomains);
+                    $origin = $_SERVER['HTTP_ORIGIN'];
```
jpwhite4: verify HTTP_ORIGIN is set before dereferencing

plessbd:
Replaced by #1206

PR #1194, opened 2019-12-19
Don't bother trying to set headers if not possible.

jpwhite4:
We often get error log reports with messages like the following:
```
PHP Warning:  Cannot modify header information - headers already sent by (output started at /opt/xdmod/share/html/index.php:180) in /opt/xdmod/share/configuration/linker.php on line 179
```

These are of course completely useless and are caused by the error
handler trying to set the headers and failing. The issue is that the
orignal message that the error handler is supposed to output does not
make it to the logs.

This change ensures that the exception handler only tries to set headers
if they have not already been sent and updates the documentation to
encourage folk to look in the xdmod exceptions log.

I manually tested this by adding a `throw new Exception` in index.php and loading the page.

### DIFF
--- a/configuration/linker.php
+++ b/configuration/linker.php
@@ -173,7 +173,7 @@ function global_uncaught_exception_handler($exception)
     $exceptionOutput = handle_uncaught_exception($exception);
 
    // If running in a server context...
-    if ($exceptionOutput['isServerContext']) {
+    if ($exceptionOutput['isServerContext'] && !headers_sent()) {
         // Set the exception's headers (if any).
         foreach ($exceptionOutput['headers'] as $headerKey => $headerValue) {
             header("$headerKey: $headerValue");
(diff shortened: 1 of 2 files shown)


jtpalmer:
In the case where the headers have been sent is it worthwhile to store the error somewhere and then display it to the user (possibly by outputting it directly into JavaScript in index.php) instead of outputting it directly?

jpwhite4:
I was think about doing something like that, but I realized that this error handler can be called from any endpoint and some output could have already been sent (with any content type). An html/json/png fragment  may already have been sent and there is not going to be an easy way to correctly close the fragment and inject a syntactically valid error message.

jpwhite4:
It would be interesting to look at refactoring index.php to split the static and dynamic parts apart. It would then be pretty easy to handle any catastrophic bootstrap errors and display a sensible message.

PR #1198, opened 2019-12-23
allow iframe login

plessbd:
detect if login is within an iframe and do no redirect to the portal, instead send a postMessage to allow the parent frame to do what it wants

something like this should work

```javascript
window.addEventListener("message", receiveMessage, false);

function receiveMessage(event) {
  if (event.origin !== "{FQDN OF XDMOD INSTANCE WITH PROTOCOL AND PORT IF NEEDED}"){
    console.log('Received message from untrusted origin, discarding');
    return;
  }
  if(event.data.application == 'xdmod'){
    if(event.data.message == 'loginComplete'){
      console.log('XDMoD has logged in successfully');
    }
	if(event.data.message == 'error'){
      console.log('ERROR: ' + event.data.info);
    }
  }
}
```
DO NOT MERGE RELATED TO #1193

### DIFF
--- a/html/gui/general/login.php
+++ b/html/gui/general/login.php
@@ -51,9 +51,18 @@ try {
   </style>
   <script type="text/javascript">
     function loadPortal() {
-      setTimeout(function(){
-        parent.location.href = '/index.php' + document.location.hash;
-      }, 1500);
+        try {
+            var href = window.top.location.href;
+            setTimeout(function(){
+                parent.location.href = '/index.php' + document.location.hash;
+            }, 1500);
+        }
+        catch(exception){
+            window.top.postMessage({
+                application:'xdmod',
+                action: 'loginComplete'
+            },'*');
+        }
     }
 
     function contactAdmin() {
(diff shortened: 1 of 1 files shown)


html/gui/general/login.php
```diff
@@ -73,12 +73,24 @@ function contactAdmin() {
...
         </p>
+        <script>
+           try {
+               var href = window.top.location.href;
+           }
+           catch(exception){
+               window.top.postMessage({
+                   application:'xdmod',
+                   action: 'error',
+                   info: '<?php echo $message; ?>'
```
jpwhite4: json_encode($message) to escape any nasties.

html/gui/general/login.php
```diff
@@ -73,12 +73,24 @@ function contactAdmin() {
...
+               },'*');
+           }
+        </script>
       </body>
       </html>
     <?php
     } else {
     ?>
-<body onload="loadPortal()">
+<body>
```
jpwhite4: Suggest keeping the script start trigger with the onload event rather than inline. It definitely worked before and I don't think we should change it.

PR #1199, opened 2020-01-14
Ingest cloud resource specifications

eiffel777:
This PR adds functionality to track the vcpu and memory specifications of cloud resources over time. A new ingestor class has been created called StateReconstructorTransformIngestor. This class simulates a Finite State Machine to construct a single row with a start and end time. Given a list of records with a time when the event the record describes takes place, records with a start and end times for unique column values based on criteria specified in the action definition file are created.

In the case of the cloud resource specifications, specification information and the date that information was recorded are used to create rows with start end end times for each unique combination of resource_id, hostname, vcpus and memory.

There are other instances where this ingestor class will be useful such as construcing start and end times for instance type configurations or the start and end times for XSEDE allocations. In each case the columns used to construct the new rows may be unique to each use case. In order to allow for ease of reuse of this class without having to make new ingestor classes each time a json object in added to the action definition file to specify the column names used in the ingestor class.

Below is the json object that should be added to the action definition file. Each field should list at least one column from the source query.

```json
state_reconstruction_fields: {
     "end_time": "",
     "new_row": [],
     "update_row": [],
     "reset_row": [],
     "order_by": []
}
```

Component tests have been added for the new ingestor class.

## Tests performed
Added new component test and tested manually in docker

## Types of changes

## Checklist:


classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,273 @@
...
+     * @see ETL\Ingestor\pdoIngestor::initialize()
+     */
+    public function initialize(EtlOverseerOptions $etlOverseerOptions = null){
+
+        if ( $this->isInitialized() ) {
+            return;
+        }
+
+        $messages = null;
+        $this->initialized = false;
```
jpwhite4: Please explain why initialised is set here

classes/ETL/Ingestor/StateReconstructorTransformIngestor.php
```diff
@@ -0,0 +1,273 @@
...
+
+        if ( ! \xd_utilities\verify_object_property_types($this->parsedDefinitionFile->state_reconstruction_fields, $stateReconstructorRequiredKeys, $messages) ) {
+            $this->logAndThrowException(
+                sprintf("Error verifying definition file 'state_reconstruction_fields' section: %s", implode(', ', $messages))
+            );
+        }
+
+        $orderby_fields_array = [];
+        foreach( $this->parsedDefinitionFile->state_reconstruction_fields->order_by as $orderby ) {
+            $orderby_fields_array[] = explode(' ', trim($orderby))[0];
```
jpwhite4: please explain why it is ignoring anything after the first space.

PR #1209, opened 2020-01-16
Documentation Updates

plessbd:
documentation updates for CORS and SSO embedding

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -207,6 +207,7 @@ Primary configuration file. Contains:
 - PhantomJS path
 - Header logo (see [Logo Image Guide](logo-image.html) for details)
 - Database configuration
+- Integration settings (see [Inegrations](integrations.html) for details)
 
 ### datawarehouse.json
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -207,7 +207,7 @@ Primary configuration file. Contains:
 - PhantomJS path
 - Header logo (see [Logo Image Guide](logo-image.html) for details)
 - Database configuration
-- Integration settings (see [Inegrations](integrations.html) for details)
+- Integration settings (see [Integrations](integrations.html) for details)
 
 ### datawarehouse.json
 
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -8,10 +8,10 @@ The Open XDMoD application settings to enabled this are described below.
 ## Open XDMoD Settings
 
 ### Cross-Origin Resource Sharing (CORS)
-To allow CORS set a list of domains that are allowed to communicate with Open XDMoD via
+To allow CORS  a list of domains that are allowed to communicate with Open XDMoD is configured in
 the `domains` setting in the `cors` section of `portal_settings.ini`.
 
-This is a comma separated list of both the scheme and authority
+This setting is a comma separated list where each item contains the scheme, host, and optional port.
 ```
 [cors]
 domains=https://integratedapp.example.tld,https://dev-integratedapp.example.tld:8080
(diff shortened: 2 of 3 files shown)


docs/configuration.md
```diff
@@ -207,6 +207,7 @@ Primary configuration file. Contains:
 - PhantomJS path
 - Header logo (see [Logo Image Guide](logo-image.html) for details)
 - Database configuration
+- Integration settings (see [Inegrations](integrations.html) for details)
```
jpwhite4:
```suggestion
- Integration settings (see [Integrations](integrations.html) for details)
```

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
+# Integrations
+
+XDMoD can be integrated with other Applications such as [Open OnDemand](https://openondemand.org/)
```
jpwhite4: To enable integration, the XDMoD instance must be configured to give explicit permission to the browser. The XDMoD application settings to enabled this are described below.

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Settings
+
+### Cross-Origin Resource Sharing (CORS)
+To allow CORS set a list of domains that are allowed to communicate with XDMoD via
+the `domains` setting in the `cors` section of `portal_settings.ini`.
+
```
jpwhite4: need to tell them that the list is comma separated. Also need to mention that the lookup is an exact string match so you can't specify just the domain. you need the full uri including the https and prot number (what about the slash?) (and give an example).

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+**NOTE: This setting can open up XDMoD to security risks if used wrong.
```
jpwhite4: if not used rightly.

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+Only enable it if you know what you are doing**
+
+## Single Sign-on (SSO) Embedding
+XDMoD allows for single sign on embedding.  When [SSO](simpleSAMLphp.html) is configured with the same IDP
```
jtpalmer:
```suggestion
Open XDMoD allows for single sign on embedding.  When [SSO](simpleSAMLphp.html) is configured with the same IDP
```

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+XDMoD can be embedded in an iframe at the login endpoint and automatically logged in.
```
jtpalmer:
```suggestion
Open XDMoD can be embedded in an iframe at the login endpoint and users will be automatically logged in.
```

docs/integrations.md
jpwhite4: XDMoD allows for single sign on embedding so that end users do not have to explicitly sign in to XDMoD after they have already signed in to the other integrated application.

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+
+When XDMoD detects login within an iframe it will send a `postMessage` to `window.top`
```
jtpalmer:
```suggestion
When Open XDMoD detects a login within an iframe it will send a `postMessage` to `window.top`
```

docs/integrations.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Settings
+
+### Cross-Origin Resource Sharing (CORS)
+To allow CORS set a list of domains that are allowed to communicate with XDMoD via
+the `domains` setting in the `cors` section of `portal_settings.ini`.
+
+**NOTE: This setting can open up XDMoD to security risks if used wrong.
+Only enable it if you know what you are doing**
+
```
jpwhite4: This is part of the developer guide setting not something a sys admin would have to do.

PR #1214, opened 2020-01-17
Hmm

jpwhite4:
Just wondering why this ripTransform function exists since it does not appear to serve any practical function except obfuscate the code.

### DIFF
--- a/classes/XDReportManager.php
+++ b/classes/XDReportManager.php
@@ -1088,7 +1088,7 @@ class XDReportManager
     }
 
     private function generateCachedFilename(
-        $insertion_rank,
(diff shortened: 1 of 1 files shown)


classes/XDReportManager.php
```diff
@@ -1098,8 +1098,6 @@ private function generateCachedFilename(
                 ? $insertion_rank['did']
                 : '';
 
-            $this->ripTransform($insertion_rank, 'did');
-
             if (
                 is_array($insertion_rank)
```
jtpalmer: Shouldn't need `is_array` anymore now that you've added the type hint.

PR #1215, opened 2020-01-22
Focus on the Username field when the Login Button is clicked

plessbd:
When clicking the login button focus on the username field when the form activates.

Verified that this works with and without SSO enabled.

This should have been part of #1115


plessbd:
Seems like this messes up the timing on the login, going to have to look at that and fix it...

PR #1218, opened 2020-01-22
Filter Var

jpwhite4:
## Types of changes

## Checklist:

### DIFF
--- a/configuration/constants.php
+++ b/configuration/constants.php
@@ -179,6 +179,21 @@ define('RESTRICTION_INSTITUTION', '/^(-)?[0-9]+$/');
 // Associated with 'id' in modw.daterange
 define('RESTRICTION_DATE_RANGE_ID', '/^[0-9]{1,2}$/');
 
+define('RESTRICTION_REPORT_ID', '/^[0-9]+-[0-9\.]+$/');
+define('RESTRICTION_REPORT_DATE', '/^[0-9]{4}(-[0-9]{2}){2}$/');
+define('RESTRICTION_REPORT_FORMATS', '/^doc|pdf$/');
+define('RESTRICTION_REPORT_FONT', '/^Arial$/');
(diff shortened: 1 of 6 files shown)


configuration/constants.php
```diff
@@ -179,6 +179,21 @@
...
+define('RESTRICTION_REPORT_DELIVERY', '/^E-Mail$/');
+
+define('RESTRICTION_REPORT_CHART_TYPE', '/^chart_pool|volatile|report|cached$/');
+define('RESTRICTION_REPORT_CHART_REF', '/^[0-9]+(-[0-9]+)?;[0-9]+$/');
+define('RESTRICTION_REPORT_CHART_DID', '/^_d[0-9]+$/');
+
+define('RESTRICTION_CHART_CACHEREF', '/^([0-9]{4}(-[0-9]{2}){2};){2}(?(?=xd_report_volatile_)xd_report_volatile_[0-9]+;[0-9]+(_d[0-9]+)?|[0-9]+-[0-9\.]+;[0-9]+)$/');
+
+define('RESTRICTION_REPORT_TMPDIR', '/^[0-9]+-[0-9\.]+-[a-zA-Z0-9\.]+$/');
+
```
jtpalmer: I've been trying to discourage the use of global constants based on the argument that global state is bad.  If these are directly related to reports then I think it would be better if they are class constants of some report class (e.g. `XDReportManager`).

jpwhite4: They are  related to the encoding used in the report_builder controller rather than the XDReportManager itself. These controller bits are individual php files (not even functions or classes). 

How about I create a \Datawarehouse\Access\ReportGenerator class and stick them in there?

jtpalmer: That would be fine.  It doesn't look like there is too much overlap between the validation in the various endpoints, but it might make sense to try and put that there as well if there was.

PR #1219, opened 2020-01-23
Add input validation to report generator controller endpoints.

jpwhite4:
## Description

This pull request adds input validation filters to the user-supplied
untrusted data provided to the report generator controller endpoints.

The modifications do not change any of the data serialization or api of
the endpoints.

The user-supplied report text (title, header, footer etc.) is modified
to filter out data that does not have a valid character encoding. This
mitigates a bug in the report generator if you tried to create a report
with a filename containing an invalid character such as 😊 then the UI would
show an error message:

```
There was a problem creating / updating this report.
(SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation 'like')
```

the new behaviour is that the report will save ok but any invalid
characters will be converted to '?'. A complete fix for this bug is
beyond the scope of this pull request.

## Motivation and Context

Potential security vulnerability using untrusted user data.

## Tests performed

There are several complex regular expressions used to validate the data
this is necessary due to the non-standard serialization used in the
report generator.  Unit tests have been added to confirm that the more
complex regular expressions used do pass through permissible data.

### DIFF
--- a/html/controllers/report_builder/download_report.php
+++ b/html/controllers/report_builder/download_report.php
@@ -22,7 +22,7 @@ try {
 
     $get = filter_input_array(INPUT_GET, $filters);
 
-    if (XDReportManager::isValidFormat($get['format']) == false) {
+    if (!XDReportManager::isValidFormat($get['format'])) {
         print "Invalid format specified";
         exit;
     }
--- a/html/controllers/report_builder/download_report.php
+++ b/html/controllers/report_builder/download_report.php
@@ -1,17 +1,33 @@
 <?php
 
+use \DataWarehouse\Access\ReportGenerator;
+
+$filters = array(
+    'format' => array(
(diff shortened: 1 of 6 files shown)


html/controllers/report_builder/download_report.php
```diff
@@ -1,17 +1,33 @@
...
     'report_loc',
     'format'
 ));
 
 try {
-    if (XDReportManager::isValidFormat($_GET['format']) == false) {
+
+    $get = filter_input_array(INPUT_GET, $filters);
+
+    if (XDReportManager::isValidFormat($get['format']) == false) {
```
jtpalmer:
```suggestion
    if (!XDReportManager::isValidFormat($get['format'])) {
```

PR #1223, opened 2020-01-24
Improve data warehouse export logging

jtpalmer:
## Description

Adds additional logging and improves existing logging to add more details.

## Motivation and Context

The logged information is necessary to provide details on the use of the data warehouse export features.

https://app.asana.com/0/15230011293693/1146676887787307

## Tests performed

No tests

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Export/BatchProcessor.php
+++ b/classes/DataWarehouse/Export/BatchProcessor.php
@@ -54,9 +54,9 @@ class BatchProcessor extends Loggable
         // Must set properties that are used in `setLogger` before calling the
         // parent constructor.
         $this->fileManager = new FileManager($logger);
+        $this->queryHandler = new QueryHandler($logger);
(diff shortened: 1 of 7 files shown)


classes/DataWarehouse/Export/BatchProcessor.php
```diff
@@ -16,6 +16,15 @@
 
 class BatchProcessor extends Loggable
 {
+    // Constants used in log messages.
+    const LOG_MODULE_KEY = 'module';
```
jpwhite4: lolzers the name of the constant is longer that the value!

jtpalmer: Also, "module" is probably a poor choice since we're using that to mean something else.  Let me know if you have any other suggestions.  I'm trying to group these messages since they include some that could be run by both the web server and the batch process.

PR #1224, opened 2020-01-29
return user supplied jobname in search

plessbd:
Add job_name to response on job search


plessbd:
> 
Eric Franz
added a public note , 2 days ago (Tue, Feb 4 2020 at 1:11 PM)
We have these patches in [...] and I think that might work. The job names are definitely there! I'll be able to report more on this when I build a completed jobs prototype now that this is available.

PR #1244, opened 2020-03-05
Make composer be quiet so we can read the logs better

plessbd:
make composer be quiet so logs are easier to read in shippable

### DIFF
--- a/shippable.yml
+++ b/shippable.yml
@@ -1,7 +1,10 @@
 language: none
 env:
-    - XDMOD_TEST_MODE=fresh_install
-    - XDMOD_TEST_MODE=upgrade
+    global:
+        - COMPOSER_ALLOW_SUPERUSER=1
+    matrix:
+        - XDMOD_TEST_MODE=fresh_install
+        - XDMOD_TEST_MODE=upgrade
 build:
     cache: true
     cache_dir_list:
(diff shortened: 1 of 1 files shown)


shippable.yml
```diff
@@ -15,11 +15,12 @@ build:
         pull: false
         options: "--user root -e HOME=/root --shm-size 2g"
     ci:
-        - composer install --no-progress
+        - export COMPOSER_ALLOW_SUPERUSER=1
```
jtpalmer: Why not make this a [global environment variable](http://docs.shippable.com/ci/env-vars/#global-variables)?

shippable.yml
```diff
@@ -15,11 +15,12 @@ build:
...
+        - composer install --no-progress -q
         - ~/bin/buildrpm xdmod
         - ./tests/ci/bootstrap.sh
         - ./tests/ci/validate.sh
-        - composer install --no-progress
+        - composer install --no-progress -q
```
jtpalmer: `--no-progress` is not necessary with the addition of `-q`.  Since `--no-progress` was being used are you sure there is no valuable information that we may miss in the future?

shippable.yml
plessbd: if i remember correctly the env here causes a new matrix build

shippable.yml
plessbd: 100% sure, no.  But errors will still cause it to fail, so i do not believe there will be any needed information missed here.  When doing personal docker builds this can still be turned on so I think that will provide all the information we need.

shippable.yml
plessbd: i was unable to find that page last night.

shippable.yml
plessbd: it will also still show warnings
`Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.`

PR #1249, opened 2020-03-09
Update openstack json event schema definition 

eiffel777:
Update the json schema for openstack events to allow null for the user_name field. This allows events whose user_name field has a null value to be ingested. Any event with a user_name of null will be assigned to the unknown value when aggregated.

## Tests performed
Tested in docker

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_schemas.d/cloud_openstack/event.schema.json
+++ b/configuration/etl/etl_schemas.d/cloud_openstack/event.schema.json
@@ -65,7 +65,7 @@
         "type": "string"
       },
       "user_name": {
-        "type": "string"
+        "type": ["string", "null"]
       },
       "root_gb": {
         "type": "integer"


plessbd:
Do we have any use cases in docker that this was actually hitting?
I didnt think we had any null cases in docker/

jpwhite4:
Please can you briefly list the testing performed on this change.

eiffel777:
@plessbd @jpwhite4 There weren't any cases for data that was already in the docker. I tested by spinning up a docker instance and then ingested some of the lakeeffect openstack data that did have null values in the user_name field. Before the change there were warnings such as this one.

> 2020-03-12 06:35:16 [warning] Record 8: NULL value found, but a string is required

These warning no longer occur once this change is applied.

PR #1250, opened 2020-03-10
Added regression tests for chart filters.

jpwhite4:
These tests confirm that the chart filters filter the data correctly
and have the correct chart subtitles. All filters available for each
realm are tested with a single metric and group by none.

This also includes the machinery to generate expected test results.

The test for Cloud Filter by instance type is absent. This is because
this filter does not work. Once the instance type filtering is fixed
then a test should be added.

### DIFF
--- a/tests/artifacts/xdmod/regression/chartFilterTests.json
+++ b/tests/artifacts/xdmod/regression/chartFilterTests.json
@@ -1,0 +1,480 @@
+[
+    {
+        "settings": {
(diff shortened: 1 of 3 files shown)


jpwhite4:
Note the sonar cloud analysis is a false positive. The crappy algorithm thinks that the code is duplicated because the file that changed has a symlink!

jtpalmer:
Why does the symlink need to exist?

PR #1254, opened 2020-03-12
Fix some bugs in FilterListBuilder

plessbd:
found a few bugs in the filterlistbuilder...

### DIFF
--- a/classes/DB/FilterListBuilder.php
+++ b/classes/DB/FilterListBuilder.php
@@ -36,16 +36,6 @@ class FilterListBuilder extends Loggable
     private static $rolesDimensionNames = null;
 
     /**
-     * Construct filter list builder.
-     */
-    public function __construct()
-    {
-        if ($this->logger === null) {
-            $this->logger = Log::singleton('null');
-        }
-    }
-
-    /**
      * Build filter lists for all realms' dimensions.
      */
     public function buildAllLists()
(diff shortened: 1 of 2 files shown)


jtpalmer:
Note that the logging issues were caused by #1240

classes/DB/FilterListBuilder.php
```diff
@@ -40,8 +40,8 @@ class FilterListBuilder extends Loggable
      */
     public function __construct()
     {
-        if ($this->_logger === null) {
-            $this->_logger = Log::singleton('null');
+        if ($this->logger === null) {
+            $this->logger = Log::singleton('null');
```
jtpalmer: Now that this class extends `\CCR\Loggable` the constructor should either be refactored to call the parent constructor or removed completely.

classes/DB/FilterListBuilder.php
```diff
@@ -106,28 +106,23 @@ public function buildDimensionLists(Query $realmQuery, GroupBy $groupBy)
...
+        try {
+            $dimensionProperties = $this->getDimensionDatabaseProperties($realmQuery, $groupBy);
+        } catch (TableNotFoundException $e) {
+            $this->logger->notice("Not creating $targetSchema.$mainTableName list table; {$e->getTable()} table not found");
+            return;
+        }
+        $dimensionColumnType = $dimensionProperties['type'];
+        try {
+            $db->beginTransaction();
+            $db->execute("DROP TABLE IF EXISTS `{$targetSchema}`.`{$mainTableName}`;");
```
jtpalmer: You cannot drop and create tables inside a transaction.

classes/DB/FilterListBuilder.php
```diff
@@ -106,28 +106,23 @@ public function buildDimensionLists(Query $realmQuery, GroupBy $groupBy)
 
         $db = DB::factory('datawarehouse');
         $targetSchema = FilterListHelper::getSchemaName();
-        $mainTableExistsResults = $db->query("SHOW TABLES FROM {$targetSchema} LIKE '$mainTableName'");
-        if (empty($mainTableExistsResults)) {
```
jtpalmer: Why must the table always be dropped and recreated instead of checking to see if it exists?

classes/DB/FilterListBuilder.php
plessbd: was old code, removed transations

classes/DB/FilterListBuilder.php
plessbd: If the datatype of the filter that gets added (as is the case in #1253) changes a truncate is not sufficient

jtpalmer:
Since the tables are now always being recreated it might be worth considering combining the `CREATE TABLE` and the `INSERT` statements  into a single `CREATE TABLE ... SELECT` statement.  Not sure if it would be worth the effort, but it could potentially simplify things.

plessbd:
I will look at doing that as part of another pull request

PR #1261, opened 2020-03-14
make sure tests log

plessbd:
Not all of the tests are logging in shippable.  Lets make sure they do.


jpwhite4:
Would be nice to move all of the common test functions to the common file. For example, the code that computes the path to phpunit.

plessbd:
that is why I started it, I am fine with more updates at a later point

PR #1263, opened 2020-03-17
Change account, instance and instance_type tables to have single unique auto increment column

eiffel777:
This PR fixes issues with the global and local filters for the Project and Configuration group by's in the Cloud realm. The instance, instance_type and account tables are set up to have a multi column increment key based on the resource_id and an auto increment id field. For the filters to work correctly there needs to be a single unique id column, which this PR adds. The multi column increment key is changed to a single column auto increment field and the group bys are updated to use the display field to filter by instead of the id field. 

There are additions to the migration etl pipeline to account for these changes. In order to change the existing multi column key to a single column key the auto increment is removed from the id column, then the multi key index is dropped and the id column is removed. When ingestion is next run at the end of the migration pipeline the id field is created again as a single column auto increment field with the appropriate auto incremented id number. Since the instance_id is used in the event table a sql statement is run to update any rows in the event table to the new instance_id.

Changes are also made to the test bootstrap.sh file to test migrating an xdmod installation that has multiple cloud resource before upgrading. The existing bootstrap file adds data for a second cloud resource after the upgrade has been completed. This PR needs to test explicitly upgrading an xdmod instance that has data for two cloud resources before upgrading xdmod. As such, the bootstrap.sh file is updated to add a cloud resource and ingest test data before the yum command is run to update xdmod. Test data is ingested again after the upgrade to allow for testing of new functionality in Open Xdmod 9.0 such as the addition of the Provider and Domain group bys.

## Motivation and Context
This makes sure that global and local filters work as expected.

## Tests performed
Tested in docker. Created two cloud resources, added data for both and then made sure the relevant tables had a single auto increment id column. Tested both a fresh install and upgrade.

## Types of changes

## Checklist:


configuration/datawarehouse.d/ref/Cloud-group-bys.json
```diff
@@ -19,10 +12,10 @@
         "order": 6,
         "attribute_values_query": {
             "records": {
-                "id": "instance_type_id",
-                "short_name": "instance_type",
+                "id": "display",
```
ryanrath: Could I snag a brief explanation on this update from `instance_type_id` => `display`?

configuration/datawarehouse.d/ref/Cloud-group-bys.json
```diff
@@ -139,10 +133,10 @@
                 "display"
             ],
             "records": {
-                "id": "account_id",
+                "id": "display",
```
ryanrath: same as `instance_type_id`?

configuration/datawarehouse.d/ref/Cloud-group-bys.json
ryanrath: nvm, that whole reading thing is important I hear.

ryanrath:
So I'm not sure if this is something that you've run into already / is even a concern but I was taking a look at `projects` data and it looks like when filtering with this group by 312 records are returned  but there's actually 313 records in `modw_cloud.accounts` table? It looks like it's specifically because there are two `accounts` with the same `display` but different `account_id` values. 
`account_id=16, display="jomlowe"` and `account_id=243, display=jomlowe`.

jpwhite4:
Please can you add these two fixed groupbys as new cases for the testChartFilters test in  tests/regression/lib/Controllers/MetricExplorerChartsTest.php.

The comments in the test provider explain how to generate the test artifacts.

tests/ci/bootstrap.sh
```diff
@@ -93,6 +93,19 @@ fi
 
 if [ "$XDMOD_TEST_MODE" = "upgrade" ];
 then
+    if [[ "$XDMOD_REALMS" = *"cloud"* ]]; then
```
jpwhite4: I don't think this is a good idea since it breaks the testing model for upgrades. I.e. we take a working xdmod instance and run the upgrades on it without modifying it. Also won't this break the existing regression tests?

eiffel777: I think the other option would be to create a new docker version that has data for two cloud resources. What I need is for data for two cloud resources to be in the docker before the upgrade happens so the migration can be properly tested. I asked @plessbd about this last week and doing it the way the PR does it now or making a new docker with data for two cloud resources were the two options that came up. 

I can work on updating the docker if that is the more preferable method.

jpwhite4: We obviously should test that the upgrade path works with two resources - but this does not necessarily have to be run in the CI infrastructure. This upgrade step will presumably only be needed for 8.5 to 9.0 upgrades anyway and will not happen for a 9.0 -> 9.x upgrade?

plessbd: correct.  I thought we have done things like this in the past and just made sure to remove them from CI when we did the new branch

jpwhite4:
Need to update the upgrade documentation too. It looks like if they accidentally run the ingestion before running the xdmod-upgrade then the etlv2 code in the ingestor will try to modify the tables with disasterous consquences.

configuration/etl/etl_sql.d/migrations/cloud-migration-8-5-1_9-0-0.sql
```diff
@@ -0,0 +1,31 @@
+-- In order to chnage from a multi column to single column unique auto increment id field
+-- we change in id field on the table to not be auto increment, then drop the auto increment
+-- key and finally drop the id field. Once ingestion is run again at the end of the migration
+-- pipeline the id field is remade as a unique field with auto incremented id's. The aggregation
```
jpwhite4: suggest you add a lock tables statement at the top for all tables that are going to get modified. This makes it (slightly) safer. Also I would do all of the drop and renames at the very end so that if something goes wrong it will be able to be recovered with no data loss.

jpwhite4:
Still need documentation updates. Can do this as a separate pull request in the interest of keeping things moving along....

eiffel777:
@jpwhite4 Yeah, I think it would be better to do as a separate pull request. I actually have a few things I need to add to the upgrade documentation because of the various changes in 9.0. I can put all my documentation changes so far into one pull request including the information for this PR.

PR #1264, opened 2020-03-17
Improve performance of the timeseries queries.

jpwhite4:
Description to follow....


classes/DataWarehouse/Data/TimeseriesDataset.php
```diff
@@ -0,0 +1,377 @@
...
+use \DataWarehouse\Query\Model\Schema;
+use \DataWarehouse\Query\Model\WhereCondition;
+use \DataWarehouse\Query\Model\OrderBy;
+
+class TimeseriesDataset
+{
+    // The summarized dataseries is assigned a dedicated group id value. This
+    // is used by the frontend code in the Usage Tab to not generate a
+    // drilldown tooltip @refer html/gui/js/DrillDownMenu.js The visualization
+    // class also checks this to set the remainder flag on the dataset for hte
```
plessbd:
```suggestion
    // class also checks this to set the remainder flag on the dataset for the
```

classes/DataWarehouse/Data/TimeseriesDataset.php
```diff
@@ -0,0 +1,377 @@
...
+    protected $series_count = null;
+
+    public function __construct(TimeseriesQuery $query)
+    {
+        $this->query = $query;
+        $this->agg_query = $query->getAggregateQuery();
+
+    }
+
+    /**
```
plessbd: we should decide if we are going to do "real" docblocks or not...

classes/DataWarehouse/Data/TimeseriesDataset.php
```diff
@@ -0,0 +1,377 @@
...
+            } else {
+                $spaceGroup = $groupBy;
+            }
+        }
+
+        return array($timeGroup, $spaceGroup);
+    }
+
+    /**
+     * return an array of timeseries datasets. If the summarize flag is set true and there
```
plessbd: do we want this as an actual `@returns array`?

classes/DataWarehouse/Query/TimeseriesQuery.php
```diff
@@ -94,6 +94,88 @@ public function __construct(
...
+     * This call does not change the sort order of the timeseries query
+     * itself, rather it is used to change the sort order of the associated
+     * aggregrate query that is returned by the getAggregateQuery() function.
+     */
+    public function addOrderByAndSetSortInfo($data_description)
+    {
+        switch ($data_description->sort_type) {
+            case 'value_asc':
+                $this->sortInfo = array(
+                array(
```
plessbd: i guess since travis doesnt care I shouldnt care about the indenting here being off by a level...

jpwhite4: Tis a cut and paste from the old code.  I'll fix it up

classes/DataWarehouse/Data/TimeseriesDataset.php
jpwhite4: I suppose we can try to do it for new code. There are phpcs rules that we can enable to enforce this programatically.

PR #1269, opened 2020-03-19
Remove unnecessary redirect in the job viewer rest stack.

jpwhite4:
The job viewer rest api has an unusual code path to get the list of jobs
for a given saved search. There is a 302 redirect when the interface
gets the list of jobs. There is no 302 redirect for any of the other
equivalent queries (such as list of information for a job). All of the
required information to respond to the request is in the original request.

This pull request removes the redirect. The path that wasa being
redirected to still remains so as to not break compatibility with any
external systems (they both call the same underlying function).

This was observed while testing #1268


jpwhite4:
Note that this code path is tested in the UI tests.

PR #1277, opened 2020-03-31
added updating version in appkernels.ini upon upgrade

nsimakov:
added updating version variable in modules portal setting upon upgrade.

In appkernels version variable doesn't do much but it will be nice to have matching config version and software.

The big question is should it really be in core xdmod or in xdmod-appkernel. This would largely depends on how user is going to upgrade their system. In xdmod-upgrade, we did mentioned update supremm/appkernel tools first and then run xdmod-upgrade. Under this scenario it should work.

Since there are no other changes in appkernels.ini besides version and I don't anticipate any change in near future, just version update should be fine for a while.

### DIFF
--- a/classes/OpenXdmod/Migration/Version851To900/ConfigFilesMigration.php
+++ b/classes/OpenXdmod/Migration/Version851To900/ConfigFilesMigration.php
@@ -25,5 +25,9 @@ class ConfigFilesMigration extends AbstractConfigFilesMigration
                'roadmap_url' => $roadmapUrl
             )
         );
+
+        foreach ($this->modulePortalSettingsPaths as $moduleName => $modulePortalSettingsPath) {
+            $this->writeModulePortalSettingsFile($moduleName);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


classes/OpenXdmod/Migration/Version851To900/ConfigFilesMigration.php
```diff
@@ -25,5 +25,9 @@ public function execute()
                'roadmap_url' => $roadmapUrl
             )
         );
+
+        foreach ($this->modulePortalSettingsPaths as $moduleName => $modulePortalSettingsPath) {
+            $this->writeModulePortalSettingsFile($moduleName);
+        }
```
jtpalmer: You shouldn't need the loop.  See https://github.com/ubccr/xdmod/blob/xdmod9.0/classes/OpenXdmod/Migration/ConfigFilesMigration.php#L93

This function call already does that:

```php
$this->writeModulePortalSettingsFiles();
```

nsimakov: changed to $this->writeModulePortalSettingsFiles();

PR #1278, opened 2020-04-01
Default GPU count to zero

jtpalmer:
## Description

Change definition of columns that contain GPU counts so they are `NOT NULL` and default to `0`.

## Motivation and Context

It was previously possible to have `NULL` values which produces warning messages:

```
2020-03-31 04:22:32 [warning] SQL warnings on table 'job_tasks' generated by action xdmod.hpcdb-xdw-ingest-jobs.job-records-job-tasks (filtering codes: 1062)
2020-03-31 04:22:32 [warning] Warning	1263	Column set to default value; NULL supplied to NOT NULL column 'gpu_count' at row 2
```

## Tests performed

No new tests.

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/jobs/hpcdb/jobs.json
+++ b/configuration/etl/etl_tables.d/jobs/hpcdb/jobs.json
@@ -110,7 +110,8 @@
             {
                 "name": "gpucount",
                 "type": "int(11)",
-                "nullable": true
+                "nullable": false,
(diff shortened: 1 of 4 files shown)


jpwhite4:
Looks good to me. We should probably have some info in the docs about how it will show as 0 gpus if the information is not available/parsable from the resource manager

PR #1280, opened 2020-04-04
Surface Error instead of sitting on "Loading..."

plessbd:
When there is an error in get_menus it can get 'lost' behind a 'Loading...' this will surface the error so that an admin can fix the configuration files.

### DIFF
--- a/html/gui/js/modules/Usage.js
+++ b/html/gui/js/modules/Usage.js
@@ -660,7 +660,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
 
                             }
                             if (resp.success === false) {
-                                Ext.MessageBox.alert("Error", resp.message);
+                                Ext.MessageBox.alert('Error', resp.message);
                             }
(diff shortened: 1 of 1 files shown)


html/gui/js/modules/Usage.js
```diff
@@ -659,6 +659,9 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                                 CCR.xdmod.ui.actionLogout.defer(1000);
 
                             }
+                            if (resp.success === false) {
+                                Ext.MessageBox.alert("Error", resp.message);
```
jtpalmer, line 663: I don't even remember this being a rule, but single quotes are required (https://travis-ci.org/github/ubccr/xdmod/jobs/670845844#L1333):

```suggestion
                                Ext.MessageBox.alert('Error', resp.message);
```

plessbd: copy and paste from a few lines above failed me...

PR #1287, opened 2020-04-20
updates needed for xsede groupbys

plessbd:
The XSEDE GroupBys need to be able to use a secondary table to link to aggregates (through the account table to the grant table) this update allows that, while getting rid of the need for the attribute_table key as it is provided by the first (and usually only) join table in the attributes value query.


plessbd:
For some reason this PR got out of sync probably related to https://www.githubstatus.com/incidents/dsf2qtzh4jpz
it will be reopened as a new PR

PR #1288, opened 2020-04-21
updates needed for xsede groupbys

plessbd:
The XSEDE GroupBys need to be able to use a secondary table to link to aggregates (through the account table to the grant table) this update allows that, while getting rid of the need for the attribute_table key as it is provided by the first (and usually only) join table in the attributes value query.

### DIFF
--- a/classes/Models/Services/Acls.php
+++ b/classes/Models/Services/Acls.php
@@ -987,7 +987,7 @@ SQL;
 SELECT
     r.name as realm,
     gb.name as group_by,
-    MAX(agb.enabled) as enabled,
+    !MAX(agb.enabled) as not_enabled,
     MAX(agb.visible) as visible
 FROM group_bys gb
   JOIN realms r ON gb.realm_id = r.realm_id
--- a/classes/Models/Services/Acls.php
+++ b/classes/Models/Services/Acls.php
@@ -984,11 +984,11 @@ SQL;
         }
 
         $query = <<<SQL
-        SELECT DISTINCT
-            r.display as realm,
-            gb.name as group_by,
-            !agb.enabled as not_enabled,
-            agb.visible
+SELECT
+    r.name as realm,
+    gb.name as group_by,
+    MAX(agb.enabled) as enabled,
+    MAX(agb.visible) as visible
 FROM group_bys gb
   JOIN realms r ON gb.realm_id = r.realm_id
   JOIN acl_group_bys agb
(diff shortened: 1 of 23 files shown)


classes/Models/Services/Acls.php
```diff
@@ -1034,6 +1034,7 @@ public static function getQueryDescripters(XDUser $user, $realmName = null, $gro
       at.name = 'data'
   )
 WHERE ua.user_id = :user_id $statisticWhere
+AND agb.visible = 1
```
jpwhite4, line 1037: If you set visible to 1 then why is visible being pulled back as a query column and then case to php bool?

jpwhite4: Similarly why does line 1075 exist? ``$descripter->setShowMenu((bool)$row['visible']);`

classes/Models/Services/Acls.php
```diff
@@ -1056,6 +1057,7 @@ public static function getQueryDescripters(XDUser $user, $realmName = null, $gro
             $params[':statistic_name'] = $statisticName;
         }
 
+        $query .= "\nORDER BY 1,2,3,4 DESC";
```
jpwhite4, line 1060: What is the impact of adding an order by here? What bug is fixed?

plessbd: It helped identify that the order was what was causing the groupbys to either show up or not show up depending on if the user was a federated user or a local user.
This might not be required after the latest change.  But I think it should be coming back in the same order all the time anyway

classes/Models/Services/Acls.php
```diff
@@ -984,11 +984,11 @@ public static function getQueryDescripters(XDUser $user, $realmName = null, $gro
...
         $query = <<<SQL
-        SELECT DISTINCT
-            r.display as realm,
-            gb.name as group_by,
-            !agb.enabled as not_enabled,
-            agb.visible
+SELECT
+    r.name as realm,
+    gb.name as group_by,
+    MAX(agb.enabled) as enabled,
```
jpwhite4, line 990: looks like you are going to have to do ``!MAX(agb.enabled) as not_enabled` so painful.

classes/Models/Services/Acls.php
jpwhite4: Since the group is by 1,2 the order by 3,4 is meaningless.

jpwhite4: If you want to enforce ordering then ORDER BY 1,2 is sufficient

PR #1294, opened 2020-04-29
Fix report generator download endpoint.

jpwhite4:
Fix regression added in https://github.com/ubccr/xdmod/pull/1219
The report identifiers look different when the report is autogenerated
by the user dashboard infrastructure.

Unit test has been updated to test this condition too.


classes/DataWarehouse/Access/ReportGenerator.php
```diff
@@ -25,7 +25,7 @@ class ReportGenerator extends Common
     /* The report download endpoint retrieves the report data from a temporary
      * directory that is created dynamically based on the report_id
      */
-    const REPORT_TMPDIR_REGEX = '/^[0-9]+-[0-9\.]+-[a-zA-Z0-9\.]+$/';
+    const REPORT_TMPDIR_REGEX = '/^[0-9]+-(?(?=autogenerated)autogenerated-[a-z]+|[0-9\.]+)-[a-zA-Z0-9\.]+$/';
```
plessbd, line 28: I do not believe the positive look ahead is needed

PR #1298, opened 2020-04-30
Update data warehouse export UI

jtpalmer:
## Description

Adds a separate column for downloading exported data and adds the descriptions of all the fields that are exported for the selected realm.

Before:
data-export-before

After:
data-export-after

Also, refactor raw statistics configuration file use.

## Motivation and Context

https://app.asana.com/0/15230011293693/1137016263723284

## Tests performed

Altered automated tests and tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Data/RawStatisticsConfiguration.php
+++ b/classes/DataWarehouse/Data/RawStatisticsConfiguration.php
@@ -1,0 +1,194 @@
+<?php
+
+namespace DataWarehouse\Data;
+
+use Configuration\XdmodConfiguration;
+use Exception;
+
+/**
+ * Provides access to `rawstatistics` configuration data.
+ */
+class RawStatisticsConfiguration
+{
+    /**
+     * Singleton instance.
+     * @var \DataWarehouse\Data\RawStatisticsConfiguration
+     */
+    private static $instance;
+
+    /**
+     * Raw statistics configuration parsed as array of arrays.
+     * @var array[]
+     */
+    private $config;
+
+    /**
+     * Factory method.
+     *
+     * @return \DataWarehouse\Data\RawStatisticsConfiguration
+     */
+    public static function factory()
+    {
+        if (!isset(static::$instance)) {
+            static::$instance = new static();
+        }
+        return static::$instance;
+    }
+
+    /**
+     * Load the `rawstatistics` configuration files.
+     */
+    private function __construct()
+    {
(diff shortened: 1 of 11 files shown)


jpwhite4:
Can you put a css style on the help text so that it matches the font face of the rest of the interface? It looks a bit jarring to me to have the help text in a different font.

classes/DataWarehouse/Data/RawStatisticsConfiguration.php
```diff
@@ -0,0 +1,194 @@
...
+                $realm
+            ));
+        }
+
+        return array_map(
+            function ($field) {
+                // Replace hierarchy constants.
+                foreach (['name', 'documentation'] as $key) {
+                    $value = $field[$key];
+                    if (strpos($value, 'HIERARCHY_') === 0 && defined($value)) {
```
jpwhite4, line 144: I thought that the XDMoDConfiguration class used backend code that translated the HEIRACHY variables automagically. Is this not the case on this code path?

jtpalmer:
I've added the same CSS class as used by the Usage descriptions:
usage-tab-help-text

help-text-font-change

classes/DataWarehouse/Data/RawStatisticsConfiguration.php
jtpalmer: That is not the case on this code path.  (This isn't new code, it's just moved to a different file.)

I am not aware of any code path where that is the case, but if you know of one then please send me the details.

jtpalmer:
@jpwhite4 please let me know if you have any other suggestions for the graphical design.  I couldn't think of much beside just putting everything in a list.  Also, do you think the heading ("{REALM} Realm Fields") is necessary or should otherwise be changed in content or style?

classes/DataWarehouse/Data/RawStatisticsConfiguration.php
jtpalmer: It looks like there is a feature in the group by refactor that does something like what you described, but it isn't implemented at the configuration class level.  It's part of `\Realm\Realm`.

jpwhite4: Ok. I assumed that the variable substitution stuff was implemented lower down the stack that in actually is.

jpwhite4:
The title "{REALM} realm fields" seems OK to me.

PR #1300, opened 2020-05-01
Add README to data warehouse export zip files

jtpalmer:
## Description

Adds a `README.txt` file to the data warehouse export zip file that contains a description of all the exported fields.

~~**NOTE: This depends on and includes all the commits from  #1298 as of now**~~

## Motivation and Context

https://app.asana.com/0/15230011293693/1146676887787308

## Tests performed

Updated automated tests.  Tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Export/FileManager.php
+++ b/classes/DataWarehouse/Export/FileManager.php
@@ -4,6 +4,7 @@ namespace DataWarehouse\Export;
 
 use CCR\Loggable;
 use DataWarehouse\Data\BatchDataset;
+use DataWarehouse\Data\RawStatisticsConfiguration;
 use DataWarehouse\Export\FileWriter\FileWriterFactory;
 use Exception;
 use Log;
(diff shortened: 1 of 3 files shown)


classes/DataWarehouse/Export/FileManager.php
```diff
@@ -309,4 +328,27 @@ public function removeDeletedRequests(array $deletedRequestIds)
...
+    private function getReadmeText($realm)
+    {
+        $fields = $this->rawStatsConfig->getBatchExportFieldDefinitions($realm);
+
+        $text = "{$realm} Realm Fields\n\n";
+
+        foreach ($fields as $field) {
+            $text .= sprintf(
+                "%s: %s\n",
+                $field['name'],
```
jpwhite4, line 346: Do we allow html in the documentation field? Is it work htmldecode this before putting in the text file?

jtpalmer: As of now, none of the documentation fields contain html.  I don't think there is anything to forbid it.

I don't see any entities, but if we do decide to allow arbitrary html then just decoding wouldn't be enough since we'd still have to strip out any tags.

PR #1303, opened 2020-05-04
Improve performance of OpenStackCloudEventAssetRootVolumeIngestor and GenericCloudEventAssetRootVolumeIngestor actions

eiffel777:
The action xdmod.OpenStackCloudEventAssetRootVolumeIngestor is taking over two minutes to complete when ingesting data from the IU Jetstream system. Looking into it more there are two main performance issues, the first is that it selects all rows from the event table and the second is a join on a column that was not indexed. Adding an index to that column and making changes to the sql query so it does not select all rows from the event table result in a action now taking between 1 and 2 seconds when tested on xdmod-dev. 

## Tests performed
Tested on xdmod-dev.

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_common/asset.json
+++ b/configuration/etl/etl_tables.d/cloud_common/asset.json
@@ -99,6 +99,13 @@
                     "asset_id"
                 ],
                 "is_unique": true
+            },
+            {
+                "name": "idx_provider_identifier",
+                "columns": [
+                    "provider_identifier"
+                ],
+                "is_unique": false
             }
         ]
     }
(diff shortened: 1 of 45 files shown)


configuration/etl/etl_tables.d/cloud_common/asset.json
```diff
@@ -99,6 +99,13 @@
                     "asset_id"
                 ],
                 "is_unique": true
+            },
+            {
+                "name": "idx_provider_identifier",
+                "columns": [
+                    "provider_identifier"
```
jpwhite4, line 106: Shouldn't resource_id also be part of this index?

jpwhite4: Since presumably this is here to improve the performance of the following join?
```
a.resource_id = ev.resource_id AND a.provider_identifier = CONCAT('root-vol-', i.provider_identifier)"
```

eiffel777: @jpwhite4 It's been added

PR #1304, opened 2020-05-05
remove duplicated code

plessbd:
sonarcloud has been complaining about this code.
This just calls the other function with the different data.


jpwhite4:
This code creates an XDMoD helper that is not logged out.

PR #1314, opened 2020-05-18
revert mistakenly removed line from #1308

plessbd:
I accidentally removed a file still 'in use' (I dont really know if it is used, but it should not have been removed since I dont really know what it is doing)


jtpalmer:
See #1308 (GitHub didn't link the ticket with the number only in the title.)

PR #1318, opened 2020-05-26
Fix batch export realm access

jtpalmer:
## Description

Change function used for data warehouse export access control.

## Motivation and Context

There's only one realm that has a different name and display, and it's only in XSEDE as of now.  The "name" is used by the `rawstatistics` JSON files, so that is what must be used.

## Tests performed

No new tests.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Access/BatchExport.php
+++ b/classes/DataWarehouse/Access/BatchExport.php
@@ -20,7 +20,7 @@ class BatchExport
     public static function getBatchExportRealms(XDUser $user)
     {
         $config = RawStatisticsConfiguration::factory();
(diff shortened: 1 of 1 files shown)


ryanrath:
Ok nvm, in metrics-dev & xdmod-dev it looks like both `moddb.realms.name` and `moddb.realms.display` are upper cased so really either function should work.

jtpalmer:
I should have mentioned it, but the space in "Resource Allocations" is causing the problem.

jtpalmer:
Also, just as a note, this was probably broken by #1298

PR #1321, opened 2020-05-27
Update ETL executeSQL logging

plessbd:
Continuing my quest for better logging
This moves notification of start of executeSQL to a notice level, and moves some debugging into info for analysis later

### DIFF
--- a/classes/ETL/Maintenance/ExecuteSql.php
+++ b/classes/ETL/Maintenance/ExecuteSql.php
@@ -193,12 +193,11 @@ class ExecuteSql extends aAction implements iAction
                 }
             }  // if ( is_object($sqlFile) )
 
-            $this->logger->info("Processing SQL file '$filename' with delimiter '$delimiter'");
 
             $sqlFileContents = file_get_contents($filename);
             $sqlFileContents = $this->variableStore->substitute(
                 $sqlFileContents,
-                "Undefined macros found in SQL"
+                "Undefined macros found in SQL file " . $filename
             );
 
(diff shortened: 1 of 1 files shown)


jtpalmer:
What is the reason for increasing the log level for these messages?  What's wrong with `info`?

plessbd:
So that when running this manually, something is output instead of see something like this (not this specific one but similar)
Where you get an end notice on something but no idea what is happening now (unless you change the log level or look at the database).

Ideally I want to also be consistent on start and end notices being at the same level.

```
2020-05-27 11:57:56 [notice] aggregate end (unit: month, periods: 1, start_date: 2020-05-13 11:40:49, end_date: none, start_time: 1590594925.4813, end_time: 1590595076.2817, elapsed_time: 150.80048)
2020-05-27 12:00:31 [notice] aggregate end (unit: quarter, periods: 1, start_date: 2020-05-13 11:40:49, end_date: none, start_time: 1590595079.357, end_time: 1590595231.5896, elapsed_time: 152.23261)
```

switching from debug -> info is so that things get logged to the database for better analysis

jtpalmer:
We can always start logging debug messages to the database.

plessbd:
we could, this was just a small step.

plessbd:
Agree, but at the same time not enough verbosity is also bad.  That is what we 'should' have `-q` for...

jtpalmer:
That is absolutely not what `-q` is for. Quiet should suppress everything that isn't an error (or warning).

PR #1327, opened 2020-06-03
Update xdmod-check-config script

jtpalmer:
## Description

Adds checks for PHP APCu and Multibyte String functions to make sure the corresponding extensions are installed and enabled.

## Motivation and Context

When I was removing the mcrypt requirement I noticed that checks for these extensions were missing.  These checks serve as a sanity check to assist in debugging.

## Tests performed

Tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/bin/xdmod-check-config
+++ b/bin/xdmod-check-config
@@ -191,6 +191,13 @@ try {
         'PHP PCRE extension installed'
     );
 
+    _debug('Checking for mb_convert_encoding function');
+    $result = function_exists('mb_convert_encoding');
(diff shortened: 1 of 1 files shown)


jtpalmer:
The test for `mbstring` isn't strictly necessary since we're including a polyfill ([symfony/polyfill-mbstring](https://github.com/symfony/polyfill-mbstring) is a dependency of a dependency of Silex).  That brings up the question of why we include `php-mbstring` as a dependency of the RPM package if we're going to ship a polyfill.

PR #1335, opened 2020-06-10
Change table alias for consistency

jtpalmer:
## Description

Changes the alias used for the aggregate table in the jobs realm `JobDataset` query.  

## Motivation and Context

The group by refactoring changed the table alias from `jf` to `agg`.  This file was not changed so it was still using `jf` as an associative array key in the PHP code even though the generated SQL is now using `agg`.  The change makes it clear that `agg` in the PHP code is the same `agg` in the corresponding SQL.  It also prevents another table from being aliased to `agg` and allowing other table to used `jf` as an alias if necessary.

## Tests performed

Tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Jobs/JobDataset.php
+++ b/classes/DataWarehouse/Query/Jobs/JobDataset.php
@@ -24,8 +24,8 @@ class JobDataset extends \DataWarehouse\Query\RawQuery
 
         $config = RawStatisticsConfiguration::factory();
 
-        // The data table is always aliased to "jf".
-        $tables = ['jf' => $this->getDataTable()];
+        // The data table is always aliased to "agg".
+        $tables = ['agg' => $this->getDataTable()];
 
         foreach ($config->getQueryTableDefinitions('Jobs') as $tableDef) {
             $alias = $tableDef['alias'];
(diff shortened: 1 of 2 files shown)


jtpalmer:
This also fixes an issue created by ubccr/xdmod-xsede#209 (which also changes `jf` to `agg`).

plessbd:
Yeah the refactor changed the alias to be "more readable", for some of the supremm stuff it did not change (yet) to make the refactor code easier to review (and because the supremm stuff was auto generated).  Sorry if I missed some.

jtpalmer:
This specific change wasn't actually required by the other changes (because `jf` was only being used as a key in a PHP array.), but when I was changing the XSEDE realms classes (which really did need to change because they contained a SQL fragment that used `jf`) I changed all the realms.  Since I changed the XSEDE jobs realm I broke it (because I didn't change the PHP in this repository).  It's more confusing than anything to be using `jf` in the PHP classes, but `agg` in the generated SQL.  That's my primary motivation in making this change (and that since there was and XSEDE class that did need to change).  I just added my previous comment to create a link between the pull requests and for future reference.

PR #1337, opened 2020-06-12
Remove legacy code.

jpwhite4:
The only path this code took was to redirect the post reponse from the sucessful internal_dashboard login back to the internal_dashboard. 

Perhaps in the past this was used for different purposes, but it is not needed in the current incarnation of the code.


jtpalmer:
That `direct_to` is also in `xdmod-supremm`: https://github.com/ubccr/xdmod-supremm/blob/af2ef772da46a3e515c6e6c6275f58932105682e/html/internal_dashboard/supremm/user_check.php#L33

PR #1338, opened 2020-06-13
Refactor Query class to allow DISTINCT queries

jtpalmer:
## Description

Adds an option to the Data Warehouse Query class to support queries that use `DISTINCT`.  Refactors subclasses to use this functionality.

Also removes a block of dead code.

## Motivation and Context

Reduce duplicate and similar code.  See #1332 for duplicate code.

## Tests performed

Uses existing tests.

## Types of changes

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Query.php
+++ b/classes/DataWarehouse/Query/Query.php
@@ -1452,7 +1452,7 @@ SQL;
      *
      * @param bool $distinct
      */
-    public function setDistinct($distinct)
+    protected function setDistinct($distinct)
     {
         $this->isDistinct = $distinct;
     }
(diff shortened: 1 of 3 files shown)


classes/DataWarehouse/Query/Query.php
```diff
@@ -1449,6 +1447,26 @@ public function isTimeseries()
         return ($this instanceof TimeseriesQuery);
     }
 
+    /**
+     * Set the query to use DISTINCT or not.
+     *
+     * @param bool $distinct
+     */
+    public function setDistinct($distinct)
```
jpwhite4, line 1455: I'd consider making this protected since I don't like the idea of code outside of the Query classes being able to change the meaning of the query.

jtpalmer: Will do.  Unlikely it will be used anywhere else anytime soon anyway.

PR #1340, opened 2020-06-16
WIP: Webserver (mainly security) Configuration Updates

jpwhite4:
## Types of changes

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -120,56 +120,68 @@ See the [Hierarchy Guide](hierarchy.html) for more details.
 Apache Configuration
 --------------------
 
-Uses port 8080 by default, if changed, must also be changed in
-`portal_settings.ini`.
+A template apache configuration file is provided. The path is `/etc/httpd/conf.d/xdmod.conf`
+in the RPM install and `etc/apache.d/xdmod.conf` in the source code install.
 
+The template configuration file includes appropriate configuration settings
+to enable the webserver listening on HTTPS port 8080.
+
+Valid SSL certificates will need to be installed and configured.  The template
+configuration file must be edited to specify the path to the SSL certificate
+file and SSL certificate key file. Refer to the [Apache SSL documentation](https://httpd.apache.org/docs/2.4/ssl/)
+for SSL configuration information.
+
+The name and port of the server must match with the `site_address` and `user_manual`
+configuration settings in `portal_settings.ini`.
+
+The template configuration file also configures the webserver to send the `Strict-Transport-Security` HTTP Header
+to indicate to  web browsers that the XDMoD instance should only be accessed using HTTPS.
+
+```apache
     Listen 8080
     <VirtualHost *:8080>
-        DocumentRoot /usr/share/xdmod/html
-        <Directory /usr/share/xdmod/html>
-            Options FollowSymLinks
-            AllowOverride All
-            DirectoryIndex index.php
-            # Apache 2.4 access controls.
-            <IfModule mod_authz_core.c>
-                Require all granted
-            </IfModule>
-        </Directory>
-        <Directory /usr/share/xdmod/html/rest>
-            RewriteEngine On
-            RewriteRule (.*) index.php [L]
-        </Directory>
-    </VirtualHost>
+        # The ServerName and ServerAdmin parameters should be set to the
+        # appropriate values for the XDMoD instance.
+        ServerName localhost
+        #ServerAdmin postmaster@localhost
 
-We recommend that you use HTTPS in production.  This will require
-additional configuration.
-
-    Listen 443
-    <VirtualHost *:443>
-
-        # Customize this section using your SSL certificate.
+        # Production XDMoD instances should use HTTPS
         SSLEngine on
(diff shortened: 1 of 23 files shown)


plessbd:
Why does this work without changing the chrome options to `allow-insecure-localhost`?

docs/configuration.md
```diff
@@ -120,56 +120,85 @@ See the [Hierarchy Guide](hierarchy.html) for more details.
...
+        #<Directory /usr/share/xdmod/vendor/simplesamlphp/simplesamlphp/www>
+        #    Options FollowSymLinks
+        #    AllowOverride All
+        #    # Apache 2.4 access controls.
+        #    <IfModule mod_authz_core.c>
+        #        Require all granted
+        #    </IfModule>
+        #</Directory>
+
+        ErrorLog /var/log/xdmod/apache-error.log
```
plessbd, line 198: I guess this is considered a DoS attack vector, should we bother having it in here? Or should we just rely on the default apache logging?

ryanrath: That depends on how we have things setup w/ logrotate. If we only rotate on a time based metric and don't include size then yes, it can be a DoS vector. Thankfully logrotate v3.8.1+ includes a `maxsize` directive that we can also set to mitigate this vector.

plessbd: That only kind of slows it down.  It still needs space to make the archive before it can delete the current log file, and if it isnt on a different file system can cause issues.

So my thought is to remove it and put in documentation to "follow best practices for the web server of your choice"

ryanrath: That only kind of slows it down? How do you figure? We already have things setup w/ `rotate 4` so if we add something like `maxsize 250M` then the logs should take up no more then ~1G total. Unless I'm missing how logrotate works, @jtpalmer  care to weigh in? Also, I'm going to be firmly in the "no" camp of removing logging from XDMoD.

plessbd: I was not suggesting removing logging, just using the defaults that Apache / webserver puts in place already.

And you are right, I wasnt thinking of the only keeping X part of that.  Was just thinking rotating at a max size.  I take that back.

Still, part of me thinks just using the defaults the webserver uses keeps us out of recommending things that are "current best practices" by people that know more.

jtpalmer: The default log rotate only runs once a day.  The `maxsize` won't prevent growing over that size.  My understanding is that that option should be used if you don't want to rotate unless over the maximum size.

I don't think using the default Apache `ErrorLog` configuration will make much difference as far as the size of the logs are concerned, it will just put everything in a single file if I'm reading this correctly:

```
#
# ErrorLog: The location of the error log file.
# If you do not specify an ErrorLog directive within a <VirtualHost>
# container, error messages relating to that virtual host will be
# logged here.  If you *do* define an error logfile for a <VirtualHost>
# container, that host's errors will be logged there and not here.
#
ErrorLog "logs/error_log"
```

(Same for `CustomLog` access log.)

We've already been told that instructions aren't good enough and we need to be secure by default, so if that is really the case then we may need to not remove the logging, but disable any logging for the vhost.  I'm not sure how that could be good for security since the access log would be useful to tell you the IP that is attacking the server.

I'd rather not remove logging since it's going to make handling support issues much more difficult.

docs/configuration.md
```diff
@@ -120,56 +120,85 @@ See the [Hierarchy Guide](hierarchy.html) for more details.
...
+to indicate to  web browsers that the XDMoD instance should only be accessed using HTTPS.
+
+```apache
     <VirtualHost *:443>
+        # The ServerName and ServerAdmin parameters should be updated.
+        ServerName localhost
+        ServerAdmin postmaster@localhost
 
-        # Customize this section using your SSL certificate.
+        # Production XDMoD instances should use HTTPS
```
jtpalmer, line 153: Did you not like my suggestions to use "Open XDMoD" instead of "XDMoD"?

docs/configuration.md
```diff
@@ -120,56 +120,85 @@ See the [Hierarchy Guide](hierarchy.html) for more details.
 Apache Configuration
 --------------------
 
-Uses port 8080 by default, if changed, must also be changed in
-`portal_settings.ini`.
+A template apache configuration file is provided. The path is `/usr/share/xdmod/templates/apache.conf`
```
jtpalmer, line 123: Do you not like my suggestions to use "Apache" instead of "apache"?

docs/configuration.md
jtpalmer: Just noticed that I was talking about `logrotate` and @ryanrath may have been referring to [`rotatelogs`](https://httpd.apache.org/docs/2.4/programs/rotatelogs.html).  Looks like the solution is to stop using `logrotate` and switch to the Apache program.

docs/configuration.md
jpwhite4: I do agree with the suggestions. For some reason Ben closed the original pull request before I could update it with you suggested text. This branch is not intended to be an actual pull request (which is why I have the experiment tag on it).

jtpalmer: Thanks for clarifying.  I did not notice that tag.

PR #1344, opened 2020-06-19
Remove bug that causes stop events to create sessions

eiffel777:
This PR fixes a bug in the CloudStateReconstructorTransformer that allowed end events such as Stop, Power Off or Shelve to start a new session. This is fixed by adding an if statement that prevents an end event from creating a new session.

A new openstack log file has been added for the regression tests. It is ingested separately to ensure we create a situation where sessions would have been incorrectly made before this change. This change means this data should not cause any new sessions to be made. 

The migration script will look for sessions with end events, display the list of sessions to the user and ask if they want to remove the session. Re-aggregation of the cloud data is already necessary for upgrading to 9.0. Below is a screenshot of how this is displayed when running the upgrade.

Screenshot from 2020-06-19 08-34-48

## Tests performed
Tested manually in docker and also added data to docker that would have created sessions before his issue was fixed.

## Types of changes

## Checklist:


classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
```diff
@@ -71,6 +71,49 @@ public function execute()
...
+                $sessionTable = $this->makeAsciiSessionTable($erroneousSessions);
+                $console->displayMessage(<<<"EOT"
+This version of Open XDMoD fixes a bug in the cloud realm that created erroneous session. We have found $numSessions erroneous sessions.
+
+$sessionTable
+EOT
+                    );
+                $console->displayBlankLine();
+                $deleteErroneousSessions = $console->prompt(
+                    'Would you like to remove these sessions and re-aggregate your cloud data',
```
plessbd, line 105: if reaggregation is automatically done at a different point, we probably just need to inform them of the ones that are getting removed and remove them...

tests/artifacts/xdmod/regression/chartFilterTests.json
```diff
@@ -363,7 +363,7 @@
         },
         "expected": {
             "subtitle": "Domain = ( adjunct,  Default )",
-            "yvalue": 81277.6197
+            "yvalue": 81277.6131
```
plessbd, line 366: why did these values change?  Since this added data that should be ignored, shouldnt these vaules stay the same?

eiffel777: This data changed because there are some sessions for the nutsetters resource that were removed as a result of this change. This does not show up in the regression tests because the data for the nutsetters resource is for 2019-06-26 and the regression test use the time frame 2018-04-18 to 2018-04-30. As an FYI, the data I added uses the 2018-04-18 to 2018-04-30 time range.

classes/OpenXdmod/Migration/Version851To900/DatabaseMigration.php
eiffel777: @plessbd I changed the wording. I still want to keep the prompt so people know and can look at what rows are getting removed.

PR #1348, opened 2020-06-23
CORS changes for OOD

plessbd:
This is to merge what was going to be 8.7 into 9.0


classes/DataWarehouse/Query/Jobs/RawData.php
```diff
@@ -84,6 +84,9 @@ public function __construct(
         $this->addField(new TableField($factTable, 'end_time_ts'));
         $this->addField(new FormulaField('-1', 'cpu_user'));
 
+        // This is used by Integrations and not currently shown on the XDMoD interface
```
jpwhite4, line 87: There will need to be a corresponding change to the supremm module too. Also can you check with Jeff to see if this interferes with his dw export changes

classes/OpenXdmod/Migration/Version851To870/ConfigFilesMigration.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
jpwhite4, line 1: Need to update the migration file with the correct version numbers.

docs/integrations.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Single Sign-on (SSO) Embedding
+Open XDMoD allows for single sign on embedding so that end users do not have to explicitly sign
+in to XDMoD after they have already signed in to the other integrated application.
+When [SSO](simpleSAMLphp.html) is configured with the same IdP XDMoD can be embedded in an
+iframe at the login endpoint and iusers will be automatically logged in.
+
+When Open XDMoD detects a login within an iframe it will send a `postMessage` to `window.top`
+
+The following is an example of how to handle this
```
jpwhite4, line 33: Please can you add another sentence or two to add some context to this example. For example - where should this  example code be added? How should this be used to enable embedded sso logins.

docs/integrations.md
```diff
@@ -0,0 +1,52 @@
...
+To enable integration, the Open XDMoD instance must be configured to give explicit permission to the browser.
+The Open XDMoD application settings to enabled this are described below.
+
+## Open XDMoD Settings
+
+### Cross-Origin Resource Sharing (CORS)
+To allow CORS  a list of domains that are allowed to communicate with Open XDMoD is configured in
+the `domains` setting in the `cors` section of `portal_settings.ini`.
+
+This setting is a comma separated list where each item contains the scheme, host, and optional port.
```
jpwhite4, line 14: need to emphasize that the strings in this setting must _exactly_ match the string sent in the Origin header by the browser. The port is not 'optional' it must be in the setting if it is needed and must be absent if it is absent

docs/integrations.md
```diff
@@ -0,0 +1,52 @@
...
+```
+[cors]
+domains=https://integratedapp.example.tld,https://dev-integratedapp.example.tld:8080
+```
+
+**NOTE: This setting can open up Open XDMoD to security risks if used improperly.
+Only enable it if you know what you are doing**
```
jpwhite4, line 21: How are they supposed to know what they are doing?

classes/DataWarehouse/Query/Jobs/RawData.php
jtpalmer: This shouldn't cause any problems with the dw export.

docs/integrations.md
plessbd: I meant optional port since the port is not sent for `https` and `http` when using their standard ports, but I will update this.

PR #1356, opened 2020-06-26
Enforce secure cookie flag for all session cookies and the rest token.

jpwhite4:
This also enables https:// for the CI builds.

The helper function start_session() was the cleanest way I could think
of to enforce the secure falg for the session cookie. If anyone can think
of a better way then let me know.


shippable.yml
```diff
@@ -37,7 +37,7 @@ build:
         - ./tests/ui/runtests.sh --headless --log-junit `pwd`/shippable/testresults --sso
         - ./vendor/phpunit/phpunit/phpunit -c ./tests/integration/phpunit.xml.dist --testsuite sso --log-junit `pwd`/shippable/testresults/xdmod-sso-integration.xml
         - mv ./configuration/portal_settings.ini.old ./configuration/portal_settings.ini
-        - test ! -s /var/log/xdmod/apache-error.log
+        - test `fgrep -v ' [ssl:warn] ' /var/log/xdmod/apache-error.log | wc -l` = 0
```
jpwhite4, line 40: This is needed because the hostname of the autogenerated ssl cert is not localhost so the webserver logs a warning (but starts up anyway).

PR #1361, opened 2020-07-01
Change table definition for vcpus field for the cloud_resource_specs table 

eiffel777:
The vcpus field for the cloud_resource_specs table had the nullable property set to be true. This is incorrect and in MySQL 5.7 or higher causes an error since this field is part of the primary key.

```
xdmod.jobs-cloud-common.CloudTableManagement (ETL\Maintenance\ManageTables): Error managing ETL table `modw_cloud`.`cloud_resource_specs`: SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
```
Setting the field to be `nullable: false` fixes the problem

## Tests performed
Tested manually

## Types of changes

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_common/cloud_resource_specs.json
+++ b/configuration/etl/etl_tables.d/cloud_common/cloud_resource_specs.json
@@ -32,7 +32,8 @@
             {
                 "name": "vcpus",
                 "type": "int(5)",
-                "nullable": false
+                "nullable": false,
+                "default": null
             },
             {
                 "name": "start_date_ts",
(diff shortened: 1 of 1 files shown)


jpwhite4:
Is it possible that there could be null values in the table from an earlier XDMoD version?

eiffel777:
@jpwhite4 This table was added in the cloud utilization PR(#1242) and doesn't exist in previous version of XDMoD that should not be something that needs to be worried about, unless I am misunderstanding your question.

PR #1379, opened 2020-07-16
Ensuring that Users have a username

ryanrath:
**NOTE: this is up for discussion / inclusion in 9.5**

## Description
Just something we ran across while testing things for the OpenOnDemand / XDMoD tutorial. Turns out that you can end up with user's without a username. These changes ensure that when `saveUser` is called that an exception is thrown if `_username` is empty. We've also updated the username column to not accept null values & to also have a unique index. 

## Motivation and Context
No more null username Users!

## Types of changes

## Checklist:


configuration/etl/etl_tables.d/xdb/users.json
```diff
@@ -12,7 +12,7 @@
             {
                 "name": "username",
                 "type": "varchar(200)",
-                "nullable": true
+                "nullable": false
```
jpwhite4, line 15: Should we worry about the situation where an existing install has a null username (for whatever reason) and then the upgrade fails?

plessbd: We talked about that exact thing.  That is why this was moved to 9.5 for the moment, so we can think about all the ways this could already be broken.  I cant see a way for the user to login, so I would say we should just delete the user that have null... as part of the upgrade

classes/XDUser.php
```diff
@@ -923,6 +923,10 @@ public function saveUser()
             throw new Exception('A user must have at least one acl.');
         }
 
+        if (is_empty($this->_username)) {
```
plessbd, line 926:
```suggestion
        if (empty($this->_username)) {
```

ryanrath:
**Note:**  For anybody reviewing this one. SonarCloud has a hissy fit whenever you merge in changes from upstream, rebase some local changes on top of that and then force push it up. The duplicate code has nothing to do with this PR.

jpwhite4 (approved):

PR #1380, opened 2020-07-17
Updates to cloud migration files to fix syntax issues

eiffel777:
Not all of the sql statements in cloud-migration-8-5-1_9-0-0.sql were being run because of syntax issues. This adds '//' between the statements so they run. The queries were also running a slower than expected so indexes were added the speed this up. The indexes are removed at the end of the file. The decreased the time from around 10 minutes to around 2 minutes.

Also in testing I found that that the section during the upgrade that shows incorrect sessions due to the bug in #1344 was returning multiple rows for each incorrect session due to not having the right join parameters. Adding a join on resource_id fixes this.

## Tests performed
Tested in docker manually and with a copy of modw_cloud from production.

## Types of changes

## Checklist:


jtpalmer:
Why don't you replace all the `;` with `//`?

jtpalmer:
You can also set the delimiter like this: https://github.com/ubccr/xdmod-supremm/blob/xdmod9.0/configuration/etl/etl.d/supremm-migration-8_5_1-9_0_0.json#L25

jtpalmer:
I'm still not clear on what works and what doesn't but the wrong statement count is reported unless the delimiter is used everywhere:

```
2020-07-17 16:20:29 [notice] Processing SQL file '/etc/xdmod/etl/etl_sql.d/migrations/cloud-migration-8-5-1_9-0-0.sql' using delimiter '//' containing 14 statements
2020-07-17 16:20:29 [notice] Finished Processing 14 SQL statements
2020-07-17 16:20:29 [notice] (action: xdmod.migration-8_5_1-9_0_0.DeleteCloudAutoincrementColumns (ETL\Maintenance\ExecuteSql), start_time: 1595002829.2376, end_time: 1595002829.3798, elapsed_time: 0.14219)
```

eiffel777:
@jtpalmer I didn't know(or forgot) that I could set the delimiter. I changed the delimiter to a semicolon. I think all the queries ran before but the wrong count was displayed.

PR #1386, opened 2020-07-20
Remove beta labels from documentation

jtpalmer:
## Description

Update documentation to remove beta labels.

## Motivation and Context

These features are now production ready.

https://app.asana.com/0/342819846538629/1183603556135434/f

## Types of changes

## Checklist:


docs/dashboard.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: User Dashboard (Beta)
+title: User Dashboard
```
plessbd, line 2: Only "issue" i have with this is if we are going to make it the default or not, but I guess that doesnt mean it is not beta anymore...Just that it is not enabled by default, should we include something in there about how to enable it? either through setup or by changing the value in `portal_settings.ini`

jtpalmer: During the meeting this morning there was discussion about documenting how to enable it so I think that's the way to go.  I'll create a task in Asana so we don't forget.

PR #1389, opened 2020-07-20
Adding an alert to notify users when ingestion has not been run.

ryanrath:
## Description
Adding an informational alert for when either ingestion or `acl-config` has
not been run.

This is to address:
https://app.asana.com/0/342819846538629/1185258719202835

## Tests performed
Manually tested. 

## Types of changes

## Checklist:

### DIFF
--- a/html/index.php
+++ b/html/index.php
@@ -248,7 +248,7 @@ $page_title = xd_utilities\getConfiguration('general', 'title');
                 Ext.onReady(function() {
                     Ext.MessageBox.alert(
                         'Invalid State Detected',
-                        'This XDMoD installation has had an invalid state detected. <br/><br/>' +
+                        'This Open XDMoD installation is in an invalid state.<br/><br/>' +
                         'Before XDMoD can be utilized you must run ingestion at least once.<br/><br/>' +
                         'If ingestion has already been run then make sure that you have run the acl-config cli application and that it has completed successfully.<br/><br/>' +
                         '<b>NOTE: XDMoD will not function correctly until this problem has been resolved.</b>'
--- a/html/index.php
+++ b/html/index.php
@@ -249,7 +249,7 @@ $page_title = xd_utilities\getConfiguration('general', 'title');
                     Ext.MessageBox.alert(
                         'Invalid State Detected',
                         'This Open XDMoD installation is in an invalid state.<br/><br/>' +
-                        'Before XDMoD can be utilized you must run ingestion at least once.<br/><br/>' +
+                        'Before Open XDMoD can be utilized you must run ingestion at least once.<br/><br/>' +
                         'If ingestion has already been run then make sure that you have run the acl-config cli application and that it has completed successfully.<br/><br/>' +
                         '<b>NOTE: XDMoD will not function correctly until this problem has been resolved.</b>'
                     );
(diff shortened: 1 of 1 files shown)


html/index.php
```diff
@@ -235,16 +235,29 @@ function isReferrer($referrer)
...
+            return $carry;
+        }, array());
+
+        if (count($realms) === 0) {
+            $code = <<<JS
+                Ext.onReady(function() {
+                    Ext.MessageBox.alert(
+                        'Invalid State Detected',
+                        'This XDMoD installation has had an invalid state detected. <br/><br/>' +
+                        'Before XDMoD can be utilized you must run ingestion at least once.<br/><br/>' +
```
plessbd, line 252: so does the no aggregated data message show up somewhere else? Cause I think we have to have both ingestion and aggregation done before it actually works?

ryanrath: I'm not sure to be honest, but I can update the phrasing to reference both ingestion / aggregation. The direct cause of this problem is that there are no records in `moddb.realms` when they try to load the front page.

jtpalmer: I tried it and the one, new message, is displayed.  Nothing after that.

html/index.php
```diff
@@ -235,16 +235,29 @@ function isReferrer($referrer)
...
+                        'If ingestion has already been run then make sure that you have run the acl-config cli application and that it has completed successfully.<br/><br/>' +
```
jtpalmer, line 253:
```suggestion
                        'If ingestion has already been run then make sure that you have run the acl-config CLI application and that it has completed successfully.<br/><br/>' +
```

html/index.php
jtpalmer, line 252:
```suggestion
                        'Before Open XDMoD can be utilized you must run ingestion at least once.<br/><br/>' +
```

html/index.php
```diff
@@ -235,16 +235,29 @@ function isReferrer($referrer)
...
+                        '<b>NOTE: XDMoD will not function correctly until this problem has been resolved.</b>'
```
jtpalmer, line 254:
```suggestion
                        '<b>NOTE: Open XDMoD will not function correctly until this problem has been resolved.</b>'
```

html/index.php
ryanrath: @jtpalmer did you add all of the changes from the PR? I know that before I added the `try-catch` around the `disabledMenus` section the page came up empty because the exception from `Acls::getDisabledMenus` function ultimately ended up outputting a blob 'o json describing the exception right in the middle of those js variables.

plessbd: I guess we only have it for jobs, and only when we are not using the dashboard
https://github.com/ubccr/xdmod/blob/xdmod9.0/html/controllers/ui_data/summary3.php#L71

Wonder if we should find a way to test for it...

ryanrath: Test for what? Not having run ingestion / aggregation and / or acl-config?

plessbd: yes :)

jtpalmer: @ryanrath I testing using your branch (at the time that I posted the comment).  I did not shred or ingest any jobs so it's possible that if I had done that, but not aggregated that something different would happen.

PR #1390, opened 2020-07-20
Rename internal dashboard button from "Dashboard" to "Admin Dashboard"

jpwhite4:
Also update the tooltip to be slightly less uninformative.

This increases the width of the button. However there is still plenty
of space in the top toolbar. The width of the page that causes the
Help button to disappear is still mcuh narrower than any usable width
for the rest of XDMoD.

image

The rest of the labels and paths are unchanged

### DIFF
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -128,11 +128,11 @@ XDMoD.GlobalToolbar.Profile = {
 // -------------------------------------------------
 
 XDMoD.GlobalToolbar.Dashboard = {
-    text: 'Dashboard',
+    text: 'Admin Dashboard',
     scale: 'small',
     iconCls: 'btn_dashboard',
     id: 'global-toolbar-dashboard',
-    tooltip: 'Internal Dashboard',
+    tooltip: 'Load the admin dashboard in a new window.',
     handler: function () {
         XDMoD.TrackEvent("Portal", "Dashboard Button Clicked");
         CCR.xdmod.initDashboard();
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -134,7 +134,7 @@ XDMoD.GlobalToolbar.Dashboard = {
     id: 'global-toolbar-dashboard',
     tooltip: 'Load the admin dashboard in a new window.',
     handler: function () {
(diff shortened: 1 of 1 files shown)


html/gui/js/CCR.js
```diff
@@ -128,11 +128,11 @@ XDMoD.GlobalToolbar.Profile = {
...
 XDMoD.GlobalToolbar.Dashboard = {
-    text: 'Dashboard',
+    text: 'Admin Dashboard',
     scale: 'small',
     iconCls: 'btn_dashboard',
     id: 'global-toolbar-dashboard',
-    tooltip: 'Internal Dashboard',
+    tooltip: 'Load the admin dashboard in a new window.',
     handler: function () {
         XDMoD.TrackEvent("Portal", "Dashboard Button Clicked");
```
jtpalmer, line 137: Is there a reason to not change this event text?

jpwhite4: I changed it because left alone we would have a button that says "Admin Dashboard" with a tooltip that says "Internal Dashboard". So (a) the tooltip does not tell you anything more than the button and (b) the tooltip somewhat contradicts the button - is it an internal dashboard or an admin dashboard?

We could, of course, change the button label to "Internal Dashboard" but there is no point having a tooltip that is identical to the button text.

html/gui/js/CCR.js
```diff
@@ -128,13 +128,13 @@ XDMoD.GlobalToolbar.Profile = {
...
-    text: 'Dashboard',
+    text: 'Admin Dashboard',
     scale: 'small',
     iconCls: 'btn_dashboard',
     id: 'global-toolbar-dashboard',
-    tooltip: 'Internal Dashboard',
+    tooltip: 'Load the admin dashboard in a new window.',
     handler: function () {
-        XDMoD.TrackEvent("Portal", "Dashboard Button Clicked");
+        XDMoD.TrackEvent("Portal", "Admin Dashboard Button Clicked");
```
jtpalmer, line 137: Looks like I shouldn't have told you to change this since we require strings to be single quotes.

```suggestion
        XDMoD.TrackEvent('Portal', 'Admin Dashboard Button Clicked');
```

PR #1391, opened 2020-07-21
Convert date for cloud resource specifications to correct timezone from UTC

eiffel777:
The specifications for a cloud system are ingested from a json file in a format that includes a timestamp. The format can be seen in `etl_schemas.d/cloud/resource_specs.schema.json`. The timestamp in these files should be in UTC and when ingested for use with the cloud utilization statistic the timestamp should be changed to the time zone on the server. This is not currently being done. This PR uses a new etl action to create a staging table called `staging_resource_specifications` that converts the timestamp to the time zone on the server and then uses that table as the source for cloud_resource_specs table. 

While fixing the timezone issue I found that the MIN() function was not being used when getting the start time for a unique set of specifications and updated the `etl_action_defs.d/cloud_common/resource_specifications.json` file to fix this.

In Asana, https://app.asana.com/0/342819846538629/1185459844560305
## Tests performed
Tested manually in docker to make sure that time zone was converted correctly and that the resource specification were given the correct start and end times.

## Types of changes

## Checklist:


configuration/etl/etl_action_defs.d/cloud_common/staging_resource_specifications.json
```diff
@@ -0,0 +1,19 @@
...
+      "table_definition": {
+          "$ref": "${table_definition_dir}/cloud_common/staging_resource_specifications.json#/table_definition"
+      },
+      "source_query": {
+          "records": {
+              "resource_id": "raw.resource_id",
+              "hostname": "raw.hostname",
+              "vcpus": "raw.vcpus",
+              "memory_mb": "raw.memory_mb",
+              "fact_date": "DATE(CONVERT_TZ(raw.fact_date, '+00:00', @@session.time_zone))"
```
jtpalmer, line 11: Is it possible to use the time zone specified in `resources.json`?  Does this handle resources in a different time zone than the database server?

eiffel777: @jtpalmer Anywhere else in the cloud realm where we need to convert timezones it does it this way so I'd like to keep it consistent with that.

jtpalmer: Update the documentation to note that cloud resources ignore the timezone configuration option (https://github.com/ubccr/xdmod/blame/xdmod9.0/docs/configuration.md#L424).

PR #1400, opened 2020-07-24
Set the "Show in X tab" label in the Metric Explorer automatically based on whether the dashboard is enabled

jpwhite4:
See https://app.asana.com/0/342819846538629/1145402702473881

This one has probably missed the 9.0.0 train, but I'm putting it up here anyway so it doesn't get forgotten about for 9.5


jpwhite4:
The ternary is in there so that the metric explorer does not break if you switch the sumary/dashboard tab off completely

plessbd (approved):

PR #1401, opened 2020-07-28
Add -c/--cluster flag to xdmod-slurm-helper

treydock:
## Description

Adds -c/--cluster flag to xdmod-slurm-helper.

## Motivation and Context

At OSC for a short period we have a new expansion coming online that is SLURM. The original cluster is Torque.  To allow both in XDMOD we created a second resource called `pitzer-exp` but the actual SLURM cluster name is `pitzer`.  Later this year we will merge the Torque Pitzer into the SLURM expansion Pitzer and both resource name and SLURM cluster name will be `pitzer`.  This change just allows the flexibility to have the cluster name in SLURM differ from the resource name in XDMOD.

## Tests performed

Been running this change for a about a week on OSC's XDMOD test host that shreds and ingests production data.

## Types of changes

## Checklist:


treydock:
This is what my cron execution looks like to make it a bit clearer how this is used:

```
15 1 * * * xdmod /usr/bin/xdmod-slurm-helper --quiet -r pitzer-exp -c pitzer
```

jtpalmer:
@treydock Our current plan is to not add any new functional changes before the 9.0 release.  Is that going to cause problems for you?  You can produce the same result as your change by running `sacct` and `xdmod-shredder` instead of using the helper script.

treydock:
I can maintain a patch until 9.5, that's not a problem.  I was going to do sacct + xdmod-shredder but didn't want to deal with the complicated nature of having correct timestamps and such.  It was easier to patch the helper script than to build up necessary sacct command in cron.

bin/xdmod-slurm-helper
```diff
@@ -398,9 +406,13 @@ Usage: xdmod-slurm-helper [-v] [-r resource]
         Output warning level logging.
 
     -r, --resource *resource*
-        Shred accounting data for the specified cluster.  This value
-        will be passed to the sacct --clusters option.  Unlike the sacct
-        command, only one cluster should be specified.
+        Shred accounting data for the specified cluster.
+       Value is used for --clusters if that argument is not specified.
```
jtpalmer, line 410:
```suggestion
        Value is used for --clusters if that argument is not specified.
```

PR #1423, opened 2020-08-20
HighcartExport process cleanup and updates

plessbd:
`ghostscript` was specifically added to make the pdf have the proper width and height.  since `rsvg-convert` can handle this, remove the dependency on `ghostscript` and change to `exiftool` for reaming metadata

This also:

removes the temporary files and passes them around in memory
uses a safer temporary filename for image export


plessbd:
this will be split into two PR's, one to remove ghostscript, one to cleanup all the files.

PR #1426, opened 2020-08-20
Cleanup exportHighcharts and Related Functions

plessbd:
Makes ExportHighcharts use a safe temporary file and stops putting files all over the place.


libraries/charting.php
```diff
@@ -46,9 +46,9 @@ function exportHighchart(
     $effectiveWidth = (int)($width*$scale);
     $effectiveHeight = (int)($height*$scale);
 
-    $base_filename = sys_get_temp_dir() . '/' . md5(rand() . microtime());
+    $base_filename = tempnam(sys_get_temp_dir(), 'exportHighchart');
```
jpwhite4, line 49: This creates a file which is never used. Presumably you actually want to do something like this:
```php
$tmp_html_filename = tempnam(sys_get_temp_dir(), 'exportHighchart');
if ($tmp_html_filename === false) {
    throw \Exception('Error creating temporary file for chart export');
}
```

libraries/charting.php
```diff
@@ -46,9 +46,7 @@ function exportHighchart(
     $effectiveWidth = (int)($width*$scale);
     $effectiveHeight = (int)($height*$scale);
 
-    $base_filename = sys_get_temp_dir() . '/' . md5(rand() . microtime());
-
-    $tmp_html_filename     = $base_filename . '.html';
+    $tmp_html_filename = tempnam(sys_get_temp_dir(), 'exportHighchart');
```
jpwhite4, line 49: need to check for error return in the tempnam() function call.

libraries/charting.php
```diff
@@ -91,15 +81,28 @@ function exportHighchart(
...
  */
-function getScreenFromChromium($file, $width, $height, $format){
+function getScreenFromChromium($html, $width, $height, $format){
+
+    // Chromium requires the file to have a .html extension
+    // cant use datauri as it will not execute embdeeded javascript
+    $tmpFile = tempnam(sys_get_temp_dir(), 'xdmod-chromiumHtml-');
+    $tmpHtmlFile = $tmpFile . '.html';
+    if ($tmpFile === false || rename($tmpFile, $tmpHtmlFile) === false) {
+        @unlink($tmpFile);
```
jpwhite4, line 91: what happens if you try to `unlink(false)`

plessbd: nothing cause `@` hides the failure

plessbd:
```
Warning: unlink(): No such file or directory
```

libraries/charting.php
```diff
@@ -136,10 +141,12 @@ function getScreenFromChromium($file, $width, $height, $format){
     $err = stream_get_contents($pipes[2]);
     fclose($pipes[1]);
     fclose($pipes[2]);
-
     $return_value = proc_close($process);
 
+    @unlink($tmpHtmlFile);
+
     if ($return_value != 0) {
+        @unlink($outputFile);
```
jpwhite4, line 149: this is going to give variable does not exist warnings on the pdf and svg export pathways.

plessbd: Doesnt seem to as far as the docker is configured....

tested by putting `@unlink($IDONOTEXIST)` at the top of the export function, running `tail -f /var/log/httpd/* /var/log/xdmod/*` and exporting a png

plessbd: still added `$outputFile = null;` to make you happy.

PR #1430, opened 2020-08-24
Verifying User Query Count to address TrustedCI feedback

ryanrath:
## Description
This change is to address
https://app.asana.com/0/808093868887967/1188274656222915 . These problems were
brought to our attention via our 2019 engagement with TrustedCI. Specifically,
we now ensure that only a single user record is returned from these queries,
anything else is considered an exception.

## Tests performed
automated tests were performed

## Checklist:

### DIFF
--- a/classes/Rest/Utilities/Authentication.php
+++ b/classes/Rest/Utilities/Authentication.php
@@ -228,7 +228,7 @@ class Authentication
             $resolver_query_params
         );
 
-        if (count($user_check) > 0) {
+        if (count($user_check) === 1) {
             $last_active_time = self::getMicrotime();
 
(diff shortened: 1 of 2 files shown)


jtpalmer:
Just for future reference, every pull request that is intended to be included in the change log must have both a "Category" and non-"Category" label.

Also, if you use the "maintenance / code quality" that indicates that it should NOT be included in the change log.

Since you checked the "suitable for changelog" box I'm assuming that you do want this included.

We may need to include all these instructions in the pull request template since it is not obvious.

ryanrath:
Sounds good, I'll keep that in mind. Thanks!

PR #1439, opened 2020-09-04
Remove Google chrome from docker

plessbd:
Remove google-chrome and replace it with just chromium since chromium is a dependency of Open XDMoD this should make the docker image smaller.
As a side effect this also fixes up the test artifacts for now that the liberation font is no longer installed by google-chrome


tests/ci/Docker/Dockerfile.centos7:xdmod-base
```diff
@@ -2,16 +2,12 @@ FROM centos:7.8.2003
 ENV BRANCH=xdmod9.0
 LABEL description="Base image containing XDMoD required software."
 
-COPY assets/google-chrome.repo /etc/yum.repos.d
```
jtpalmer, line 5: Can `tests/ci/Docker/assets/google-chrome.repo` be removed now?

plessbd: yes

ryanrath (approved):

PR #1440, opened 2020-09-04
Make tests more robust

plessbd:
Was able to reliably make this test fail when trying to use chromium instead of chrome for the tests.
Since this could possibly show up in chrome as well, pushing this on its own.

### DIFF
--- a/tests/ui/test/specs/xdmod/reportGenerator.page.js
+++ b/tests/ui/test/specs/xdmod/reportGenerator.page.js
@@ -899,7 +899,12 @@ class ReportGenerator {
         browser.waitForVisible(this.selectors.myReports.toolbar.newBasedOnMenu());
         const reportCount = this.getMyReportsRows().length;
         if (!center) {
-            browser.click(this.selectors.myReports.toolbar.newBasedOnTemplate(templateName));
+            browser.waitUntilAnimEnd(this.selectors.myReports.toolbar.newBasedOnTemplate(templateName));
+            /**
+             * It seems extJS does not attach the events right away, pause to allow events to exist before clicking on the element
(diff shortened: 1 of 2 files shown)


tests/ui/test/specs/xdmod/reportGenerator.page.js
```diff
@@ -899,7 +899,12 @@ class ReportGenerator {
         browser.waitForVisible(this.selectors.myReports.toolbar.newBasedOnMenu());
         const reportCount = this.getMyReportsRows().length;
         if (!center) {
-            browser.click(this.selectors.myReports.toolbar.newBasedOnTemplate(templateName));
+            browser.waitUntilAnimEnd(this.selectors.myReports.toolbar.newBasedOnTemplate(templateName));
+            /**
+             * It seems extJS does not attach the events right away, pause to allow events to exist before clicking on the element
+             */
+            browser.pause(250);
+            browser.waitAndClick(this.selectors.myReports.toolbar.newBasedOnTemplate(templateName));
```
jtpalmer, line 907: This function `selectNewBasedOnTemplate` is only called after `clickNewBasedOn`, which already pauses (https://github.com/ubccr/xdmod/blob/xdmod9.5/tests/ui/test/specs/xdmod/reportGenerator.page.js#L875).

I'm just curious if you know why that isn't enough pausing?  Could waiting for the animation to end be added to that function instead?

plessbd: updated to wait for 750ms in that function, kept the waitAndClick on this selector just for precaution.

ryanrath (approved):

PR #1446, opened 2020-09-11
Fix institution loading during manual user creation

jtpalmer:
## Description

Updates the `TGUserDropDown` (used for the "Map To" field when creating users in the internal dashboard) to execute the cascade options after the field is set programmatically. (Previously the cascade options were only executed after the `select` event which is only triggered by a user selecting an item from the drop down.)

## Motivation and Context

Using the "Find" option during user creation wasn't setting the institution which then produces an error message when attempting to create the user.

https://app.asana.com/0/807770147051221/1193376020829900/f

## Tests performed

Tested manually.

## Checklist:

### DIFF
--- a/html/gui/js/TGUserDropDown.js
+++ b/html/gui/js/TGUserDropDown.js
@@ -43,18 +43,24 @@ CCR.xdmod.ui.TGUserDropDown = Ext.extend(Ext.form.ComboBox, {
     /**
      * Set the value and raw value.
      *
-     * @param {integer} v - The person ID.
-     * @param {string} l - The person display value.
-     * @param {boolean} c - Execute cascade function if true (defaults to true).
+     * @param {number} personId - A person ID.
+     * @param {string} personName - A person name.
+     * @param {boolean} updateCascadeComponent - Execute cascade function if
+     * true (defaults to true).
      */
-    initializeWithValue: function (v, l, c) {
-        var cascade = (typeof c !== 'undefined') ? c : true;
+    initializeWithValue: function (
+        personId,
+        personName,
+        updateCascadeComponent
+    ) {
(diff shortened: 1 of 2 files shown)


ryanrath (approved):
Looks good as soon as Shippable finishes passing.

html/gui/js/TGUserDropDown.js
```diff
@@ -40,9 +40,22 @@ CCR.xdmod.ui.TGUserDropDown = Ext.extend(Ext.form.ComboBox, {
...
-    initializeWithValue: function (v, l) {
-        this.setValue(v, l);
+    /**
+     * Set the value and raw value.
+     *
+     * @param {integer} v - The person ID.
+     * @param {string} l - The person display value.
+     * @param {boolean} c - Execute cascade function if true (defaults to true).
+     */
+    initializeWithValue: function (v, l, c) {
```
ryanrath, line 50: Yay for more docblocks! Personally, I think these one letter variables could use an re-name. Especially since your're adding a new one anyway.

ryanrath (changes requested):
Looks good minus function argument names.

html/gui/js/TGUserDropDown.js
jtpalmer: Let me know if you have any suggestions for the variable names.  I'm not really sure what the old ones were intended to mean.

ryanrath: sure, I mean you have the descriptions of the variables there so you could base 'um off of that. Maybe something like: 

```
initializeWithValue: function(personId, personDisplay, cascade);
```
or a little more "generic" 

```
initializeWithValue: function(value, displayValue, cascade); 
```

or something else entirely if you have other ideas. These were just off the top of my head.

ryanrath (approved):

PR #1448, opened 2020-09-16
Change raw statistics variable substitution format

jtpalmer:
## Description

Change the format used for variable substitution in `rawstatistics.json` configuration files.

This changes the syntax for variable substitution to match the format used by the ETL configuration files (as well as using the same PHP class).  It also adds the other PHP constants that are not current used to the list of substitutions.

I'll update the upgrade documentation later since there will may be more changes made to the format of these files.

## Motivation and Context

This is intended to make the substitutions more uniform with those used by the ETL and `datawarehouse.json` configuration file.

It is also a step working towards https://app.asana.com/0/342819846538629/1185671232325739/f

## Tests performed

Tested manually in the job viewer.

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Data/RawStatisticsConfiguration.php
+++ b/classes/DataWarehouse/Data/RawStatisticsConfiguration.php
@@ -3,6 +3,7 @@
 namespace DataWarehouse\Data;
 
 use Configuration\XdmodConfiguration;
+use ETL\VariableStore;
 use Exception;
 
 /**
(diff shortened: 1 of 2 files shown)


jpwhite4:
Will also need a corresponding change to the supremm stuff. Will be sufficient to switch out the:
```
fos_id: {
            name: "config://hierarchy.json:bottom_level_label",
```
in `etl/js/config/supremm/etl.schema.js` to:
```
fos_id: {
           name: "${HEIRARCHY...
```
or is it more complicated?

jpwhite4 (approved):

jtpalmer:
@jpwhite4 I'm hoping that's all that will need to change, but I haven't tested it yet.

PR #1455, opened 2020-09-30
Update SSO to pull the configured SP's IdP metadata directly

plessbd:
The previous code was bad.  It just got the "last" configured information.
This update uses the configured authsource to get its IdP and use that information to get the proper metadata.

This has been tested on metrics-dev by switching the configured authsource

### DIFF
--- a/classes/Authentication/SAML/XDSamlAuthentication.php
+++ b/classes/Authentication/SAML/XDSamlAuthentication.php
@@ -220,30 +220,31 @@ EML;
      */
     public function getLoginLink()
     {
-        if ($this->isSamlConfigured()) {
-            $idpAuth = \SimpleSAML_Metadata_MetaDataStorageHandler::getMetadataHandler()->getList();
-            $orgDisplay = "";
-            $icon = "";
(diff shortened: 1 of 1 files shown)


classes/Authentication/SAML/XDSamlAuthentication.php
```diff
@@ -220,30 +220,31 @@ public function getLoginURL($returnTo)
...
-            }
-            return array(
-                'organization' => $orgDisplay,
-                'icon' => $icon
-            );
-        } else {
+        if (!$this->isSamlConfigured()) {
             return false;
         }
+        $idp = \SimpleSAML_Metadata_MetaDataStorageHandler::getMetadataHandler()->getMetadata(
```
jpwhite4, line 226: What does getMetadata() return if the metadata is absent? Do you need an if statement to handle this case?

plessbd: null, so the if statements below will catch it as empty and handle the case.

ryanrath (approved):

PR #1456, opened 2020-10-06
Fix usage report access drill down logic

jtpalmer:
## Description

Fixes use of `\Models\Services\Acls::getQueryDescripters()`.  This function returns an array so `getDrillTargets` cannot be called on the returned value.

## Motivation and Context

This was generating errors during the report manager cron job:

```
2020-10-03 07:00:19 [critical] Call to a member function getDrillTargets() on a non-object
(file: /data/www/xdmod/share/classes/DataWarehouse/Access/Usage.php, line: 820, type: 1)
```

https://app.asana.com/0/159049597309611/1197571582623258/f

## Tests performed

Tested manually on xdmod-dev.

## Checklist:


plessbd (approved):

ryanrath (approved):

classes/DataWarehouse/Access/Usage.php
```diff
@@ -811,18 +811,33 @@ public function getCharts(XDUser $user, $chartsKey = 'data') {
 
                 // Check if the user's chart pool contains this chart.
                 $usageChartInPool = $chartPool->chartExistsInQueue($usageChartArgsStr);
-                $queryDescripter = Acls::getQueryDescripters(
+                $queryDescripters = Acls::getQueryDescripters(
```
jpwhite4, line 814: Under what circumstances does getQueryDescrtptors return an array or a queryDescriptor instance?

jtpalmer:
This doesn't fix the underlying issue since it doesn't explain how `$usageRealm` could ever not be set to a realm, which is the real problem.

PR #1458, opened 2020-10-12
Migrate Logging Infrastructure

ryanrath:
## Description
This PR swaps all Pear Log usage w/ Monolog. 
  - Added a new `CCR\Logging` class that takes the place of `CCR\Log` and utilizes Monolog. 
  - Maintained usage of general log settings in portal_settings.ini 
  - Removed usage of the custom `trace` level for now, but this can be re-added if we create a fork of Monolog as we did w/ Log. 
  - Updated / Added doc blocks to reflect using Monolog vs. Pear. 
  - A few minor updates to some of our tests as Monolog uppercases the level name. 
  - The one big difference between Pear / Monolog I found was that Monolog does not support non-string log messages out of the box. To resolve this I wrapped each of these instances w/ a `json_encode` but we have a number of other options when addressing this. We could stick w/ converting it pre-log call and use `print_r($message, true)` or a custom conversion function. If we choose to fork Monolog for a new Trace level then we could also update `Logger::addRecord` to utilize the same `extractMessage` function that Pear's Log class does. 

## Motivation and Context
This PR came about from looking at what would be required to update XDMoD to support integrating Symfony Flex which requires PHP7. After speaking with Jeff about his students work towards the Centos8 / PHP7 goal he pointed me to https://docs.google.com/document/d/1fIr5ZzTrYdfO3aWP_sNouC-iojNfePqn3axMapJVsu4 which documents some of the currently known blockers for migrating XDMoD to Centos8 / PHP7. I started experimenting w/ a local docker image based on Jeff's students work and noticed that Pear Log required MDB2 which is not compatible w/ PHP7. After a brief conversation he had mentioned that we should probably replace our Logging prior to moving to PHP7, which got me thinking about how hard it would be to actually do. Turns out that it wasn't hard at all, just tedious. I also have branches ready to go for AppKernels, XSEDE and SUPREMM. 

## Tests performed
All automated tests for OpenXDMoD fresh install & upgrade

## Checklist:


jpwhite4:
There are a large number of code changes that appear to just change CCR::Log::X to CCR::Logging::X. Why the namespace change?

background_scripts/batch_export_manager.php
```diff
@@ -80,15 +81,15 @@
         'mail' => false,
         'consoleLogLevel' => $logLevel
     );
-    $logger = Log::factory('batch-export', $logConf);
+    $logger = Logging::singleton('batch-export', array('console' => array(), 'mysql' => array()));
```
jpwhite4, line 84: Why not just write a Log::factory function that does what we need? Then you wouldn't need to change most of this code.

background_scripts/etl_aggregation_table_manager.php
```diff
@@ -154,7 +155,11 @@
 
 if ( NULL !== $scriptOptions['verbosity'] ) $conf['consoleLogLevel'] = $scriptOptions['verbosity'];
 
-$logger = Log::factory('etl_aggregation_table_manager', $conf);
+$logger = Logging::factory('etl_aggregation_table_manager', array(
```
jpwhite4, line 158: Why don't you make the Logging::factory function accept the `$conf` varialbe in its current form? Also why change the namespace?

classes/CCR/Logging.php
```diff
@@ -0,0 +1,164 @@
...
+     * @param array $config
+     *
+     * @return LoggerInterface
+     *
+     * @throws \Exception                If a missing directory is not buildable
+     * @throws \InvalidArgumentException If `name` is not a resource or string
+     */
+    public static function factory($name, array $config = array('null' => array()))
+    {
+        $logger = new \Monolog\Logger($name);
```
jpwhite4, line 37: Shouldn't this be static?

background_scripts/report_schedule_manager.php
```diff
@@ -61,15 +59,21 @@
...
+    'email' => array(
+        'subject' => 'Report Scheduler'. APPLICATION_ENV == 'dev' ? ' [Dev]' : ''
+    )
+));
 
 // =====================================================================
 
 // NOTE: "process_start_time" is needed for log summary.
-$logger->notice(array(
+$logger->notice(json_encode(array(
```
jpwhite4, line 73: Why didn't you write a wrapper function that calls the monolog logger with a string? This would significantly reduce the amount of code changes (and hence the 'attack' surface of mistakes creeping in) it would also make it significantly easier to review the code and would have been less work.

background_scripts/batch_export_manager.php
jpwhite4: Also this code change leaves an unused variable `$logConf` - I highly suggest making a Log::factory() that accepts the original log configuration. then this file would not need to change at all.

jpwhite4: Also the old code used `$logLevel` as the console log level. The new code seems to ignore the logLevel setting. Again this bug would not exist if you make a Log::factory() that accepts the original log configuration

classes/CCR/Logging.php
```diff
@@ -0,0 +1,164 @@
...
+
+        foreach ($config as $type => $typeConfig) {
+            $bubble = array_key_exists('bubble', $typeConfig) ? $typeConfig['bubble'] : true;
+
+            switch ($type) {
+                case 'console':
+                    $level = array_key_exists('level', $typeConfig)
+                        ? $typeConfig['level']
+                        : self::getLevel(\xd_utilities\getConfiguration('logger', 'default_level_console'), Logger::NOTICE);
```
jpwhite4, line 46: The old code defaulted to WARNING if the log level setting was absent from the configuration file:
```php
   protected static function getDefaultLogLevel($logType)
    {
        $option = 'default_level_' . $logType;

        try {
            $levelName = self::getConfiguration($option);
            $level = constant(get_class() . '::' . strtoupper($levelName));
        } catch (Exception $e) {
            $level = self::WARNING;
        }

        return $level;
    }
```

ryanrath:
This PR is being replaced by https://github.com/ubccr/xdmod/pull/1461

PR #1461, opened 2020-10-29
Migrate Logging: Redux

ryanrath:
## Description
This PR swaps all Pear Log usage w/ Monolog. 
  - Added a new `CCR\Logging` class that takes the place of `CCR\Log` and utilizes Monolog. 
  - Maintained usage of general log settings in portal_settings.ini 
  - Removed usage of the custom `trace` level for now, but this can be re-added if we create a fork of Monolog as we did w/ Log. 
  - Updated / Added doc blocks to reflect using Monolog vs. Pear. 
  - A few minor updates to some of our tests as Monolog uppercases the level name. 
  - The one big difference between Pear / Monolog I found was that Monolog does not support non-string log messages out of the box. To resolve this I created a new class `\CCR\Logger` that is utilized in `CCR\Log::factory` and `\CCR\Log::singleton`. This class provides a shim layer between our code that provides arrays as log messages & Monolog that will json_encode these messages before passing them on to the default Monolog handlers.

## Motivation and Context
This PR came about from looking at what would be required to update XDMoD to support integrating Symfony Flex which requires PHP7. After speaking with Jeff about his students work towards the Centos8 / PHP7 goal he pointed me to https://docs.google.com/document/d/1fIr5ZzTrYdfO3aWP_sNouC-iojNfePqn3axMapJVsu4 which documents some of the currently known blockers for migrating XDMoD to Centos8 / PHP7. I started experimenting w/ a local docker image based on Jeff's students work and noticed that Pear Log required MDB2 which is not compatible w/ PHP7. After a brief conversation he had mentioned that we should probably replace our Logging prior to moving to PHP7, which got me thinking about how hard it would be to actually do. Turns out that it wasn't hard at all, just tedious. After this PR is merged I have a branch ready to go w/ the basics of centos8 support for XDMoD. 

## Tests performed
All automated tests for OpenXDMoD fresh install & upgrade

## Checklist:


classes/DB/PDODBIngestor.php
```diff
@@ -97,7 +97,11 @@ public function ingest()
 
 
 	}
-	
+
+    public function setLogger(\Psr\Log\LoggerInterface $logger)
```
jpwhite4, line 101: I'm very wary of adding TODOs to the code. Please can you explain why this function has been added and what is the impact of having the empty implementation

ryanrath: Alrighty, so first off I agree about the TODOs, I should have caught that there was one added, sorry 'bout that. Here's the context for this particular class, It didn't actually fully implement the `Ingestor` interface as it did not provide an implementation for `setLogger`. Actually, now that I look at it, it doesn't actually have a `logger` member variable at all and all of it's "logging" is done w/ `print` and `echo` statements. 

So I could fix this up a couple of different ways, which would be your preference? 

- I could just remove the function. 
- I could add actual logging to the class by replacing all the `print` / `echo` w/ calls to the logger. 
- Something else entirely?

jpwhite4: I suggest that in a separate pull request we 'git rm' this file since it is not used. Then it will not even be here in this pull request and we need not worry about it.

ryanrath: I like it! I'll get that PR up post haste.

composer.json
```diff
@@ -28,9 +28,11 @@
         "symfony/process": "~2.0",
         "taq/pdooci": "^1.0",
         "tildeio/rsvpjs-min-file": "^3.0.18",
-        "ubccr/log": "1.13.2",
         "ubccr/simplesamlphp-module-authglobus": "^1.3",
-        "ubccr/simplesamlphp-module-authoidcoauth2": "^1.1"
+        "ubccr/simplesamlphp-module-authoidcoauth2": "^1.1",
+        "monolog/monolog": "^1.25",
+        "ext-pdo": "*",
```
jpwhite4, line 34: Remove for this pull request - we can add in later

tests/component/lib/ETL/IngestorTest.php
```diff
@@ -117,7 +117,9 @@ public function testHideSqlWarnings() {
 
     public function testHideSqlWarningCodes() {
         $result = $this->executeOverseer('xdmod.ingestor-tests.test-sql-warnings', '-o "hide_sql_warning_codes=1366"');
-
+        if ($result['exit_status'] !== 0) {
```
jpwhite4, line 120: Is debug code?

tests/component/lib/ETL/IngestorTest.php
```diff
@@ -141,13 +143,15 @@ public function testHideSqlWarningCodes() {
 
         // Run the same action, but filter all expected warning codes.
         $result = $this->executeOverseer('xdmod.ingestor-tests.test-sql-warnings', '-o "hide_sql_warning_codes=[1264,1366]"');
-
+        if ($result['exit_status'] !== 0) {
```
jpwhite4, line 146: Debug?

libraries/security.php
```diff
@@ -11,14 +11,16 @@
...
 {
-    $cParams = session_get_cookie_params();
-    session_set_cookie_params(
-        $cParams["lifetime"],
-        $cParams["path"],
-        $cParams['domain'],
-        true
-    );
-    @session_start();
+    if (session_status() !== PHP_SESSION_ACTIVE) {
```
jpwhite4, line 14: double check this change

classes/CCR/Loggable.php
```diff
@@ -95,7 +95,7 @@ public function logAndThrowException($message, array $options = null)
     {
         $logMessage = array();
         $message = "{$this}: " . ( is_string($message) ? $message : "" );
-        $logLevel = PEAR_LOG_ERR;
+        $logLevel = Log::DEBUG;
```
jpwhite4, line 98: The level switched from ERR to DEBUG. Is this correct?

jpwhite4:
Issugest that you implement a new MonoLog formatter class that outputs the log records in the identical way that they used to be.

This would include the strtolower(level_name). A cursory glance at the monolog source code suggests that you could also handle the array to string encoding in the formatter class too.

For an example of the strtolower see e.g. https://github.com/Seldaek/monolog/blob/9776fa1d548716490a7d49ffbb608e5c2dda589b/src/Monolog/Formatter/FluentdFormatter.php

jpwhite4:
So the database output uses a json encoding: ``` select * from log_table limit 10;
+------+---------------------+------------------+----------+---------------------------------------------------------------------------------+
| id   | logtime             | ident            | priority | message                                                                         |
+------+---------------------+------------------+----------+---------------------------------------------------------------------------------+
|    1 | 2020-08-14 18:49:13 | xdmod-import-csv |        5 | {"message":"xdmod-import-csv start","process_start_time":"2020-08-14 18:49:13"} |
|    2 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Importing type 'hierarchy'"}                                        |
|    3 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Starting database transaction"}                                     |
|    4 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Committing database transaction"}                                   |
|    5 | 2020-08-14 18:49:13 | xdmod-import-csv |        5 | {"message":"xdmod-import-csv end","process_end_time":"2020-08-14 18:49:13"}     |
|    6 | 2020-08-14 18:49:13 | xdmod-import-csv |        5 | {"message":"xdmod-import-csv start","process_start_time":"2020-08-14 18:49:13"} |
|    7 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Importing type 'group-to-hierarchy'"}                               |
|    8 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Starting database transaction"}                                     |
|    9 | 2020-08-14 18:49:13 | xdmod-import-csv |        6 | {"message":"Committing database transaction"}                                   |
|   10 | 2020-08-14 18:49:13 | xdmod-import-csv |        5 | {"message":"xdmod-import-csv end","process_end_time":"2020-08-14 18:49:13"}     |
+------+---------------------+------------------+----------+---------------------------------------------------------------------------------+
```

So you are going to have to come up with a way of making it use the json encoding with the DB handler and the CCRLine-style encoding with the Console handler.

One option could be to always serialize to json in the CCR::Logger::extractMessage() function and then have the CCRLineFormatter do a speculative json_decode() and then run the  `$parts[] = '(' . implode(', ', $nonMessageParts) . ')';`.

File: classes/DB/PDODBIngestor.php (hunk 1 above)
Comment (ryanrath) [reply classes/DB/PDODBIngestor.php]: This was taken care of in: https://github.com/ubccr/xdmod/pull/1462

File: classes/CCR/Loggable.php (hunk 6 above)
Comment (ryanrath) [reply classes/CCR/Loggable.php]: This is swapped back to `Log::ERR` in the latest version of the code.

File: classes/DB/ArrayIngestor.php [hunk 7]
```diff
@@ -4,6 +4,8 @@
  *
  */
 use CCR\DB\iDatabase;
+use CCR\Log;
```
jpwhite4, line 7: I don't think you need this use statement since CCR\Log is not present in the file?

ryanrath: I believe this is used on line 40 at the end of the `__construct` function.

composer.json
ryanrath: Just confirming that the `ext-*` entries have been removed from composer.json

libraries/security.php
ryanrath: Yup, this was removed as it was meant to be in the Centos8 branch.

jpwhite4 (changes requested):
The output format for the database logging is still incorrect:

image

classes/CCR/CCRDBHandler.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    protected function write(array $record)
+    {
+        $sql = sprintf("INSERT INTO %s.%s (id, logtime, ident, priority, message) VALUES(:id, NOW(), :ident, :priority, :message)", $this->schema, $this->table);
+
+        $this->db->execute($sql, array(
+            ':id' => $this->getNextId(),
+            ':ident' => $record['channel'],
+            ':priority' => $record['level'],
+            ':message' => $record['formatted']
```
jpwhite4, line 82: You don't want to output record['formatted'] as the message.

jpwhite4: you probably want record['message'] but I have not verified this in the actual code

ryanrath: Great catch, thanks! Yeah, I believe this was left over from moving the formatting around to different places within the code. I tested locally by spinning up a local docker w/ an upgrade. Ran xdmod-ingestor to generate some log messages and confirmed that they were **not** displayed correctly in the interface. Updated the field to `message` and repeated the docker upgrade test and confirmed that the log messages were correctly displayed. 

One thing I did note is that testing the `upgrade` path did not generate any log records, which in turn meant that when I clicked on the `Log Data` tab all that was displayed was a blank Tab Panel, as opposed to displaying the log interface, just without data, or an informational message indicating that things were working, just that there was no data to display.

ryanrath: Also, I've updated `CCRDBHandlerTest` so that it: 
  - Is constructing it's logger the same way that the rest of the code does. 
  - Actually inspects the format of the generated log records to make sure that we don't run into this again.

classes/CCR/Log.php
```diff
@@ -92,12 +98,15 @@ public static function factory(
...
 
             if ($e !== null && ($e['type'] & $mask) == 0) {
-                $logger->crit(array(
-                    'message' => $e['message'],
-                    'file'    => $e['file'],
-                    'line'    => $e['line'],
-                    'type'    => $e['type'],
-                ));
+                $logger->crit(
+                    sprintf(
```
jpwhite4, line 102: Is this sprintf correct here? Don't you want to send an array in so that it can be json_encoded properly?

classes/CCR/CCRDBHandler.php
```diff
@@ -0,0 +1,122 @@
...
+     * @see AbstractProcessingHandler::write()
+     */
+    protected function write(array $record)
+    {
+        $sql = sprintf("INSERT INTO %s.%s (id, logtime, ident, priority, message) VALUES(:id, NOW(), :ident, :priority, :message)", $this->schema, $this->table);
+
+        $this->db->execute($sql, array(
+            ':id' => $this->getNextId(),
+            ':ident' => $record['channel'],
+            ':priority' => $record['level'],
```
jpwhite4, line 81: Is this correct? The priority values in the old system were mapped to the numerical values of the PEAR log constants.
Old code ```
MariaDB [(none)]> SELECT DISTINCT priority FROM mod_logger.log_table;
+----------+
| priority |
+----------+
|        2 |
|        3 |
|        4 |
|        5 |
|        6 |
+----------+
```

this stuff:
```
SELECT DISTINCT priority FROM mod_logger.log_table;
+----------+
| priority |
+----------+
|      200 |
|      250 |
+----------+
2 rows in set (0.03 sec)
```

**Review by jpwhite4 — CHANGES_REQUESTED** [review]

File: classes/CCR/CCRDBHandler.php (hunk 10 above)
Comment (jpwhite4) [reply classes/CCR/CCRDBHandler.php:81]: Needless to say these values can't change because they are all coupled to the stuff in the dashboard.

File: html/internal_dashboard/controllers/log/get_levels.php [hunk 11]
```diff
@@ -1,25 +1,23 @@
...
         'response' => array(
-            array('id' => PEAR_LOG_EMERG,   'name' => 'Emergency'),
-            array('id' => PEAR_LOG_ALERT,   'name' => 'Alert'),
-            array('id' => PEAR_LOG_CRIT,    'name' => 'Critical'),
-            array('id' => PEAR_LOG_ERR,     'name' => 'Error'),
-            array('id' => PEAR_LOG_WARNING, 'name' => 'Warning'),
-            array('id' => PEAR_LOG_NOTICE,  'name' => 'Notice'),
-            array('id' => PEAR_LOG_INFO,    'name' => 'Info'),
-            array('id' => PEAR_LOG_DEBUG,   'name' => 'Debug'),
+            array('id' => \CCR\Log::EMERG,   'name' => 'Emergency'),
```
jpwhite4, line 13: This is no good because it you change the internal constants for the log levels then you will not be able to access the old log data.

classes/CCR/Log.php
ryanrath: I can change it back, this was added pre being able to handle arrays as messages.

classes/CCR/CCRDBHandler.php
ryanrath: Agreed, this wouldn't work for an existing installation. I figure we have a couple of options, let me know if you have a preference or a different idea. ( This applies to the comment for `get_levels.php` too ) 
  - I could setup a mapping table so that we can select records for the same level given two different priority values. 
  - Add a migration sql statement that updates the old pear level values to the new PSR\Log values.
    - **I personally like this one as it's less to manage going forward**

html/internal_dashboard/controllers/log/get_levels.php
ryanrath: Yup, see https://github.com/ubccr/xdmod/pull/1461#discussion_r535629178

classes/CCR/CCRDBHandler.php
plessbd: I would think the SQL statement would be better since like you said less to manage

jpwhite4: Or what about option 3)

don't use `$record['level']` when you insert into the database. Instead insert the numerical value that corresponds to the old one. The 'new' priority levels would never hit the database.

jpwhite4: especially since the numerical values you are using here are actually monolog values and not part of the PSR rfc.

jpwhite4: SELECT * FROM mod_logger.log_level;

plessbd: You know what they say about assumption, I thought they were part of the PSR rfc.

Option 3 sounds good then.

configuration/linker.php
```diff
@@ -121,16 +123,20 @@ function handle_uncaught_exception($exception)
...
     $logfile = LOG_DIR . "/" . xd_utilities\getConfiguration('general', 'exceptions_logfile');
 
-    $logConf = array('mode' => 0644);
-    $logger = \Log::factory('file', $logfile, 'exception', $logConf);
+    $logConf = array(
+        'mode' => 0644,
+        'file' => $logfile
+    );
+
+    $logger = Log::factory('file', $logConf);
```
jpwhite4, line 131: I may well be over thinking this. But are you allowed to use PSR loaded classes in the linker? After all, the linker itself it responsible for loading them.

Do you have a test case (whether manual or automated) that gets to this place in the code?

ryanrath:
@jpwhite4 @plessbd The latest commits implement the code levels from https://tools.ietf.org/html/rfc5424 . I also fixed a problem I had missed earlier that had console logs with duplicated info: 

- 2020-12-04 02:12:47 xdmod-import-csv [notice] 2020-12-04 02:12:47 xdmod-import-csv [notice] xdmod-import-csv start (process_start_time: 2020-12-04 02:30:47)

as oppossed to:
- 2020-12-04 02:12:47 xdmod-import-csv [notice] xdmod-import-csv start (process_start_time: 2020-12-04 02:30:47)

configuration/linker.php
ryanrath: Soooo, I wrote a test file located @ `/usr/share/xdmod/test.php` with the contents: 
```<?php

require_once 'configuration/linker.php';

$log = \CCR\Log::factory('testing123', array('db' => false, 'file' => false, 'email' => false));
$log->error('test');

throw new Exception('Testing Exception');
```

And here's what that produced: 

```
[root@baea16a23b9a xdmod]# php test.php
2020-12-04 02:12:03 testing123 [error] test
PHP Fatal error:  Uncaught exception 'Psr\Log\InvalidArgumentException' with message 'Level "3" is not defined, use one of: 100, 200, 250, 300, 400, 500, 550, 600' in /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php:516
Stack trace:
#0 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(298): Monolog\Logger::getLevelName(3)
#1 /usr/share/xdmod/classes/CCR/Logger.php(30): Monolog\Logger->addRecord(3, '{"message":"{\"...', Array)
#2 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(614): CCR\Logger->addRecord(3, '{"message":"Exc...', Array)
#3 /usr/share/xdmod/classes/CCR/Logger.php(22): Monolog\Logger->log(3, '{"message":"Exc...', Array)
#4 /usr/share/xdmod/configuration/linker.php(133): CCR\Logger->log(3, 'Exception Code:...')
#5 /usr/share/xdmod/configuration/linker.php(179): handle_uncaught_exception(Object(Exception))
#6 [internal function]: global_uncaught_exception_handler(Object(Exception))
#7 {main}
  thrown in /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php on line 516
2020-12-04 02:12:03 testing123 [critical] Uncaught exception 'Psr\Log\InvalidArgumentException' with message 'Level "3" is not defined, use one of: 100, 200, 250, 300, 400, 500, 550, 600' in /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php:516 Stack trace: #0 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(298): Monolog\Logger::getLevelName(3) #1 /usr/share/xdmod/classes/CCR/Logger.php(30): Monolog\Logger->addRecord(3, '{"message":"{\"...', Array) #2 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(614): CCR\Logger->addRecord(3, '{"message":"Exc...', Array) #3 /usr/share/xdmod/classes/CCR/Logger.php(22): Monolog\Logger->log(3, '{"message":"Exc...', Array) #4 /usr/share/xdmod/configuration/linker.php(133): CCR\Logger->log(3, 'Exception Code:...') #5 /usr/share/xdmod/configuration/linker.php(179): handle_uncaught_exception(Object(Exception)) #6 [internal function]: global_uncaught_exception_handler(Object(Exception)) #7 {main}   thrown (file: /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php, line: 516, type: 1)
2020-12-04 02:12:03 file [critical] Uncaught exception 'Psr\Log\InvalidArgumentException' with message 'Level "3" is not defined, use one of: 100, 200, 250, 300, 400, 500, 550, 600' in /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php:516 Stack trace: #0 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(298): Monolog\Logger::getLevelName(3) #1 /usr/share/xdmod/classes/CCR/Logger.php(30): Monolog\Logger->addRecord(3, '{"message":"{\"...', Array) #2 /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php(614): CCR\Logger->addRecord(3, '{"message":"Exc...', Array) #3 /usr/share/xdmod/classes/CCR/Logger.php(22): Monolog\Logger->log(3, '{"message":"Exc...', Array) #4 /usr/share/xdmod/configuration/linker.php(133): CCR\Logger->log(3, 'Exception Code:...') #5 /usr/share/xdmod/configuration/linker.php(179): handle_uncaught_exception(Object(Exception)) #6 [internal function]: global_uncaught_exception_handler(Object(Exception)) #7 {main}   thrown (file: /usr/share/xdmod/vendor/monolog/monolog/src/Monolog/Logger.php, line: 516, type: 1)
```

If I'm interpreting that correctly, which it's entirely possible that I'm not ( it is kinda late ), then it looks like it's ok to utilize the PSR loaded class, there's just a problem w/ how I implemented the log levels for RFC5424.

plessbd:
Ahh utc....

Timezones strike again...

classes/CCR/Log.php
```diff
@@ -278,6 +343,39 @@ protected static function getDefaultLogLevel($logType)
         return $level;
     }
 
+    public static function convertLevel($monologLevel)
```
jpwhite4, line 346: Please can you change this function name to better describe what it is doing. Something like monologLevelToCCRLevel() or monoLevelToRfc().....

jpwhite4: Also it is not clear why this function is in the class. The only place it is used is in the database handler. Wouldn't it make more sense to be a private member of that? Are there other places where we need to do this conversion?

jpwhite4: Also is is safe to throw and exception here? This code may be called in the top level exception handler.

ryanrath: - re: function name - lol, I guess great minds think a like because I've included this change in the latest commits.
- Good point: I included it in this class as I see `Log` as the central point of control for anything logging in XDMoD. It also seemed a general enough function that it might be needed elsewhere so I lumped it in w/ Log. I've also added a second conversion function `convertToMonologLevel` as I found that if you provide a log level that is less than Monolog's DEBUG, which is the case for all of the PSR log levels, then you'll just always be debugging. Another mark in the "Have these functions together in this class" column, at least in my head. 
- Good question... I'll try and write up a quick test case locally to see what happens / how it's handled.

ryanrath: @jpwhite4 By top level exception handler do you mean like in `linker.php::handle_uncaught_exception` ?

ryanrath: So if you were talking about `handle_uncaught_exception` in `linker.php` then here's the quick test I wrote (`test.php`): 

```php
<?php

require_once 'configuration/linker.php';

$config = array(
    'db' => false,
    'mail' => false,
    'file' => false,
    'consoleLogLevel' => -1
);

$logger = \CCR\Log::factory('testing', $config);

$logger->notice("Notice Me Senpai!");
```

Which generated the following output:
```
[root@40866e5dcf2f xdmod]# php test.php
2020-12-04 06:12:15 file [error] {"message":"Exception Code: 0"}
2020-12-04 06:12:15 file [error] {"message":"Message: Unknown Log Level"}
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"Unknown Log Level","code":0}
```

I also tried updating the logger config in `handle_uncaught_exception` to: 
```php
    $logConf = array(
        'mode' => 0644,
        'file' => $logfile,
        'consoleLogLevel' => -1
    );
```

Which generated the following:
```
[root@40866e5dcf2f xdmod]# php test.php
PHP Fatal error:  Uncaught exception 'Exception' with message 'Unknown Log Level' in /usr/share/xdmod/classes/CCR/Log.php:390
Stack trace:
#0 /usr/share/xdmod/classes/CCR/Log.php(201): CCR\Log::convertToMonologLevel(-1)
#1 [internal function]: CCR\Log::getConsoleHandler('file', Array)
#2 /usr/share/xdmod/classes/CCR/Log.php(170): call_user_func(Array, 'file', Array)
#3 /usr/share/xdmod/classes/CCR/Log.php(91): CCR\Log::getLogger('file', Array)
#4 /usr/share/xdmod/configuration/linker.php(132): CCR\Log::factory('file', Array)
#5 /usr/share/xdmod/configuration/linker.php(180): handle_uncaught_exception(Object(Exception))
#6 [internal function]: global_uncaught_exception_handler(Object(Exception))
#7 {main}
  thrown in /usr/share/xdmod/classes/CCR/Log.php on line 390
[root@40866e5dcf2f xdmod]#
```

So in thinking about it, we could provide a default level instead of throwing an exception. But then it could be possible that this "fails silently" when values we don't expect are provided as opposed to bringing attention to it. 

Thoughts?

classes/CCR/CCRDBHandler.php
ryanrath: @jpwhite4 I've gone with your suggestion of option 3. It's implemented in the latest version of the code.

classes/CCR/CCRDBHandler.php
```diff
@@ -0,0 +1,122 @@
...
+     * @throws Exception if any of the sql statements fail to execute.
+     */
+    protected function getNextId()
+    {
+        $this->db->beginTransaction();
+
+        $insertSQL = sprintf('INSERT INTO %s.log_id_seq (sequence) VALUES(NULL);', $this->schema);
+
+        $results = $this->db->execute($insertSQL);
+        if ($results <= 0) {
```
jpwhite4, line 101: The execute function returns true or false according to the php docs.

jpwhite4: https://www.php.net/manual/en/pdostatement.execute.php

classes/CCR/CCRDBHandler.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $results  = $this->db->query('SELECT LAST_INSERT_ID() as id');
+        $id = $results[0]['id'];
+
+        if (!is_numeric($id)) {
+            $this->db->rollBack();
+            throw new Exception('Retrieving latest log id value failed.');
+        }
+
+        // We don't check the number of rows deleted as while it most often will be 1, it is also expected to be zero
```
jpwhite4, line 114: Surely the reason why you don't check the number of rows deleted is because it doesn't matter. It could be any number.

classes/CCR/Log.php
```diff
@@ -278,6 +354,46 @@ protected static function getDefaultLogLevel($logType)
...
     }
 
+    /**
+     * Convert a \Monolog\Logger log level value to a CCR\Log log level.
+     *
+     * @param int $monologLevel the Monolog log level to be converted to a CCR log level.
+     * @return int the CCR log level that corresponds to the provided $monologLevel.
+     * @throws Exception if the provided $monologLevel is not found.
+     */
+    public static function convertToCCRLevel($monologLevel)
```
jpwhite4, line 364: Given that this function is called many many time.Surely it is much more efficient to not do an array flip on every call. Why not store the flipped array as a variable?

configuration/linker.php
jpwhite4: What were the contents of exception.log?

tests/component/lib/ETL/BaseEtlTest.php
```diff
@@ -74,7 +78,7 @@ protected function executeEtlAction(
         EtlConfiguration $etlConfig,
         EtlOverseerOptions $overseerOptions
     ) {
-        $action = aAction::factory($etlConfig, $actionName, self::$logger);
+        $action = aAction::factory($etlConfig, $actionName);
```
jpwhite4, line 81: Why the change here?

tests/component/lib/ETL/IngestorTest.php
```diff
@@ -86,8 +84,8 @@ public function testSqlWarnings() {
             }
         }
 
-        $this->assertGreaterThanOrEqual(4, $numWarnings, 'Expected number of SQL warnings');
-        $this->assertEquals('', $result['stderr'], "Std Error");
+        $this->assertGreaterThanOrEqual(1, $numWarnings, 'Expected number of SQL warnings');
```
jpwhite4, line 87: Why did the number of warnings change?

tests/unit/lib/OpenXdmod/Tests/HostListParserTest.php
```diff
@@ -23,7 +23,15 @@ public function setUp()
         $this->parser = new HostListParser();
 
         if (isset($GLOBALS['LOG_DEBUG']) && $GLOBALS['LOG_DEBUG']) {
-            $logger = \Log::factory('console', '', 'test', \CCR\Log::DEBUG);
+            $logger = \CCR\Log::factory(
+                'hostlistparser_test',
+                array(
+                    'consoleLogLevel' => \CCR\Log::DEBUG,
```
jpwhite4, line 29: Why the API change here?

jpwhite4 (changes requested):

classes/CCR/CCRDBHandler.php
ryanrath: True, but `$this-db` is an `iDatabase` /  `PDODB` which returns `$stmt->rowCount()` https://github.com/ubccr/xdmod/blob/4b05b1cddc7912541da65cdb96e25b9883d1554b/classes/CCR/DB/PDODB.php#L190

classes/CCR/CCRDBHandler.php
ryanrath: pretty much, I'll update the comment.

classes/CCR/Log.php
ryanrath: sure, I'll add the variable & update the function.

tests/component/lib/ETL/IngestorTest.php
ryanrath: That's actually a good question. I believe this changed while I was testing and the test never complained after changing things back because 4 >= 1. I'll make sure this is set back to the original 4.

classes/CCR/CCRDBHandler.php
jpwhite4: But that function throws an exception if the pdo::execute fails so there is no way that you can get to line 102 here.

tests/unit/lib/OpenXdmod/Tests/HostListParserTest.php
ryanrath: So this is an interesting one, initially this was just another artifact of earlier testing. I had wanted to be explicit that this was to just be a console logger. Changing the log options back to the original ( or equivalent to the original w/ out the db, mail, or file => false ) doesn't cause any problems. Which got me to thinking about whether or not this code was actually ever used. I went about trying to trigger it but ran into problems since there was no `LOG_DEBUG` in `$GLOBALS`. Nor was it immediately apparent how to set a value in GLOBALS. So, do we know why they used GLOBALS here?  I did a git blame but this code has been in place since we initially imported it into git. If we want to use env variables instead of GLOBALS then I tried this: 

```php
        $logDebug = getenv('LOG_DEBUG');
        if ($logDebug) {
            $logger = \CCR\Log::factory(
                'hostlistparser_test',
                array(
                    'consoleLogLevel' => \CCR\Log::DEBUG
                )
            );
            $this->parser->setLogger($logger);
        }
```

which actually worked when I specified `LOG_DEBUG=1`. 

Thoughts?

classes/CCR/Log.php
jpwhite4: What were the contents of the execption.log file when you did your test?

classes/CCR/CCRDBHandler.php
ryanrath: True, the insert statement would have to somehow succeed but not affect any rows. Which for a straight `INSERT ... VALUES` statement I don't think is possible. It might be different if it was a `INSERT ... SELECT ` but it's not, so I'll go ahead and remove the `if` statement after this `execute`.

classes/CCR/Log.php
ryanrath: With a valid logger in `linker.php`  it was: 

```
[root@2da6bfc3594c xdmod]# cat /var/log/xdmod/exceptions.log
2020-12-22 08:12:52 file [error] {"message":"Exception Code: 0"}
2020-12-22 08:12:52 file [error] {"message":"Message: Unknown Log Level"}
```

With the logger config updated w/ the invalid log level: 

```php
    $logConf = array(
        'mode' => 0644,
        'file' => $logfile,
        'consoleLogLevel' => -1
    );
```

`exceptions.log` is empty: 
```
[root@2da6bfc3594c xdmod]# php test.php
PHP Fatal error:  Uncaught exception 'Exception' with message 'Unknown Log Level' in /usr/share/xdmod/classes/CCR/Log.php:395
Stack trace:
#0 /usr/share/xdmod/classes/CCR/Log.php(223): CCR\Log::convertToMonologLevel(-1)
#1 [internal function]: CCR\Log::getConsoleHandler('file', Array)
#2 /usr/share/xdmod/classes/CCR/Log.php(192): call_user_func(Array, 'file', Array)
#3 /usr/share/xdmod/classes/CCR/Log.php(113): CCR\Log::getLogger('file', Array)
#4 /usr/share/xdmod/configuration/linker.php(132): CCR\Log::factory('file', Array)
#5 /usr/share/xdmod/configuration/linker.php(180): handle_uncaught_exception(Object(Exception))
#6 [internal function]: global_uncaught_exception_handler(Object(Exception))
#7 {main}
  thrown in /usr/share/xdmod/classes/CCR/Log.php on line 395
[root@2da6bfc3594c xdmod]# ls -alh /var/log/xdmod/
total 20K
drwxrwx--- 1 apache xdmod 4.0K Dec 22 19:48 .
drwxr-xr-x 1 root   root  4.0K Aug 14 18:49 ..
-rw-r--r-- 1 root   root   648 Dec 22 19:50 apache-error.log
-rw-rw---- 1 apache xdmod    0 Dec 22 19:50 exceptions.log
-rw-rw---- 1 apache xdmod    0 Dec 22 19:50 query.log
```

jpwhite4:
Test code:```
<?php

require_once '/usr/share/xdmod/configuration/linker.php';

throw new Exception("This message should be written to /var/log/xdmod/exceptions.log");
```

original code output:
```
Dec 22 21:05:36 exception [error] Exception Code: 0
Dec 22 21:05:36 exception [error] Message: This message should be written to /var/log/xdmod/exceptions.log
Dec 22 21:05:36 exception [info] Origin: /tmp/test.php (line 5)
Dec 22 21:05:36 exception [info] Trace:
#0 {main}
-------------------------------------------------------
```

current output:
```
2020-12-22 09:12:23 file [error] {"message":"Exception Code: 0"}
2020-12-22 09:12:23 file [error] {"message":"Message: This message should be written to \/usr\/share\/xdmod\/logs\/exception.log"}
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"This message should be written to \/usr\/share\/xdmod\/logs\/exception.log","code":0}
```

File: tests/unit/lib/OpenXdmod/Tests/HostListParserTest.php (hunk 19 above)
Comment (ryanrath) [reply tests/unit/lib/OpenXdmod/Tests/HostListParserTest.php:29]: @jpwhite4 So I was playing around with this some more and I got the `GLOBALS` to work with some combination of the following in either `constants.php` or `HostListParserTest`

```php
global $LOG_DEBUG;
$LOG_DEBUG=true;
```

So my question would be do we want to continue using `$GLOBALS` or do we want to use env variables?

ryanrath:
@jpwhte4 Alrighty, so in regards to: https://github.com/ubccr/xdmod/pull/1461#issuecomment-749776407 

Here are the issues I found: 

  1. Log messages presented in JSON
  2. Missing `INFO` level messages
  3. Additional JSON encoded data is displayed currently where none was previously.

Here's what I've been able to resolve so far: 

  1. Adjusted the way the messages were being provided to the logger by wrapping them in an array `array( 'message' => <original log message> )`. This ensured that the messages were no longer displayed in JSON. 
  2.  Added a `'consoleLogLevel' => \CCR\Log::INFO` property to the log conf. This ensures we see the `INFO` level log messages that were previously absent. 
  3. Modified the `global_uncaught_exception_handler` so that it only `echo $exceptionOutput['content'];` if it's operating within a server context. 
  
**Old**
```php
function global_uncaught_exception_handler($exception)
{
    // Perform logging and output building for the exception.
    $exceptionOutput = handle_uncaught_exception($exception);

    // If running in a server context...
    if ($exceptionOutput['isServerContext'] && !headers_sent()) {
        // Set the exception's headers (if any).
        foreach ($exceptionOutput['headers'] as $headerKey => $headerValue) {
            header("$headerKey: $headerValue");
        }

        // Set the status code header.
        $httpCode = $exceptionOutput['httpCode'];
        header("{$_SERVER['SERVER_PROTOCOL']} $httpCode ".HttpCodeMessages::$messages[$httpCode]);
    }

    // Print the exception's content.
    echo $exceptionOutput['content'];
} // global_uncaught_exception_handler
```

**New**

```php
function global_uncaught_exception_handler($exception)
{
   // Perform logging and output building for the exception.
    $exceptionOutput = handle_uncaught_exception($exception);

   // If running in a server context...
    if ($exceptionOutput['isServerContext']) {
        if (!headers_sent()) {
            // Set the exception's headers (if any).
            foreach ($exceptionOutput['headers'] as $headerKey => $headerValue) {
                header("$headerKey: $headerValue");
            }

            // Set the status code header.
            $httpCode = $exceptionOutput['httpCode'];
            header("{$_SERVER['SERVER_PROTOCOL']} $httpCode ".HttpCodeMessages::$messages[$httpCode]);
        }

        echo $exceptionOutput['content'];
    }
} // global_uncaught_exception_handler
```

Output w/ the changes I made:
```
[root@9aad8718f5bd xdmod]# php /usr/share/xdmod/test.php
2020-12-23 07:12:10 file [error] Exception Code: 0
2020-12-23 07:12:10 file [error] Message: This message should be written to /var/log/xdmod/exceptions.log
2020-12-23 07:12:10 file [info] Origin: /usr/share/xdmod/test.php (line 5)
2020-12-23 07:12:10 file [info] Trace:
#0 {main}
-------------------------------------------------------
[root@9aad8718f5bd xdmod]#
```

jpwhite4:
The output is still changed from the old code:
old:
```
2020-12-23 15:05:15 [notice] Start processing section 'xdmod.xdw-aggregate-storage'
```

new:
```
2021-01-04 08:01:34 xdmod-ingestor [notice] Start processing section 'xdmod.xdw-ingest-storage'
```

ryanrath:
@jpwhite4 re: https://github.com/ubccr/xdmod/pull/1461#issuecomment-754205216 Do we want to exclude the information about what script / logger is generating the log message? It's not a hard change, but it strikes me that this might be useful information to have.

@jpwhite4 Ok, just to double check, did https://github.com/ubccr/xdmod/pull/1461#issuecomment-750449657 address your previous concern about the output from unhandled exceptions?

jpwhite4:
The file and console log format output is incorrect. Also the datestamp is also wrong:
old code:
```
2021-01-04 22:06:36 [debug] DataWarehouse\Query\AggregateQuery(Jobs, groupbys=(jobsize), statistics=(total_cpu_hours)) getCountQueryString()
```

new code:
```
2021-01-04 10:01:07 datawarehouse.query [debug] DataWarehouse\Query\AggregateQuery(Jobs, groupbys=(jobsize), statistics=(total_cpu_hours)) getCountQueryString()
```

the two problems are (1) the new code puts the string 'datawarehouse.query' into the output. and (2) the hours, minutss and seconds in the timestamp are wrong.

jpwhite4:
> @jpwhite4 re: [#1461 (comment)](https://github.com/ubccr/xdmod/pull/1461#issuecomment-754205216) Do we want to exclude the information about what script / logger is generating the log message? It's not a hard change, but it strikes me that this might be useful information to have.
> 
> @jpwhite4 Ok, just to double check, did [#1461 (comment)](https://github.com/ubccr/xdmod/pull/1461#issuecomment-750449657) address your previous concern about the output from unhandled exceptions?

This pull request should not change the log output.

If you want to change the log output format then do it in a different pull request. If you do want to change the log output format then you will have to do a risk analysis to check whether you will break some other processes / components. One example of breakage is anything that currently parses the log output

jpwhite4:
Your problem is in:
```php
const TIME_FORMAT = 'Y-m-d h:m:s';
```
lowercase h is hours in 12 hour format and lower case m is months.

ryanrath:
@jpwhite4 Thanks for this catch: https://github.com/ubccr/xdmod/pull/1461#issuecomment-754260331 The latest commit: https://github.com/ubccr/xdmod/pull/1461/commits/bceaf7aef0b6e10946ae94d3f7ce7190ae6cef07 contains the changes to resolve those format discrepancies.

jpwhite4:
The output from the test.php is still different:
Original code:
```
php test.php
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"This message should be written to \/var\/log\/xdmod\/exceptions.log","code":0}

[root@ea97b64d1520 xdmod]# cat /var/log/xdmod/exceptions.log
Jan 06 17:10:11 exception [error] Exception Code: 0
Jan 06 17:10:11 exception [error] Message: This message should be written to /var/log/xdmod/exceptions.log
Jan 06 17:10:11 exception [info] Origin: /usr/share/xdmod/test.php (line 5)
Jan 06 17:10:11 exception [info] Trace:
#0 {main}
-------------------------------------------------------
```

current build:
```
php test.php
2021-01-06 17:12:11 [error] Exception Code: 0
2021-01-06 17:12:11 [error] Message: This message should be written to /var/log/xdmod/exceptions.log
2021-01-06 17:12:11 [info] Origin: /var/log/xdmod/test.php (line 5)
2021-01-06 17:12:11 [info] Trace:
#0 {main}
-------------------------------------------------------
[root@49402b35fe4a xdmod]# cat /var/log/xdmod/exceptions.log
2021-01-06 17:09:01 [error] Exception Code: 0
2021-01-06 17:09:01 [error] Message: This message should be written to /var/log/xdmod/exceptions.log
```

configuration/linker.php
```diff
@@ -173,19 +180,21 @@ function global_uncaught_exception_handler($exception)
     $exceptionOutput = handle_uncaught_exception($exception);
 
    // If running in a server context...
-    if ($exceptionOutput['isServerContext'] && !headers_sent()) {
-        // Set the exception's headers (if any).
-        foreach ($exceptionOutput['headers'] as $headerKey => $headerValue) {
-            header("$headerKey: $headerValue");
+    if ($exceptionOutput['isServerContext']) {
```
jpwhite4, line 183: Please explain why this code changed.

configuration/linker.php
```diff
@@ -121,16 +123,21 @@ function handle_uncaught_exception($exception)
 {
     $logfile = LOG_DIR . "/" . xd_utilities\getConfiguration('general', 'exceptions_logfile');
 
-    $logConf = array('mode' => 0644);
-    $logger = \Log::factory('file', $logfile, 'exception', $logConf);
+    $logConf = array(
+        'mode' => 0644,
+        'file' => $logfile,
+        'consoleLogLevel' => CCR\Log::INFO
```
jpwhite4, line 129: The old code created a file logger here. But this change seems to create a file logger and a console logger and set the console logger to log at info. I don't think this is the correct thing to do since it will likely break the frontend.

configuration/linker.php
ryanrath: I implemented this change to address this comment: https://github.com/ubccr/xdmod/pull/1461#issuecomment-749776407

Specifically it was so that this line was not displayed: 
```
2020-12-22 09:12:23 file [error] {"message":"Message: This message should be written to \/usr\/share\/xdmod\/logs\/exception.log"}
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"This message should be written to \/usr\/share\/xdmod\/logs\/exception.log","code":0}
```

configuration/linker.php
jpwhite4: Also this code is sending emails too. Which the old code did not.

tests/component/lib/ETL/BaseEtlTest.php
ryanrath: No good reason. I believe my thought process was something along the lines of `self::$logger` is never instantiated ( via `self::createLogger` ) so why provide it? Which should have been to obviously facilitate the debugging of tests. I've re-added it in https://github.com/ubccr/xdmod/pull/1461/commits/f8fe8c39c407cd242b2b6f21d3fc35dd6d20b127

configuration/linker.php
ryanrath: Yup, per our earlier convo I updated the logger config here: https://github.com/ubccr/xdmod/pull/1461/commits/29de561878156e3a2b46bba55f52246fbac0a124

eiffel777 (approved):

tests/unit/lib/OpenXdmod/Tests/HostListParserTest.php
jpwhite4: The code in question was added in:

commit e970d8f70c60c40dcdcf0388177741ca013fad04
Author: Jeffrey T. Palmer <jtpalmer@buffalo.edu>
Date:   Tue Sep 15 10:03:49 2015 -0400

    Add compressed host list parser

I assume that it was an artifact of development, but its probably worth checking with Jeff just in case.

jpwhite4 (approved):

jpwhite4:
Please can you squash and merge when you merge since there are lots of commits.

ryanrath:
@jpwhite4 Absolutely, thanks!

PR #1463, opened 2020-11-03
A dirty dirty attempt at migrating our Travis tests

ryanrath:
~**NOTE: This is super rough and I'm just playing around with getting things to work. Large amount of re-org / cleanup incoming afterwards.**~

**NOTE: This PR https://github.com/ubccr/xdmod-qa/pull/14 must be merged before this one. After it's merged then the `XDMOD_QA_BRANCH` default value can be updated back to `v1`. and the `GIT_URL` can be set to ubccr's.**

## Description
**:construction sign:** Please excuse the messs **:construction sign:**
**\<blink\>** Migrating Tests In Progress... **\</blink\>**

## Motivation and Context
TravisCI has updated their usage plans such that it's not realistic for us to continue utilizing the service. The changes introduced in this PR along with https://github.com/ubccr/xdmod-qa/pull/13 migrates the tests we were previously running in Travis to Shippable. 

## Tests performed
These changes were tested as much as possible locally before making sure that they also passed the required PR tests. 

## Checklist:


jpwhite4:
One thought is that we don't need to run the code style tests in both the fresh_install and upgrade builds, so it will save a (small) amount of time to just run in one of them.

shippable.yml
```diff
@@ -3,6 +3,11 @@ env:
     global:
         - COMPOSER_ALLOW_SUPERUSER=1
         - XDMOD_REALMS='jobs,storage,cloud'
+        - NODE_VERSION=6
+        - XDMOD_TEST_ARTIFACTS_MIRROR="$HOME/xdmod-test-artifacts.git"
```
plessbd, line 7: odd that this is still "needed" since it has not been used in years...

shippable.yml
```diff
@@ -3,6 +3,11 @@ env:
...
+        - XDMOD_IS_CORE=true
+        - XDMOD_SOURCE_DIR=$SHIPPABLE_BUILD_DIR
+        - XDMOD_INSTALL_DIR=/usr/share/xdmod
```
plessbd, line 10: we might not be able to do this with the rpm install and source install using the same directory.  Or if we do we might need to clean things up...

plessbd:
I would think we just want the style tests on the upgrade since the upgrade tests go a tiny bit faster

plessbd:
I thought of some things to keep in mind in no paticular order and of course after we get things working anyway...
TODOS:
Convert travis to multiple files so that they can be "folded" by shippable
Make travis tests put test artifacts into shippable directory
Add nvm install and noode install via npm to docker build (unless we can figure out how to not use it...)
Make sure nvm install doesnt mess with supremm node install....

shippable.yml
ryanrath: No idea, I literally just copied the all the environment variables from travis.yml over to make sure I wasn't missing anything.

ryanrath:
@plessbd Alrighty so re: your TODO's: 
- Not sure I understand the "convert travis to multiple files..." part? The folding I get though. 
- I'm not sure if these tests are natively able to output junit type test artifacts. Although we could maybe generate our own?
- "Add nvm install... " this is done w/ the 0.4 version of our docker image. 
- Hrmm, would this be in the job performance docker images? Or maybe the hpcts? Certainly something we should look at though.

plessbd:
I dont think shiibabble supports just being able to fold the code like travis did, which is why i mentioned turning it into multiple scripts

I know phpunit supports junit output, and eslint also supports it https://eslint.org/docs/user-guide/formatters/#junit

yes it would be part of job performance, you could probabbly use the hpctoolset one as well, but I would just use the job performance as it is much smaller to test with.

ryanrath:
- Gotcha, it could certainly be done obviously, but is it really necessary? 
- Sweet, I've updated the `migrate_travis` branch of xdmod-qa so that both the phpunit & eslint tests now output junit test artifacts to the `shippable/testresults` directory. 
- Coolio, I'll do some local testing w/ the job performance image and letcha know how it goes.

plessbd:
It just makes it "easier" to read, and not a priority getting it to work is the priority
Awesome work.

tests/ci/scripts/qa-test-setup.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Only run if we're doing an upgrade
+if [[ "$XDMOD_TEST_MODE" == "upgrade" ]]; then
+
+    # Check if an XDMOD_QA_BRANCH env variable has been set, if not then set a default.
+    if [[ -z "$XDMOD_QA_BRANCH" ]]; then
```
jpwhite4, line 9: Not that is matters, but you can use bash variable expansion to set a default value for a variable:
```bash
XDMOD_QA_BRANCH=${XDMOD_QA_BRANCH:-v1}
```

plessbd (dismissed):

plessbd (approved):

plessbd (approved):

PR #1465, opened 2020-11-19
Improve configurability for the stats defined in the etlv1 schema.

jpwhite4:
This adds support for:
1) using an include directive (or any valid config directive) in the documentation
   section for a statistic.
2) explicitly specify timeseries and aggregate queries.

Support for the `:timeseries` macro is removed. This was used in two
places in the schema definition used in the Job Performance realm. There
is a pull request in that repo that updates the schema def.


plessbd:
Is there a pr that actually uses these changes?

plessbd (approved):

jpwhite4:
For the corresponding change in the job performance module see:  https://github.com/ubccr/xdmod-supremm/pull/266

PR #1473, opened 2020-12-23
Add cloud instance viewer

eiffel777:
This PR adds functionality to track both active and inactive sessions and adds to the Job Viewer the ability to view details about a particular instance including all of the sessions of a VM on a timeline showing when they are active and inactive. Data for the Cloud realm should be re-ingested since we are tracking some new events.

The tests have changed to include the cloud realm in any test that checks to see what raw data realms exist. The regression tests have changed because stop events for an active session are no longer the `.end` event such as `compute.instance.power_off.end` but are now the `.start` events like `compute.instance.power_off.start`. The time difference between these events in the test data is a few seconds slightly changing the core hours and wall hours. 

## Motivation and Context
It is useful to see the lifecycle of a VM and also let's us track data about inactive sessions in the future.

## Tests performed
Tested locally and on metrics-dev.

## Checklist:


classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -114,15 +181,30 @@ protected function transform(array $srcRecord, &$orderId)
...
         if (($this->_instance_state['instance_id'] !== $srcRecord['instance_id']) || ($this->_instance_state['resource_id'] !== $srcRecord['resource_id'])) {
             $transformedRecord[] = $this->_instance_state;
-            if (!in_array($srcRecord['event_type_id'], $this->_stop_event_ids) && $srcRecord['event_type_id'] != self::START_ERROR) {
+            if ($srcRecord['event_type_id'] != self::TERMINATE && $srcRecord['event_type_id'] != self::START_ERROR) {
                 $this->initInstance($srcRecord);
             }
-        } elseif (in_array($srcRecord['event_type_id'], $this->_start_event_ids)) {
-            $this->updateInstance($srcRecord);
-        } elseif (in_array($srcRecord['event_type_id'], $this->_stop_event_ids)) {
+        } elseif (in_array($srcRecord['event_type_id'], [1,16])) {
```
jpwhite4, line 187: The hardcoded 1,16 seems out of place here since you have gone to the trouble of making the event ids php constants and then putting them in class member variables.

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -56,17 +69,71 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
...
+          self::STOP,
+          self::REQUEST_TERMINATE,
+          self::REQUEST_RESUME,
+          self::RESUME,
+          self::STATE_REPORT,
+          self::SUSPEND,
+          self::SHELVE,
+          self::UNSHELVE,
+          self::POWER_OFF_START,
+          self::POWER_OFF,
```
jpwhite4, line 85: I think this code is going to need some explanatory comments. How is POWER_OFF a start event? How come the start_event_ids and end_event_ids are exactly the same except for TERMINATE is not a start event?

classes/ETL/Ingestor/CloudStateReconstructorTransformIngestor.php
```diff
@@ -56,17 +69,71 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Log
...
+        $this->_all_event_ids = array_unique(array_merge($this->_start_event_ids, $this->_end_event_ids));
         $this->_end_time = $etlConfig->getVariableStore()->endDate ? date('Y-m-d H:i:s', strtotime($etlConfig->getVariableStore()->endDate)) : null;
 
         $this->resetInstance();
     }
 
     private function initInstance($srcRecord)
     {
-        $default_end_time = isset($this->_end_time) ? $this->_end_time : $srcRecord['event_time_ts'];
+        $beginOfDay = strtotime("today", $srcRecord['event_time_ts']);
```
jpwhite4, line 135: Is this 'timezone' safe?

tests/artifacts/xdmod/component/export/realm_manager/output/realms-for-user.json
```diff
@@ -1,11 +1,15 @@
 {
     "Normal user": [
         "usr",
-        [
-            {
+        {
+            "0": {
```
jsperhac, line 5: Note that recently merged PR #1481 re-indexes the arrays returned from getRealms() so that explicit indices are not needed here (previously there were potentially gaps in the array keys returned). This also applies to the realms.json testing artifact.

html/gui/js/modules/job_viewer/VMStateChartPanel.js
```diff
@@ -0,0 +1,192 @@
...
+
+        return chartText;
+    },
+    listeners: {
+        activate: function () {
+            Ext.History.add(this.historyToken);
+        },
+        load_record: function (panel, record) {
+            var self = this;
+            var active_states = [2, 8, 16, 20, 57, 59, 61];
```
ryanrath, line 51: So I'm wondering if it might be worth it to have a this data reside within a config file so that the data could be shared between the front & back end without having to manually make code changes? This obviously works but just a thought to maybe make any future changes easier. Another thing to keep in mind is that we're going to end up having to support multiple cloud's and each one is going to have different states etc. Maybe that's not something we worry about now, but it's something that we will have to address.

ryanrath (commented):
Alrighty, here's my first go through. Ultimately it looks good, most of things I found were with an eye to make our future selves hate our past selves a little less. The only thing of any substance that I found / thought of is the whole having to ultimately support multiple Cloud infrastructures thing that I think is more of a point of discussion and definitely not a show stopper.

ryanrath (approved):

PR #1475, opened 2021-01-04
Updating QA Setup Script to not explicitly set remote branch for build.sh

ryanrath:
## Description
I've updated the `qa-test-setup` script  to not explicitly set a remote branch for `build.sh`. This was causing problems because it was using the `BASE_BRANCH` env variable which in certain situations would be empty. This ultimately had `build.sh` erroring out w/ the message: `Could not resolve remote branch: `. 

## Motivation and Context
It's a good thing to have our QA tests actually run.

## Tests performed
I've tested this change locally w/ both a base XDMoD & SUPREMM build using the last merged PR's for both as the target. Both builds complete successfully.

## Checklist:

### DIFF
--- a/tests/ci/scripts/qa-test-setup.sh
+++ b/tests/ci/scripts/qa-test-setup.sh
@@ -10,9 +10,9 @@ if [[ "$XDMOD_TEST_MODE" == "upgrade" ]]; then
     QA_GIT_URL=${QA_GIT_URL:-https://github.com/ubccr/xdmod-qa.git}
 
 
-    # Check if XDMOD_SOURCE_DIR env variable exists, if not then we can't continue.
+    # Check if SHIPPABLE_BUILD_DIR env variable exists, if not then we can't continue.
     if [[ -z "$SHIPPABLE_BUILD_DIR" ]]; then
-        echo "XDMOD_SOURCE_DIR must be set before running this script."
+        echo "The SHIPPABLE_BUILD_DIR env variable must be set before running this script."
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


tests/ci/scripts/qa-test-setup.sh
```diff
@@ -24,10 +24,10 @@ if [[ "$XDMOD_TEST_MODE" == "upgrade" ]]; then
     # Setup the xdmod-qa environment / requirements.
     $HOME/.qa/scripts/install.sh
 
-    # If we're running on Shippable then make sure to include the
-    # base branch that we're merging into.
+    # If we're running on Shippable and a base branch is present then explicitly specify
+    # it as the remote branch for build.sh.
```
jpwhite4, line 28: Under what circumstances would you be running on shippable and a base branch would not be present? What would happen to the build in this case?

ryanrath: These seem to be the conditions:
https://app.shippable.com/github/ubccr/xdmod/runs/6422/2/scripts
https://app.shippable.com/github/ubccr/xdmod-supremm/runs/1586/2/scripts

*Expand `reports` and search for `BASE_BRANCH`*

And it looks like the build fails.

jtpalmer: When is the base branch ever set?  I don't see it anywhere else.

ryanrath: It's a env variable that Shippable provides, sometimes.

ryanrath:
Just submitted a PR https://github.com/ubccr/xdmod-qa/pull/19 that should resolve the problems w/ this PR.

jpwhite4 (approved):

PR #1480, opened 2021-01-07
Make cloud realm compatible with ONLY_FULL_GROUP_BY

eiffel777:
This changes makes changes so the cloud realm is compatible with MySQL when ONLY_FULL_GROUP_BY mode is enabled.

For the cloud aggregation the entails adding columns to the group by for the aggregation query. For the cloud resource specifications aggregation it revealed a bug when summing the core time for a time period because of a misplaced parenthesis. Re-aggregation of the cloud resource specifications is necessary because of this.

## Checklist:


configuration/etl/etl_action_defs.d/cloud_common/cloud_resource_specs_aggregation.json
```diff
@@ -27,7 +27,7 @@
                 "year": "${:YEAR_VALUE}",
                 "${AGGREGATION_UNIT}": "${:PERIOD_VALUE}",
                 "resource_id": "crs.resource_id",
-                "core_time_available": "COALESCE(SUM(${wallduration}) * crs.vcpus, 0)"
+                "core_time_available": "COALESCE(SUM(${wallduration} * crs.vcpus), 0)"
```
jpwhite4, line 30: Is this correct? Shouldn't the crs.vpus be in the statistic?

eiffel777:
@jpwhite4 @ryanrath This is updated and ready to be reviewed

classes/OpenXdmod/Migration/Version900To950/DatabaseMigration.php
```diff
@@ -0,0 +1,33 @@
+<?php
+namespace OpenXdmod\Migration\Version851To900;
```
jpwhite4, line 2: Wrong namespace - also this change conflicts with the other cloud resources pull request.

jpwhite4 (changes requested):

ryanrath (commented):
I'll admit this is nit-picky but I'd prefer to have the `use` statements that aren't being used removed. Which I believe would just leave the `CCR\DB` and `ETL\Utilities` statements.

jpwhite4 (approved):

PR #1483, opened 2021-01-14
Add data endpoint that can parse webserver logs.

jpwhite4:
This adds a couple of libraries that can parse webserver
log files, parse user agent strings and a geoip lookup library.

The geoip library is Apache Licenced, but the corresponding
geoip database is not redistributable so is not included.

The WebServerLogFile endpoint has configuration settings to
specify the log file format (and hence the records extracted) and
a config setting for the path to the goeip database. If there
is no geoip database then it will still work.

See the xdmod-ood module for CI tests and usage etc.


plessbd:
for reference: https://github.com/jpwhite4/xdmod-ood

jsperhac (approved):

jsperhac:
I eyeballed this PR and approved it prior to merging with 9.5. Looks like there are some tests failing, however.

ryanrath:
@jsperhac I believe that the test failures are due to the chromium update that's breaking the report generator.

PR #1485, opened 2021-01-15
ETL Viewer

ryanrath:
## Description
I figured it was about time to get what I have for this up so that other people can take a look at it. You'll need to spin this up in a docker & then log in to the admin interface & you should see a new tab "ETL Viewer". Right now it just displays every pipeline / action / action details in a TreeGrid. You can ignore the "File" drop down on the right side of the interface. In the beginning I had it in my head that we'd want to start looking at things on a file by file basis, hence this dropdown. But the code really isn't setup to do this at all so I just pivoted to viewing the final data structures produced by the ETLConfiguration class. 

I think that an easy next step would be to add a "Search" box that maybe searches each node & expands the nodes that match? Or it could be that it just returns all the matches and which pipeline / action it was found in. After which , if you click a search result, it will expand / take you to the node you selected in the overall data structure? 

## Motivation and Context
The ETL Config system is pretty spiff, but it can be hard to visualize / figure out where everything is / how it goes together. 
Ultimately I want to add a visualization that makes it easy to understand where data is coming from & where it's going for each pipeline / action. I'll be adding a couple of screenshots of what I've experimented with so far. 

## Checklist:


jsperhac:
@ryanrath, I am excited to hear about this feature...super cool.

ryanrath:
Here's a super rough concept that I worked on for a way of visualizing info for the ETL Viewer.

etl_viewer_visualizer

The big problem that I've run into so far w/ this method has been working out how to automatically layout the elements for any given pipeliine. I manually selected the data for this image to illustrate the basic concept of what I was looking to achieve but it turns out that working out how to get cytoscape to lay things out in this 3 column manner is somewhat... challenging to say the least. Ultimately I think that I'd need to write a custom layout engine ( which cytoscape supports ) if this was something we wanted to move forward with. 

If anybody has any other ideas / thoughts please hit me up as I would really love to get further input on this.

PR #1490, opened 2021-02-02
Ensure correct log message from ETL overseer.

jpwhite4:
The php type engine gets confused when you try to string convert an
iAction object in the Logger code. Possibly something to do with the way
the class is constructed or the loader????

Anyway the old code results in the iAction object being treated as an
Array like object in the classes/CCR/CCRLineFormatter.php on line 44.
This change ensures a string is sent to the logger since that is
what we want.

### DIFF
--- a/classes/ETL/EtlOverseer.php
+++ b/classes/ETL/EtlOverseer.php
@@ -460,7 +460,7 @@ class EtlOverseer extends \CCR\Loggable implements iEtlOverseer
         $this->logger->info(array(
                                 'message'     => 'start',
                                 'action_name' => $actionName,
-                                'action'      => $actionObj,
+                                'action'      => (string) $actionObj,
                                 'start_date'  => $this->etlOverseerOptions->getStartDate(),
                                 'end_date'    => $this->etlOverseerOptions->getEndDate(),
                                 ));
@@ -486,7 +486,7 @@ class EtlOverseer extends \CCR\Loggable implements iEtlOverseer
         $this->logger->info(array(
                                 'message'    => 'end',
                                 'action_name' => $actionName,
-                                'action'     => $actionObj
+                                'action'     => (string) $actionObj,
                                 ));
     }  // _execute()
 }  // class EtlOverseer


ryanrath (approved):

ryanrath:
So I believe I've found the root problem that was causing this behavior. In `\CCR\Logger.php::extractMessage` starting at line 46: 
```php
 if (is_array($record)) {
   return json_encode($record);
}
```

This code is the reason that a PHP Warning about an Array to String conversion is thrown in  `CCRLineFormater::extractMessage` on line 44 `$nonMessageParts[] = "$key: $value";`. 

Here's an example: Given the following logger call
```php
$logger->info(array(
    'message'     => 'start',
    'action_name' => $actionName,
    'action'      => $actionObj,
    'start_date'  => $overseerOptions->getStartDate(),
    'end_date'    => $overseerOptions->getEndDate(),
));
```

the result of `Logger::extractMessage` looks like this `{"message":"start","action_name":"db-model-test.create-table","action":{},"start_date":null,"end_date":null}`. Note the empty `action` object. 

Which, when it get's to `CCRLineFormatter::extractMessage` and is subsequently decoded on line 29 w/ `json_decode($record['message'], true)`. The value of `action` is converted to an associative array. Which causes the PHP Warning about converting an Array to string to be thrown on line 44 as the code attempts to convert the empty array value of `action` to a string. 

There's a couple of different ways we can fix this. One is obviously the way you've presented here which definitely works. 

Another would be to update `Logger::extractMessage`  to look like this: 
```php
protected function extractMessage($record)
{
    if (is_array($record)) {
        while (list($key, $value) = each($record)) {
            if (is_object($value)) {
                $record[$key] = (string) $value;
            }
        }
        return json_encode($record);
    }
    return json_encode(array('message' => $record));
}
```

And the last way to resolve this would be to implement the `JsonSerializable` interface for the ETL Action classes which would make sure that they produce the expected output when  `json_encode` is used. 

Thoughts?

ryanrath:
@jpwhite4 So a quick update, I was going back through the PHP7 & MariaDB updates and it turns out that I need to include the changes to `Logger::extractMessage`. The problem I ran into was when running the DBModelTest component tests they all would error out with `SQLSTATE[42000]: Syntax error or access violation: 1049 Unknown database 'test'`.  The reason for this particular error is that the function ( `BaseETLTest::executeEtlAction`) used to execute an ETL action was directly constructing / executing an ETL action object as opposed to using the ETL Overseer. Once I updated this function to utilize the overseer I started getting our good 'ol friend `Array to string converstion` error.  Which I was able to resolve with the changes I mentioned above.

jpwhite4:
This pull request is superseded by the #1493

PR #1493, opened 2021-02-08
Updating the logging code to not throw away all class information

ryanrath:
## Description
We ran into a problem described in this PR: https://github.com/ubccr/xdmod/pull/1490 After some research I found that the
root of the problem is this function json_encode'ing `$record` when the contents
of `$record` may contain objects that do not support `JsonSerializable` but that
do implement `toString()`. After a conversation w/ Joe it's been decided that
our Logging API will be to support the logging of arrays w/ objects that
implement a `toString()` function. The eagle-eyed may notice that this solution
differs from the one I put forth in the above PR. The reason being that the
original solution only supported the use case of having an array w/ top level
value objects and would not work as expected if a nested array is supplied with
an object that implements `toString()`.

## Motivation and Context
It's nice to have a logging api that works as expected. 

## Tests performed
All automated tests.

## Checklist:

### DIFF
--- a/classes/CCR/Logger.php
+++ b/classes/CCR/Logger.php
@@ -44,8 +44,28 @@ class Logger extends \Monolog\Logger implements LoggerInterface
     protected function extractMessage($record)
     {
         if (is_array($record)) {
-            return json_encode($record);
+            return json_encode($this->recursivelyStringifyObjects($record));
         }
         return json_encode(array('message' => $record));
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


jpwhite4:
Loks good. Please can you add a description of what this logger supports at the top of the file (so we don't forgret in future).

Something like:
If an object is passed to the logger then it will use the __toString() function to convert the object to a string.

ryanrath:
@jpwhite4 Absolutely, I'll add that and push it up momentarily.

jpwhite4 (approved):

PR #1500, opened 2021-02-25
Add indexes that will help federated module ingest be faster

eiffel777:
These indexes are being added to speed up the fed-ingest pipeline in the [federation](https://github.com/ubccr/xdmod-federated) module. 

Both indexes are used by the [instance-jobhosts action](https://github.com/ubccr/xdmod-federated/blob/xdmod9.0/configuration/etl/etl_action_defs.d/federated/jobs/jobhosts.json).

## Description
Adding indexes so ingestors in the federated module perform better

## Tests performed
Tested in docker and with federated module

## Checklist:


ryanrath (approved):

jpwhite4:
Did you experiment with adding any compound indexes on origin_id, resource_id, host_id? What queries are using these indexes?

eiffel777:
@jpwhite4 

I don't think I tested a compound index. Looking it back over it I could add resource_id to each index. That might make it a little faster. Let me do some testing and see if there is a difference.

Both indexes are used in the same query in the federated module. The etl_action_def.d/ file where the query is defined is [here](https://github.com/ubccr/xdmod-federated/blob/xdmod9.0/configuration/etl/etl_action_defs.d/federated/jobs/jobhosts.json)

Here is the join that uses `host_origin_id_idx`

```
{
        "schema": "${DESTINATION_SCHEMA}",
        "name": "hosts",
        "alias": "dh",
        "on": "dh.resource_id = drf.id AND sh.id = dh.host_origin_id"
}
```

The `job_id_origin_id_idx` is used for this join.

```
{
        "schema": "${DESTINATION_SCHEMA}",
        "name": "job_tasks",
        "alias": "djt",
        "on": "djt.job_id_origin_id = sjh.job_id AND djt.resource_id = drf.id"
}
```

eiffel777:
I did some testing with adding the resource_id to the indexes and it doesn't help much. It's only a couple of seconds so it doesn't seem like it would be very helpful to add them.

PR #1504, opened 2021-03-05
Fix extraneous cloud sessions

eiffel777:
https://app.asana.com/0/808093868887967/1199187338120726

This fixes an issue where extra sessions may exist in the `session_records` table when ingesting data to fill a time gap where data was not ingested. An example of a gap in the data could be the file copying not running on a given day and then a catch-up run being done later. The `CloudStateReconstructorTransformIngestor` constructs all sessions for all instances that have an event greater than the last_modified_start_date specified when you run `xdmod-ingestor` or the relevant `etl_overseer.php` command and places the in the event_reconstructed table. 

To make sure we have the correct sessions this change removes all the sessions for instances we are aggregating from the session_records table. This means that when the `cloud-session-records` action runs it will load all the sessions from the event_reconstructed table which has the correct sessions for each instance and none of the extra sessions.

It also removes rows from `cloudfact_by_day_sessionlist` that have session_ids tied to instances that are being aggregated. If this is not done it would result in rows in `cloudfact_by_day_sessionlist` that have session_ids that no longer exist.

## Motivation and Context
Preventing sessions that are incorrect from existing in the `session_records` table

## Tests performed
Tested in docker

## Checklist:


jpwhite4:
When you re-aggregate a row it should automatically remove the corresponding rows from the session table. Its not clear to me why you need this extra step.

jpwhite4:
I think i have found the answer to my question. But the question is now why didn't you use the dedicated aggregation class that is designed for the purpose of handling aggregation with session\|job tables?

jpwhite4 (approved):

PR #1505, opened 2021-03-05
Update simple SAML documentation to mention the icon.

jpwhite4:
The documentation made no mention of the fact that you can use an image
in the login dialog box. Now it does!


plessbd (approved):
@plessbd should have added this a long time ago, shame on him...

PR #1506, opened 2021-03-09
Add missing attributes when gearing over from Usage to Metric

jpwhite4:
This fixes an issue with adding data to charts that were originally configured by "gearing over" from the Usage tab.

Steps to reproduce the original problem:
1) View a chart in the Usage tab
2) Gear over to load the chart in the Metric Explorer
3) Add data for a _different_ realm to the chart
4) The Realm label  should appear as a prefix for the dataset labels for each dataset. But is is missing for the first dataset.

For the original bug report see:

See https://app.asana.com/0/808093868887967/1184327725706337

Note the bug is fixed by the one line that populates the category object property. However, in investigating the bug there were several other configuration keys that were not defined. This change also ensures that all the required properties are given a value.

### DIFF
--- a/html/gui/js/modules/Usage.js
+++ b/html/gui/js/modules/Usage.js
@@ -2173,6 +2173,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                                         id: Math.random(),
                                         metric: n.attributes.statistic,
                                         realm: n.attributes.realm,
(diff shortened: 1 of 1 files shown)


jpwhite4:
Before: 
image

After:
image

jsperhac (approved):
Looks good to me, Joe. Thanks for fixing!

PR #1510, opened 2021-03-11
Updating file permissions for XDMoD's bin directory

ryanrath:
## Description

We're updating the permissions of files in XDMoD's `bin` directory from `0755`
to `0750` to ensure only users in the xdmod group ( or root ) can execute these
files. Since these privs are more restrictive I've also updated the group for the bin directories files to `xdmod` to ensure that our automated builds work correctly. 

To ensure that these changes are also made when a source install is performed I've added a section to the source install documentation that has the command that should be executed. 

## Motivation and Context
These changes are to address TrustedCI 6.2.5, tracked at
https://app.asana.com/0/15230011293655/1188274656222921 .

## Tests performed
All Automated tests were performed. Also, the permissions of the files in question were manually checked locally to ensure they appear as expected.

## Checklist:


docs/install-source.md
```diff
@@ -52,6 +52,10 @@ Update log directory and file ownership and permissions:
 The `exceptions.log` and `query.log` may be written to by both Apache and Open
 XDMoD commands.
 
+Update the bin directory permissions:
+
+    # chmod 0750 /opt/xdmod/bin/*
```
jpwhite4, line 57: probably better to just suggest  `chmod 750 /opt/xdmod/bin/*` because the other chmod commands all just set the three flags and not four.

docs/install-source.md
```diff
@@ -52,6 +52,10 @@ Update log directory and file ownership and permissions:
 The `exceptions.log` and `query.log` may be written to by both Apache and Open
 XDMoD commands.
 
+Update the bin directory permissions:
```
jpwhite4, line 55:
```suggestion
Update the bin directory permissions so that only users with the `xdmod` group can run XDMoD commands:
```

jpwhite4 (approved):

PR #1519, opened 2021-03-29
Change chromium output error detection.

jpwhite4:
Starting with chromium-headless-89.0.4389.82-1.el7.x86_64 chrome crashes on
exit (due to the fortify stack smashing code firing). This appears to happen
on exit and does not prevent chrome from rendering the svg. This change
works around the problem by using the presence of valid output as the
test condition and ignores the chrome exit code.


jpwhite4:
Note that the existing image export tests confirm that the export is actually working properly.

ryanrath (approved):

PR #1523, opened 2021-04-12
Move truncate_declaration directives to correct spot for cloud resource specs migration

eiffel777:
The truncate_destination directive for the cloud_resource_specs table during the upgrade was in the wrong spot meaning the table was not truncated.

## Tests performed
Tested in docker

## Checklist:

### DIFF
--- a/configuration/etl/etl.d/xdmod-migration-9_0_0-9_5_0.json
+++ b/configuration/etl/etl.d/xdmod-migration-9_0_0-9_5_0.json
@@ -38,8 +38,7 @@
                    "type": "mysql",
                    "name": "Cloud DB",
                    "config": "datawarehouse",
-                   "schema": "modw_cloud",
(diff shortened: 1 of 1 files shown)


configuration/etl/etl.d/xdmod-migration-9_0_0-9_5_0.json
```diff
@@ -239,6 +239,7 @@
             "description": "Aggregate cloud records.",
             "definition_file": "cloud_common/cloud_resource_specs_aggregation.json",
             "table_prefix": "resourcespecsfact_by_",
+            "truncate_destination": true,
```
jpwhite4, line 242: The aggregator class will automatically delete the target rows in the table when it runs so usually truncate_destination is not used in an aggregator. Please can you explain why it is needed here.

jpwhite4 (approved):

PR #1527, opened 2021-04-20
Updates for cloud Data Export and migration pipelines.

eiffel777:
This PR fixes three issues.

1. Using the correct table for the time field in the `JobDataset.php` file. The `session_records` table has the `start_time_ts` and `end_time_ts` columns that should be used for the start time and end time parameters. It was trying to use the instance table which does not have those columns.
2. Changing the `host` table to have a single autoincrement column instead of a composite autoincrement. In PR #1263 we changed most of the cloud realm dimension tables that use a composite autoincrement key to use a single unique column for the autoincrement key. At the time the `host` table was not used for anything so it was not converted. It is now being used in the cloud instance viewer and Data Export and should be changed to match the other cloud realm dimension tables so it can provide accurate data. The changes for the `host` table match changes made for the `instance`, `account`, and `instance_type` tables in PR #1263 and #1272 including a SQL statement that is run during the migration to add the new autoincrement key and update rows in the event table with the appropriate `host_id`.
3. Renaming the cloud migration pipelines so they are not automatically run. Having the cloud migration pipelines with `migration-9_0_0-9_5_0` in them meant they automatically ran. We do not want that as not all XDMoD installations may be using the cloud realm. 

## Tests performed
Tested fresh install and upgrade in docker

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Cloud/JobDataset.php
+++ b/classes/DataWarehouse/Query/Cloud/JobDataset.php
@@ -44,10 +44,27 @@ class JobDataset extends \DataWarehouse\Query\RawQuery
                 '=',
                 new TableField($tables[$join['foreignTableAlias']], $join['foreignKey'])
             ));
+
+            /*if(is_array($join[0])) {
+                foreach($join as $j) {
+                    $this->addWhereCondition(new WhereCondition(
+                        new TableField($table, $j['primaryKey']),
+                        '=',
+                        new TableField($tables[$j['foreignTableAlias']], $j['foreignKey'])
+                    ));
+                }
+            } else {
+                $this->addWhereCondition(new WhereCondition(
+                    new TableField($table, $join['primaryKey']),
+                    '=',
+                    new TableField($tables[$join['foreignTableAlias']], $join['foreignKey'])
+                ));
+            }*/
         }
 
         // This table is defined in the configuration file, but used in the section below.
         $factTable = $tables['i'];
+        $sessionTable = $tables['sr'];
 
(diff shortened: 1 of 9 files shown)


jtpalmer:
I'm assuming this is unrelated, but thought I would mention it here while I was checking this:

```
2021-04-21 18:20:16 [notice] Altering table `modw_cloud`.`session_records`
2021-04-21 18:20:16 [warning] SQL warnings on table '`modw_cloud`.`session_records`' generated by action xdmod.cloud-migration-9-0-0_9-5-0.cloud-session-records
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
2021-04-21 18:20:16 [warning] Warning 1048 Column 'memorybucket_id' cannot be null
```

jtpalmer (approved):

PR #1539, opened 2021-07-12
Update jQuery library to 3.6.0

jpwhite4:
Will also have update for the other xdmod modules too.


jtpalmer (approved):
Why are the tests failing?

jpwhite4:
Can't drop in 3.6 due to breaking changes between 1 and 3 :-(

PR #1542, opened 2021-07-14
Remove all calls to jQuery functions.

jpwhite4:
WIP - need to find a licence compatible replacement for extend.....

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -11,11 +11,6 @@ License: New BSD License
 Author:  Zend Technologies USA, Inc.
 Link:    http://framework.zend.com/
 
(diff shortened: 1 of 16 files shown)


versable:
This might be of value:
https://youmightnotneedjquery.com/#deep_extend

aestoltm:
There still does not exist a native JS function for replacing jQuery.extend(). The [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#transferring_values) function causes the original object that is transferred from to become unusable.

The currently provided solution in the PR might be our best option for the moment.

PR #1545, opened 2021-07-27
Remove PHP GNU Multiple Precision dependency

jtpalmer:
## Description

 
Removes references to `php-gmp` from the documentation and `xdmod-check-config`.

## Motivation and Context

I can't find anywhere that GMP is required.  None of the `gmp_*` functions are used anywhere in Open XDMoD.

## Tests performed

No new tests.

## Checklist:

### DIFF
--- a/bin/xdmod-check-config
+++ b/bin/xdmod-check-config
@@ -124,13 +124,6 @@ try {
         'PHP GD extension installed'
     );
 
-    _debug('Checking for gmp_abs function');
-    $result = function_exists('gmp_abs');
-    displayResult(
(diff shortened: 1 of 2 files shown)


jpwhite4:
I wonder if it was used in the distant past when more heavy lifiting was done in php?

jpwhite4 (approved):

jtpalmer:
I tried to find when it may have been used in the past, but couldn't find any direct references at all.  I added those checks so I must have thought it was being used by something at some time, maybe it was used by an old dependency.

PR #1546, opened 2021-08-03
Change slurm helper default end time

jtpalmer:
## Description

Changes the default end time used by the `xdmod-slurm-helper` to be the current time.

## Motivation and Context

With the previous default end time it was possible that the default end time would be before the default start time.  This resulted in an error message from `sacct`.

See https://app.asana.com/0/159049597309611/1199962348183946/f

## Tests performed

Tested manually on metrics-dev.

## Checklist:

### DIFF
--- a/bin/xdmod-slurm-helper
+++ b/bin/xdmod-slurm-helper
@@ -277,10 +277,7 @@ function getSacctCmdArgs(
         $args[] = $endTime;
     } else {
         $end = new DateTime('now');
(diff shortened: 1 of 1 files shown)


jpwhite4:
When we do the changelog it would be good to list this change as a potential braking change - there may be folks who are relying on the fact that it defaults to only grabbing yesterday's data.

jpwhite4 (approved):

PR #1547, opened 2021-08-05
Update docs to include the OnDemand module.

jpwhite4:
Also switch SUPReMM to Job Performance (SUPReMM) since I was editing the file anyway.


jpwhite4:
Note this is going into 9.5 so we can update the 9.5 docs on the website. Will also have another pull request that merges up to 10

jtpalmer (approved):

jtpalmer:
It looks like this was removed from the xdmod9.5 branch.  Just noting that here if this branch should be reopened.

PR #1551, opened 2021-08-13
Update jQuery to 3.6.0

jpwhite4:
The `ready()` function in jQuery 3 has different behaviour
than the one in 1.x and cannot be used in the chart export
code path. This has been replaced with an event listener for
DOMContentLoaded. Only other changes are updating the filename.


jpwhite4:
Note there is also a corresponding pull request to update the supremm module too https://github.com/ubccr/xdmod-supremm/pull/283

ryanrath (approved):

PR #1553, opened 2021-08-13
Remove last vestiges of shippable.

jpwhite4:
Shippable no longer exists so we don't need the configuration files. Also
expunge the SHIPPABLE environment variables.


jtpalmer:
Is the root `Dockerfile` used for anything now?

jtpalmer (approved):

PR #1557, opened 2021-08-17
Add VM Host group by for cloud realm

eiffel777:
This PR adds a VM Host group by for the Cloud realm. New regression test artifacts are included as well as adding migration pipelines since the Cloud realm will need to be re-aggregated.

## Tests performed
Tested in docker with CCR and Jetstream data.

## Checklist:


configuration/datawarehouse.d/ref/Cloud-group-bys.json
```diff
@@ -84,6 +84,37 @@
     "month": {
         "$ref": "datawarehouse.d/ref/group-by-time-period.json#/month"
     },
+    "host": {
+        "attribute_table_schema": "modw_cloud",
+        "attribute_to_aggregate_table_key_map": [
```
jpwhite4, line 89: This is not going to work properly if there is more than one resource in the database: the primary key on the host table is resource_id, hostname. So host_id may have two different values for the same hostname string.

jpwhite4 (changes requested):

PR #1558, opened 2021-08-20
Add rest documentation

jpwhite4:
This includes the rest schema file containing docs for the controller interfaces
    and the rest interfaces. Some of these are missing descriptions, but this at least
    is a place holder to get going.

    Also add a static html page that pulls in a 3rd party js library that auto-generates
    sensible looking html output.

image


jtpalmer:
@jpwhite4 Want to merge this into xdmod10.5 soon so we don't forget?  Or is there more you've got for 10.0?

PR #1566, opened 2021-09-09
Handle MySQL server has gone away errors.

jpwhite4:
This change handles the scenario when there is a long delay
between two different log message calls (for the database logger).
This can happen when aggregating / ingesting large amounts of data
and the log level is set low.

The change is to attempt a single retry if the specific 'server has
gone away' error is seen. This shouldn't change the behaviour for any
other types of sql error. If the retry fails then it does not try again.

There is not an easy way to test this in the CI builds so I manually tested
as follows:

This was tested by changing the `/etc/my.cnf.d/server.cnf` file in a
docker instance and adding the following:

```ini
wait_timeout = 30
```

I then put a `sleep(60)` as the last line of the run_aggregation() function
in /usr/lib64/xdmod/aggregate_supremm.php and ran it. (I also had the EtlJournalHelper
patch https://github.com/ubccr/xdmod-supremm/pull/285 installed).

Confirmed that the full processing happens with this patch installed and
errors out with a
```json
{"success":false,"count":0,"total":0,"totalCount":0,"results":[],"data":[],"message":"SQLSTATE[HY000]: General error: 2006 MySQL server has gone away","code":"HY000"}
```
otherwise.

### DIFF
--- a/classes/CCR/CCRDBHandler.php
+++ b/classes/CCR/CCRDBHandler.php
@@ -93,7 +93,20 @@ class CCRDBHandler extends AbstractProcessingHandler
      */
     protected function getNextId()
     {
-        $this->db->beginTransaction();
+        try {
+            $this->db->beginTransaction();
+        } catch (\PDOException $e) {
+            if ($e->errorInfo[0] === 'HY000' && $e->errorInfo[1] === 2006) {
+                // This is the MySQL server gone away error, which is seen when
+                // there is a long delay between log messages and the
+                // connection times out. It occurs here since this is the first DB
+                // call for a log message.
+                $this->db->disconnect();
+                $this->db->beginTransaction();
+            } else {
+                throw $e;
+            }
+        }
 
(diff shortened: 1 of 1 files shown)


jtpalmer (approved):

classes/CCR/CCRDBHandler.php
```diff
@@ -93,7 +93,20 @@ protected function write(array $record)
      */
     protected function getNextId()
     {
-        $this->db->beginTransaction();
+        try {
+            $this->db->beginTransaction();
+        } catch (\PDOException $e) {
+            if ($e->errorInfo[0] === 'HY000' && $e->errorInfo[1] === 2006) {
```
jtpalmer, line 99:
```suggestion
            if ($e->errorInfo[0] === 'HY000' && ($e->errorInfo[1] === 2006 || $e->errorInfo[1] === 2013)) {
```

PR #1571, opened 2021-11-08
Add SHOW WARNINGS to StructuredFileIngestor

eiffel777:
The StructuredFileIngestor does not run `SHOW WARNINGS` after ingesting data like the PDOIngestor does. This PR adds running the `SHOW WARNINGS` command for the StructuredFileIngestor and displays any warnings that exist.

## Motivation and Context
Make sure that any warnings from MySQL after inserting data is shown. This matches what is done in the PDOIngestor. There are some warnings that were being missed because of this and will be fixed in another pull request.

## Tests performed
Tested in docker

## Checklist:


classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -325,10 +327,20 @@ function ($k, $v) {
                         array('exception' => $e, 'endpoint' => $this)
                     );
                 }
+
+                $warning = $this->destinationHandle->query("SHOW WARNINGS");
+
+                if ( count($warning) > 0 ) {
+                    $warnings[$etlTableKey] = array_key_exists($etlTableKey, $warnings) ? array_merge($warnings[$etlTableKey], $warning) : $warning;
```
jpwhite4, line 334: Elsewhere in the etlv2 code there is logic to suppress warning messages for certain error codes. These are configured by well defined configuration keys:

https://github.com/ubccr/xdmod/blob/d827ad80e7f2baaa565d31c20a13a0a273fd9ffc/classes/ETL/Ingestor/IngestorOptions.php#L98-L103

The question is how does this change fit in with this existing code? Should the "hide_sql_warnings" and "hide_sql_warning_codes" configuration parameters be supported (hint: yes).

eiffel777: @jpwhite4 The does support suppressing the warning messages for certain error codes. The `logSqlWarnings` function takes care of this.

https://github.com/ubccr/xdmod/blob/4523ea1240a0d2112a0b1f50c5aa43584e326ca5/classes/ETL/aRdbmsDestinationAction.php#L898-L900

https://github.com/ubccr/xdmod/blob/4523ea1240a0d2112a0b1f50c5aa43584e326ca5/classes/ETL/aRdbmsDestinationAction.php#L911-L926

PR #1581, opened 2021-12-10
Updates to mod_logger tables

ryanrath:
## Description
- I've moved from the previous INSERT / DELETE algorithm to using an UPDATE.
This allows us to remove the explicit transaction management as we only have one
query that applies modifications as opposed to two. This new method was chosen
after a conversation with @jpwhite & @gmdean.
- To implement this change there were a couple of small changes that needed to
be made to the log_id_seq table schema & it's initial population.
  - `mod_logger`.`log_id_seq`.`sequence` has had it's `AUTO_INCREMENT` keyword
  removed as we'll be manually controlling the way the sequence is generated.
  - Since the `sequence` column is no longer an `AUTO_INCREMENT` we need to add
  an initial value, this has been done in `data/db/mod_logger.sql` by inserting
  an initial value of `0`.

## Motivation and Context
These changes are an attempt to remediate lock problems that we're seeing since
the change to InnoDB.

## Tests performed
These changes have been manually tested locally w/ both an upgrade & fresh
install.

Example test ( Executed at various points during upgrade & fresh install )
```sql
SELECT
( SELECT sequence FROM log_id_seq ) as current_id,
( SELECT COUNT(*) FROM log_table ) as log_count,
( SELECT MAX(id) FROM log_table ) as current_log_id;
```

## Checklist:

### DIFF
--- a/classes/CCR/CCRDBHandler.php
+++ b/classes/CCR/CCRDBHandler.php
@@ -93,35 +93,10 @@ class CCRDBHandler extends AbstractProcessingHandler
      */
     protected function getNextId()
     {
-        try {
-            $this->db->beginTransaction();
-        } catch (\PDOException $e) {
-            if ($e->errorInfo[0] === 'HY000' && $e->errorInfo[1] === 2006) {
-                // This is the MySQL server gone away error, which is seen when
-                // there is a long delay between log messages and the
-                // connection times out. It occurs here since this is the first DB
-                // call for a log message.
-                $this->db->disconnect();
-                $this->db->beginTransaction();
-            } else {
-                throw $e;
-            }
-        }
-
-        try {
-            $this->db->execute(sprintf('INSERT INTO %s.log_id_seq (sequence) VALUES(NULL);', $this->schema));
-
-            $results  = $this->db->query('SELECT LAST_INSERT_ID() as id');
-
-            $id = $results[0]['id'];
-
-            $this->db->execute(sprintf("DELETE FROM %s.log_id_seq WHERE sequence < :id", $this->schema), array(':id' => $id));
-
-            $this->db->commit();
-        } catch (Exception $e) {
-            $this->db->rollBack();
-            throw $e;
-        }
+        $this->db->execute(sprintf('UPDATE %s.log_id_seq SET sequence = LAST_INSERT_ID(sequence+1);', $this->schema));
+        $stmt = $this->db->query('SELECT LAST_INSERT_ID() as id', array(), true);
+        $stmt->execute();
+        $id = $stmt->fetchAll(\PDO::FETCH_COLUMN, 0)[0];
 
         return (int) $id;
     }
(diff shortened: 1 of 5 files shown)


classes/CCR/CCRDBHandler.php
```diff
@@ -93,35 +93,10 @@ protected function write(array $record)
...
-            $id = $results[0]['id'];
-
-            $this->db->execute(sprintf("DELETE FROM %s.log_id_seq WHERE sequence < :id", $this->schema), array(':id' => $id));
-
-            $this->db->commit();
-        } catch (Exception $e) {
-            $this->db->rollBack();
-            throw $e;
-        }
+        $this->db->execute(sprintf('UPDATE %s.log_id_seq SET sequence = LAST_INSERT_ID(sequence+1);', $this->schema));
```
jpwhite4, line 96: You'll need to handle the MySQL gone away error - see the catch statements starting on line 98 and if statment on 99.

jpwhite4: You can test the "mysql gone away" error is handled properly following the instructions in https://github.com/ubccr/xdmod/pull/1566 (note you don't have to use the aggregate_supremm script - anything that does two log messages with a sleep in between should work).

jpwhite4 (approved):

classes/CCR/CCRDBHandler.php
ryanrath: Alrighty, manually tested this in a docker container w/ these changes added by adding the following line ( from #1566): to `/etc/my.cnf.d/server.cnf`
```ini
wait_timeout = 30
```

Creating the following test script: `/usr/share/xdmod/test.php`
```php
<?php

include './configuration/linker.php';

use \CCR\Log;

$logger = Log::factory(
            'TestingLogger',
            array(
                'file' => false,
                'db' => true,
                'mail' => false,
                'console' => false
            )
        );
try {
    $logger->info("Testing123");
    sleep(120);
    $logger->info("Oh Noes!");
} catch (Exception $e) {
    // Write any unexpected exceptions directly to STDERR since they may not
    // have been logged and it may not be able to create a log instance.
    fwrite(STDERR, "Data warehouse batch export failed\n");
    do {
        fwrite(STDERR, $e->getMessage() . "\n" . $e->getTraceAsString() . "\n");
    } while ($e = $e->getPrevious());
    exit(1);
}
```

and to prove to myself that things were working I inserted a couple of echo's in `classes/CCR/CCRDBHandler.php` right after the `catch` clause in `getNextId` to make sure we are getting an exception and then right after the `if` statement in the `catch` clause to make sure it's the exception that we're expecting. 

Running the script produced the following output:
```shell
[root@3f4f62c19799 xdmod]# php test.php
We encountered some sort of error!
The error was server gone away!
[root@3f4f62c19799 xdmod]# echo $?
0
```

jpwhite4 (dismissed):

jpwhite4 (approved):

PR #1584, opened 2021-12-13
Convert tables in modw_aggregates to InnoDB and add class to manage aggregate tables

eiffel777:
This PR sets the table engine for the `jobfact_by_` and `storagefact_by_` aggregate tables to be InnoDB. Since there is not ETLv2 class to manage aggregate tables I created one called `ManageAggregateTables` that extends `ManageTables` and adds support for the aggregation_unit ETLv2 variable.

## Tests performed
Tested in docker and on metrics-staging

## Checklist:


classes/ETL/Maintenance/ManageAggregateTables.php
```diff
@@ -0,0 +1,139 @@
...
+                    $msg = "Error managing ETL table " . $etlTable->getFullName() . ": " . $e->getMessage();
+                    $this->logAndThrowException($msg);
+                }
+            }
+        }
+
+        $time_end = microtime(true);
+        $time = $time_end - $time_start;
+
+        $this->logger->notice(array('action'       => (string) $this,
```
jpwhite4, line 133: Just wanted to confirm that the string cast for `$this` works properly. I have a vague recollection that we saw "[object]" elsewhere in the logs in the past (not logs from this code change).

eiffel777: @jpwhite4 Yes, it appears to be working correctly. Here is an example of the output

```
2021-12-14 20:32:14 [notice] (action: xdmod.migration-9_5_0-10_0_0.update-mod-aggregate-tables-by-day (ETL\Maintenance\ManageAggregateTables), start_time: 1639513933.711, end_time: 1639513934.2564, elapsed_time: 0.54539)
```

jpwhite4 (approved):

PR #1588, opened 2021-12-15
Updated etl_profile.js to add an extra group_by parameter "show_all_dimension_values".

hltaylor26:
## Description
This update adds the show_all_dimension_values flag to any group_bys generated by the etl.schema file in other realms. Defaults to false when created. 

## Motivation and Context
Allows for specific dimensions to return all values when need. For example, we want all CPU User buckets to show in the Efficiency Tab and setting this flag to true allows for the possibility for any user to access all possible buckets (not just buckets which contain their data). 

## Tests performed
Tested in docker. 

## Checklist:


etl/js/lib/etl_profile.js
```diff
@@ -698,7 +699,8 @@ var generateGroupBy = function (itemAlias, column)
         },
         data_sort_order: null,
         description_html: description,
-        name: label || itemAlias
+        name: label || itemAlias,
+        show_all_dimension_values: column.show_all_dimension_values || false
```
jpwhite4, line 703: probably don't need the `|| false` here since the column has this property (from line 373)

etl/js/lib/etl_profile.js
```diff
@@ -764,7 +766,7 @@ ETLProfile.prototype.integrateWithXDMoD = function () {
                         }
                         xdmodInteg.addGroupBy(itemName, tableColumns[tc].roles);
                     }
-                }
+                }
```
jpwhite4, line 769: Please remove this whitespace addition. It messes up the `git blame`

jpwhite4 (changes requested):

jpwhite4 (dismissed):

jpwhite4 (approved):

PR #1592, opened 2021-12-16
Fix warnings seen when ingesting cloud or storage files

eiffel777:
After merging PR #1586, MySQL warnings that were not displayed are now displayed. Those warnings come from the shredding of storage and cloud data. 

For storage data, the `dt` column truncates the date string specified because it contains timezone information. Date time strings with timezones are not compatible with `datetime` columns. Changing this column to a `varchar` and then using `STR_TO_DATE` to get the correct date format fixes this warning. Since the timezone information does not appear to be used in the storage realm, this warning does not cause a bug.

Example of storage data warning:
```
2021-12-16 16:31:30 [warning] Warning 1265 Data truncated for column 'dt' at row 1
```

The date warning seen in the storage data exists for the cloud resource specifications as well. The warning is prevented by converting to a `varchar` column and using `STR_TO_DATE` when loading the date into a `datetime` column. The other warning is for the `size` column, defined as `int(11)`. The `size` field for `image.*` events are measured in bytes, and the Jetstream test data has images sizes that are too big for that column definition. The column can be changed to a `bigint` to fix this. The `image.*` events are not used and discarded during ingestion, so this does not cause a bug or require a re-ingest of data.

Example of warning for cloud resource specifications:
```
2021-12-16 16:27:50 [warning] Warning 1265 Data truncated for column 'fact_date' at row 1
```

Example of warning for cloud events:
```
2021-12-16 16:27:49 [warning] Warning 1264 Out of range value for column 'size' at row 1
```

## Tests performed
Tested in docker

## Checklist:


configuration/etl/etl_action_defs.d/storage/hpcdb/usage.json
```diff
@@ -11,8 +11,8 @@
             "account_id": "p.pi_id",
             "allocation_id": "pr.pi_resource_id",
             "allocation_breakdown_id": "upr.user_pi_resource_id",
-            "dt": "su.dt",
-            "ts": "UNIX_TIMESTAMP(su.dt)",
+            "dt": "STR_TO_DATE(su.dt, '%Y-%m-%dT%H:%i:%sZ')",
+            "ts": "UNIX_TIMESTAMP(STR_TO_DATE(su.dt, '%Y-%m-%dT%H:%i:%sZ'))",
```
jpwhite4, line 15: Does this conversion work properly? The storage docs clearly state that the time must be in UTC. I think you need to do something like the following:
```
CONVERT_TZ(STR_TO_DATE(su.dt, '%Y-%m-%dT%H:%i:%sZ'), '+00:00', @@session.time_zone);
```

jpwhite4 (changes requested):

configuration/etl/etl_action_defs.d/storage/hpcdb/usage.json
jpwhite4: https://open.xdmod.org/9.5/storage.html

jpwhite4 (approved):

PR #1603, opened 2022-01-27
Updating Highcharts version

ryanrath:
## Description
These changes are due to issues we ran into while working on the metrics-dev
centos8 test box. I was unable to freshly download Highcharts 4.2.5 via composer
due to highcharts web server returning an invalid content-encoding value. This
hadn't been an issue before due to us either caching the downloaded zip file (
i.e. the current centos8 build ) or was downloading it on Centos7 which has an
old enough version of curl that it doesn't care if the server provides an
invalid content-encoding.

With the change in version there was a slight change in the way that Highcharts
constructs it's chart svg so this necessitated a slight modification to a few of
the UI test helper functions.

I have a ticket into highcharts github repo
https://github.com/highcharts/highcharts/issues/16926 that is, at the time of
writing, has been routed to the website team and is actively being worked on.

## Motivation and Context
These changes allow us to successfully install on OS's with newer versions of curl 

## Tests performed
UI tests were performed on both Centos7 & Centos8

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,9 +67,9 @@
                 ],
                 "homepage": "https://www.highcharts.com/products/highcharts",
                 "dist": {
-                    "url": "https://code.highcharts.com/zips/Highcharts-4.2.5.zip",
+                    "url": "https://code.highcharts.com/zips/Highcharts-4.2.7.zip",
                     "type": "zip",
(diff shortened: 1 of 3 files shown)


jpwhite4:
I suggest that we close this and leave as unmerged. Suggested plan is to keep HighCharts at the current version and instead work on getting its replacement integrated into the code.

PR #1606, opened 2022-02-04
Refactor admin dashboard user listing query and add indexes to Users and SessionManager tables

jtpalmer:
## Description

Refactor SQL and add indexes to `moddb.Users` and `moddb.SessionManager` tables.

Note that the `SessionManager` table is already included in the InnoDB migration so does not need to be included in the migration changes here.

## Motivation and Context

User reported this query taking longer than a minute resulting in a time out in the admin dashboard.

https://app.asana.com/0/807770147051221/1201757805272796/f

https://ubccr.freshdesk.com/a/tickets/29169

## Tests performed

Ran original query using development database and compared output with that from the new query.

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/xdb/users.json
+++ b/configuration/etl/etl_tables.d/xdb/users.json
@@ -130,6 +130,12 @@
                 ],
                 "type": "BTREE",
                 "is_unique":  true
+            },
+            {
+                "name": "idx_user_type",
+                "columns": [
+                    "user_type"
+                ]
             }
         ],
         "triggers": [
--- a/classes/XDAdmin.php
+++ b/classes/XDAdmin.php
@@ -53,25 +53,16 @@ class XDAdmin
                 u.first_name,
                 u.last_name,
                 u.account_is_active,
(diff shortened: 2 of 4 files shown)


configuration/etl/etl_tables.d/xdb/users.json
```diff
@@ -130,6 +130,12 @@
                 ],
                 "type": "BTREE",
                 "is_unique":  true
+            },
+            {
```
ryanrath, line 134: Is this index needed? I don't see `user_type` in the query.

jtpalmer: It may be overkill, but `user_type = ...` is what is substituted in `$filterSQL` (https://github.com/ubccr/xdmod/blob/xdmod10.0/classes/XDAdmin.php#L46).

ryanrath: Oooo, that makes sense then. Carry on! :)

ryanrath (approved):

PR #1608, opened 2022-02-08
Drop schema version history tables

ryanrath:
## Description
The `schema_version_history` table is not being used currently in an effort to clean up the code base it is being removed here. 

## Motivation and Context
These changes originally arose from work I was doing to make sure that XDMoD can run on CentOS8 w/ newer versions of PHP & MariaDB. Specifically, I ran into problems when the ETL process would load the default data into these tables due to the data having a value of `NOW()` for the `action_datetime` column. This was caused by a change MariaDB made to be more strict in how it deals with incoming data so whereas in the past it would coerce `'NOW()'` ( notice the single quotes ) to `NOW()`, in more recent versions it does not coerce the string to a function call and instead treats it as a string. 

At first I attempted to update `action_datetime` to have a default value of `CURRENT_TIMESTAMP` ( and removing the `action_datetime` column from the data files. ) and while this worked well for the version of MariaDB available on CentOS8, the version on CentOS7 does not support `datetime` columns having a default value of `CURRENT_TIMESTAMP` ( this wouldn't be added until MySQL 5.6 or MariaDB 10.0 ). Upon discussion w/ @jtpalmer and @jpwhite4 it was decided that since this table wasn't currently being used it might as well be removed. 

## Tests performed
Automated tests were run locally on CentOS7 / 8. 

## Checklist:


db/data/mod_logger.sql
```diff
@@ -1,35 +0,0 @@
...
-/*!40103 SET TIME_ZONE=@OLD_TIME_ZONE */;
-
-/*!40101 SET SQL_MODE=@OLD_SQL_MODE */;
-/*!40014 SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS */;
-/*!40014 SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS */;
-/*!40101 SET CHARACTER_SET_CLIENT=@OLD_CHARACTER_SET_CLIENT */;
-/*!40101 SET CHARACTER_SET_RESULTS=@OLD_CHARACTER_SET_RESULTS */;
-/*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
-/*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;
-
```
jtpalmer, line 33: I think this is still needed.

jtpalmer (changes requested):
Please remove the `getDatabaseVersion` function from `classes/Xdmod/Version.php`.  It isn't used anywhere that I can find.  Also please update the description of the pull request.

db/data/mod_logger.sql
ryanrath: Yup, just replaced it, minus the schema_version_history insert at the end.

configuration/etl/etl_sql.d/migrations/9.5.0-10.0.0/drop_schema_version_history.sql
```diff
@@ -0,0 +1,7 @@
+-- DROP the schema_version_history table in mod_hpcdb, mod_logger, moddb, modw
```
jtpalmer, line 1:
```suggestion
-- DROP the schema_version_history table in mod_hpcdb, mod_logger, mod_shredder, moddb, modw
```

I'm against comments like this, but if you're going to include it then might as well include all the schemas.

configuration/etl/etl.d/xdmod-migration-9_5_0-10_0_0.json
```diff
@@ -203,39 +220,7 @@
...
-            "endpoints": {
-                "destination": {
-                    "type": "mysql",
-                    "name": "XDMoD Data Warehouse",
-                    "config": "datawarehouse",
-                    "schema": "mod_shredder"
-                }
-            }
-        },
+         },
```
jtpalmer, line 223:
```suggestion
        },
```

More indentation that shouldn't change.

classes/Xdmod/Version.php
```diff
@@ -35,14 +35,7 @@ class Version
...
 
-    /**
-     * Name of the schema version history table.
-     *
-     * @var string
-     */
-    protected static $versionTableName = 'schema_version_history';
-
-    /**
+      /**
```
jtpalmer, line 38:
```suggestion
    /**
```
Indentation shouldn't change here.

configuration/etl/etl_sql.d/migrations/9.5.0-10.0.0/drop_schema_version_history.sql
ryanrath: agreed about including all of the tables in the comment. What kind of comment would you prefer to see here?

jtpalmer: I don't think any comment is necessary.  If you really need one you could explain why the tables are being dropped.

jtpalmer (approved):
Please update the description of the pull request.

PR #1612, opened 2022-02-23
Updating the datasetid float comparison (PHP7.2)

ryanrath:
## Description
Comparing floats as strings is not desirable due to the way they are represented
internally. Per [https://www.php.net/manual/en/language.types.float.php](https://www.php.net/manual/en/language.types.float.php) they
should be compared to within a certain number of digits. This was found by
investigating an error where the "Show Raw Data" context menu item was not
showing the associated jobs. It turns out that the id's being compared were the
following:
- 0.22694672549385309
- 0.22694672549385

The latter number having been truncated after 14 digits, which was a problem
as they were being compared as strings, not floats. By comparing them as floats
we can mitigate this type of rounding / truncating error in the future.

## Motivation and Context
This was found as part of manual on testing XDMoD & SUPREMM on CentOS8. 

## Tests performed
Manual Testing on the metrics-dev-centos8 exploratory machine. 

## Checklist:

### DIFF
--- a/html/controllers/metric_explorer/get_rawdata.php
+++ b/html/controllers/metric_explorer/get_rawdata.php
@@ -72,11 +72,13 @@ try {
 
     $datasetid = $_REQUEST['datasetId'];
 
+    // used to compare to floats, in this case the datasetId and the data_series->id
+    $epsilon = 0.00000000000001;
+
     // find requested dataset.
     $data_description = null;
-
     foreach ($all_data_series as $data_description_index => $data_series) {
-        if ("{$data_series->id}" == "$datasetid") {
+        if (abs($data_series->id - $datasetid) < $epsilon) {
             $data_description = $data_series;
             break;
         }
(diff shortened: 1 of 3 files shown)


jtpalmer:
Do we know why these ID values are floating point?  And can they be replaced with something else?

jtpalmer:
Do we know why these ID values are floating point?  And can they be replaced with something else?

ryanrath:
@jtpalmer So I just went digging for this and the only thing I can figure is that it was to provide a "primary key" to be able to find a particular record in the data series store. If you want to take a look the id is created on the front end in `AddDataPanel.js:87` when you first add a data series to a Metric Explorer Chart. After that it's passed back and forth between the front / back end while working with the data.

jtpalmer:
If a random float is used then replacing it with a random integer should solve that issue.  (It really shouldn't be random, but that's another issue.)  Are these values stored in the database?

ryanrath:
@jtpalmer I'll try replacing the `Math.random()` call w/ `Ext.randomInt(Number.MIN_SAFE_VALUE, Number.MAX_SAFE_VALUE);`. I'm running tests locally now.

ryanrath:
Quick note: All of the changes to CCR.js were due to it being lumped into the files the style check considered to be changed. They are strictly formatting changes.

jtpalmer (approved):

PR #1622, opened 2022-04-21
Switch image comparison algorithm for regression tests.

jpwhite4:
This changes the regression tests for the usage charts to use
the `average_hash` image comparison algorithm from the imagehash
python module.

There are still two sets of expected results one for Centos 7 and
one for Centos 8. The image hash changes do not impact the SVG
regression tests - so these are different and then hashes themselves
are not completely the same between 7 & 8.

These changes should remove the need to regenerate the expected results when
the XDMoD version number changes.


jpwhite4:
Note that of the 1888 different tests the Centos 7 hashes and Centos 8 hashes are identical for 1813 of them. So ~4% of the hashes are different.

It is possible to change the hash algorithm parameters to make it more permissive to changes, however since the svg stuff has to be different anyway I don't think this is worth the disadvantage of missing actual errors.

eiffel777 (dismissed):

eiffel777 (approved):

PR #1656, opened 2022-06-30
Adding mongodb dependencies to XDMoD

ryanrath:
## Description
- The main function of these changes is to add `mongodb/mongodb` as a composer dependency to XDMoD. 
- Refactored the previous `config->platform->php` parameter to `require->php` due to platform->php only supporting one value, while XDMoD can now be run w/ either 5.4 or 7.2. 
- Added a new `allow-plugins` section to `config` so that `composer install` can be unattended ( it would previously ask interactively if these plugins should be allowed to be installed ). 
- composer.lock was updated to reflect the changes above. 

## Motivation and Context
In our efforts to modernize our supported OS from Centos7 -> Rocky Linux 8 we've run into a few stumbling blocks. One of which is that the previous php mongo library was available and installed via the system package manager ( php-pecl-mongo ) while it is not available in Rocky / Centos 8. Another is that the underlying library installed via php-pecl-mongo has been abandoned and superseded by `mongodb` which is available for install only through the PHP Pear Library and requires projects using it to include the `mongodb/mongodb` composer dependency. 

## Tests performed
All automated tests were performed on the latest Centos8 & Rocky 8 Job Performance docker images. 

## Checklist:


.circleci/config.yml
```diff
@@ -64,12 +90,18 @@ jobs:
...
+            - run:
+                name: Setup Environment Variables for centos7
+                command: |
+                  echo "export COMPOSER=composer.json" >> $BASH_ENV
       - run:
           name: Setup & Run QA Tests
-          command: ./tests/ci/scripts/qa-test-setup.sh
-      - run:
-          name: Make sure that the Composer Test Depedencies are installed
-          command: composer install --no-progress
```
jpwhite4, line 72: Looks like there used to be a `composer install` call that has been removed. Was this just never needed in the first place?

.circleci/config.yml
```diff
@@ -64,12 +90,18 @@ jobs:
       - run:
           name: Make sure that the Composer Test Depedencies are installed
           command: composer install --no-progress
+      - when:
+          condition:
+            equal: [ *centos7-executor, << parameters.os >> ]
```
jpwhite4, line 95: Why no corresponding code path for centos8?

.circleci/config.yml
ryanrath: Nope, this is leftover from first moving things to CircleCI and there was some funkiness going on with the test dependencies not being installed. I verified that the deps were being installed in the previous call to `composer install`.

.circleci/config.yml
ryanrath: So it's definitely not obvious from just looking at the config file, which is something that I should change, but basically the qa-test-setup.sh script only does something if CircleCI is performing an `upgrade` which only occurs for CentOS7. I can't honestly remember if I did this pre-knowing how to do use the `when` conditional or if it was something to do with testing current environment variables.

ryanrath: Just keeping track of out of band communications: per our convo I've removed the conditional so that the `COMPOSER` env variable is always reset to `composer.json` here in the latest comits.

PR #1659, opened 2022-07-19
fix: Remove closing PHP tags

zooley:
Remove closing PHP tags

## Description
Remove closing PHP tags in files that only contain PHP. PHP blocks intermixed with HTML retain their closing tags.

## Motivation and Context
If a file contains only PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

Refs: https://www.php.net/manual/en/language.basic-syntax.phptags.php


zooley:
Looks like the full build check got stuck. Anyway to re-launch? I could possibly add a throw-away commit to the PR to see if that re-triggers it.

jpwhite4 (approved):

PR #1660, opened 2022-07-20
fix: Use correct scope when calling method

zooley:
Use correct scope when calling destroy() method

## Description
Change the scope from a global function to local class method.

## Motivation and Context
`__destruct`  method referenced a seemingly nonexistent global `destroy()` function. It should have been calling its own class' `destroy()` method.


jpwhite4 (approved):
Thanks for contributing to XDMoD

PR #1663, opened 2022-08-11
Voltron Destroy

jpwhite4:
Experiment to evaluate what changes would be necessary to use plotly to render the Job Viewer charts.

Looks pretty promising.


jpwhite4:
Will be replaced with a full implementation later.

PR #1669, opened 2022-08-24
Update Contact Us button for ACCESS XDMoD

connersaeli:
This PR updates the Contact Us button to instead redirect to a configured URL for ACCESS tickets. This only works if the configured URL is present, else it will default to Open XDMoD's functionality.

## Description

1.) Checks if 'tech_support_url' is set in configuration.
2.) If 'tech_support_url' is present, then the Contact Us button redirects to that site instead of the dropdown menu.

Without tech_support_url set:
image

With tech_support_url set:
image
In the screenshot the Contact Us button will redirect to the configured URL.

## Motivation and Context

ACCESS XDMoD does not need the same dropdown menu for this button as Open XDMoD.

https://app.asana.com/0/1202679277743034/1202840692396198

## Tests performed

Tested on my development port.

## Checklist:

### DIFF
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -182,6 +182,19 @@ XDMoD.GlobalToolbar.Roadmap = {
 };
 
 XDMoD.GlobalToolbar.Contact = function () {
+    if (CCR.xdmod.support_url) {
+        return {
+            text: 'Contact Us',
+            tooltip: 'Contact Us',
+            scale: 'small',
+            iconCls: 'contact_16',
+            id: 'global-toolbar-contact-us',
+            handler: function() {
+                window.location.href = CCR.xdmod.support_url
+            };
+        };
+    }
+    
     var contactHandler = function(){
         XDMoD.TrackEvent('Portal', 'Contact Us -> ' + this.text + ' Button Clicked');
         switch(this.text){
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -191,7 +191,7 @@ XDMoD.GlobalToolbar.Contact = function () {
             id: 'global-toolbar-contact-us',
             handler: function() {
                 window.location.href = CCR.xdmod.support_url
-            };
+            }
         };
     }
     
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -190,7 +190,7 @@ XDMoD.GlobalToolbar.Contact = function () {
             iconCls: 'contact_16',
             id: 'global-toolbar-contact-us',
             handler: function () {
-                window.location.href = CCR.xdmod.support_url;
+                window.open(CCR.xdmod.support_url, "_blank");
             }
         };
     }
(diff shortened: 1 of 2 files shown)


html/index.php
```diff
@@ -263,6 +263,13 @@ function isReferrer($referrer)
         $tech_support_recipient = xd_utilities\getConfiguration('general', 'tech_support_recipient');
         print "CCR.xdmod.tech_support_recipient = CCR.xdmod.support_email = '$tech_support_recipient';\n";
 
+        $tech_support_url = false;
+        try {
+            $tech_support_url = xd_utilities\getConfiguration('general', 'tech_support_url');
+        } catch (exception $ex) {
+        }
+        print "CCR.xdmod.support_url = CCR.xdmod.support_url = '$tech_support_url';\n";
```
jpwhite4, line 271:
```suggestion
        print 'CCR.xdmod.support_url = ' . json_encode($tech_support_url) . ";\n";
```

You had a typo in the string: "CCR.xdmod.support_url = CCR.xdmod.support_url".

Also you are going to have to encode the $tech_supprt_url variable to make sure it is valid json. I suggest using json_encode. This will ensure that any special characters in the string are escaped and if the variable is set to `false` then it will get output as javascript `false`. The current code will set a javascript string that has the letters  f a l s e in it. This will evaluate to true!

html/gui/js/CCR.js
```diff
@@ -182,6 +182,19 @@ XDMoD.GlobalToolbar.Roadmap = {
...
 XDMoD.GlobalToolbar.Contact = function () {
+    if (CCR.xdmod.support_url) {
+        return {
+            text: 'Contact Us',
+            tooltip: 'Contact Us',
+            scale: 'small',
+            iconCls: 'contact_16',
+            id: 'global-toolbar-contact-us',
+            handler: function () {
+                window.location.href = CCR.xdmod.support_url;
```
jpwhite4, line 193: Does this open in a new window or in the current one?

jpwhite4 (changes requested):

html/gui/js/CCR.js
connersaeli: This opens in the same window.

html/index.php
jpwhite4: image

html/gui/js/CCR.js
jpwhite4: Can to change this to open in a new tab (like we do with the user manual). Often you want to add screenshots when submitting a support ticket and you can't easily do that if the portal is gone

connersaeli: Ok no problem

jpwhite4: looks like `window.open("user_manual.php");` is used for the user manual code

html/index.php
connersaeli: Ok I see your point about using json_encode. Just to clarify, this is to also handle the case where tech_support_url is set as 'false' in the configuration file?

html/gui/js/CCR.js
connersaeli: Just added that change

html/index.php
jpwhite4: The scenario I was thinking of was when `tech_support_url` is absent from the file. Having thought about it I realize that php boolean variable false will evaluate to empty string in your code, which is javascript falsy. So this case won't happen. 

But certainly to need to make sure that you encode the `$tech_support_url` variable properly

jpwhite4 (approved):

PR #1670, opened 2022-08-24
Update text description of XDMoD.

jpwhite4:
Change the main link to the ACCESS instance.


connersaeli (approved):
Will www.xsede.org still be available for the lifetime of ACCESS?

PR #1677, opened 2022-10-06
Add ability to for modules to specifiy report templates.

jpwhite4:
The report template acls were controlled by the contents of one file
 (`etc/etl/etl_data.d/acls/report-template-acls-staging.json`) and the
acl data base was loaded with the file contents each time acl-config was run. Since Open XDMoD modules cannot modify existing files, there was no mechanism for a module to add a new report template acl.

This adds a subdirectory (`report-template-acls-staging.d`) that is scanned for files that will be appended to the report acl templates table. Open XDMoD modules can just add their settings to a file in this directory and it will be loaded when acl-config runs.

It also fixes a problem with the enum_templates controlller endpoint where it would return a json object rather than a json array if the DashBoard report templates were not at the end of the list of templates. This situation will occur if a module is adding report templates since they get added at the end of the list (after the Dashboard reports).

### DIFF
--- a/classes/ETL/Utilities.php
+++ b/classes/ETL/Utilities.php
@@ -229,7 +229,7 @@ class Utilities
         return $localVariableMap;
     }  // quoteVariables()
 
-    public static function runEtlPipeline(array $pipelines, $logger, array $params = array())
+    public static function runEtlPipeline(array $pipelines, $logger, array $params = array(), $module = 'xdmod')
     {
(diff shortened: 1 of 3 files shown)


jpwhite4:
Note - there will be a pull request into the ondemand module to confirm that report templates in modules work properly.

ryanrath (approved):
LGTM

PR #1679, opened 2022-10-19
Improve legend text generation algorithm

jpwhite4:
Previously timeseries plots with multiple dataset would always append the units for the legend text. This results in lots of redundant text if multiple dataset with the same units are plotted.

This pull request changes the code to only show the units if it actually is useful to show (i.e. if the units are different between the datasets or if  the 'single y-axis' config setting is set).

Before:

image

After:

image


connersaeli (approved):

connersaeli:
Just a quick question, the attribute in the class constructor is `share_y_axis`, but in your changes it accesses `self->_shareYaxis`. Is there something I am missing here?

connersaeli:
Never mind, I found the answer.

connersaeli (approved):

PR #1685, opened 2023-01-23
Add vm state dimension

eiffel777:
This PR adds the ability to assign the state of a VM to a row in `modw_cloud.session_records`. This will help us to add data about non-active VM's to the Cloud Realm. A new dimension table has been added `modw_cloud.instance_state` that lists the different states a VM can be in. Active, Inactive and Shelved are the default states but others can be added in one chooses to. The state of a VM is tied to an event type and as such a column is added to `modw_cloud.event_types` called `instance_state_id` that lists the state of the instance when that event occurs. Some of the events in the table are not connected to instances, such as volume creation, deletion, etc. These instance state of these events are NULL.

The RESIZE event is also added to the list of state change events. Doing so allows us to more accurately track how the details of a VM, specifically the number of VCPU's and Memory, change over time, including when a VM is inactive.

## Motivation and Context
Both of these changes helps us to better track information about inactive VM's

## Tests performed
Tested in docker

## Checklist:


classes/OpenXdmod/Migration/Version1000To1050/DatabasesMigration.php
```diff
@@ -19,5 +19,21 @@ public function execute()
     {
         parent::execute();
 
+        $dbh = DB::factory('datawarehouse');
```
jpwhite4, line 22: Does this need to be extra php code to call the pipelines? If you put an action in the `configuration/etl/etl.d/xdmod-migration-10_0_0-10_5_0.json` file then it will get run automatically.

eiffel777: @jpwhite4 This is used precisely because we do not want these pipelines to run automatically. These migrations should only run on installs that are using the Cloud realm.

jpwhite4 (dismissed):

PR #1694, opened 2023-02-21
Add API token GUI components to "My Profile" window.

aaronweeden:
## Description

This adds GUI components that handle fetching, generating, and deleting API tokens. The back end functionality was implemented in https://github.com/ubccr/xdmod/pull/1695.

### API Token tab

The "My Profile" window has a new "API Token" tab:

Screenshot of the new API Token tab in the My Profile window

The contents of the tab show the current state of the user's API token.

### Fetching the token

When the "My Profile" window is opened, a GET request is made to the `/users/current/api/token` REST endpoint.

If the token exists and has not yet expired, the user is shown the creation date and expiration date of the token. The user is encouraged to delete their token and generate a new one if they have lost it, and they are shown a button to delete the token:

Screenshot of the API Token tab with an unexpired token

If the token exists but has expired, the user is shown the expiration date of the token and is asked to delete it and generate a new one:

Screenshot of the API Token tab with an unexpired token

If no token exists, the user is shown a button to generate a new one:

Screenshot of the API Token tab with no token

If there was a server error fetching the token, the user is shown a button to try again:

Screenshot of the API Token tab showing a server error fetching the token

### Generating the token

If the user clicks the generate button, a POST request is made to the `/users/current/api/token` REST endpoint. If successful, the response contains the text of the token as well as the expiration date, both of which are shown to the user (the window is also resized to fit the text). The user is encouraged to copy the text, and the user is shown a button to copy the text to the system clipboard and a button to delete the token:

Screenshot of the API Token tab showing a new token

If a server error occurred while generating the token, the user is shown an error message popup and is encouraged to try again:

Screenshot of the creation error popup

If the user clicks the "Copy to Clipboard" button, the token text is copied to the system clipboard, and a "Copied!" message appears:

Screenshot of the Copied message

### Deleting the token

If the user clicks the delete button, they are prompted to confirm that they want to delete the token:

Screenshot of the deletion confirmation popup

If the user confirms the deletion, a DELETE request is made to the `/users/current/api/token` REST endpoint. If successful, the user is shown that they have no token. If a server error occurred while deleting the token, the user is instead shown an error message popup and is encouraged to try again:

Screenshot of the deletion error popup

## Motivation and Context

For the new Data Analytics Framework, we are going to be allowing users to connect to XDMoD using an API token. Users need a way to obtain their API token through the GUI.

## Tests performed

1. Set up the test environment.
    1. New environment:
[...]

PR #1695, opened 2023-02-28
API Token Framework

ryanrath:
## Motivation and Context
To support the new Data Analytics Framework authentication / authorization needs this PR adds API Tokens to XDMoD, these tokens are meant to allow programmatic access to very specific endpoints which have also been added. An API Token can only be generated by a traditionally authenticated user and a user may only have one active at any one time. Users will be able to interact with the new API Token framework via the `My Profile` portion of the UI being implemented in https://github.com/ubccr/xdmod/pull/1694 . 

## Tests performed
Integration tests have been added for all new endpoints. This includes the API Token management endpoints as well as the endpoints that can only be accessed w/ a valid API Token. 

## Checklist:


classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,118 @@
...
+use Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException;
+use XDUser;
+
+/**
+ *
+ */
+class Tokens
+{
+
+    const TOKEN_FORMAT = '|^([\w=]+)\.([a-zA-Z0-9/.$]+)$|';
```
jpwhite4, line 18: Is this set of characters guaranteed to be appropriate for the hash algorithm? I can't see any documentation that lists the set of characters that are allowed to be in the hash (and hence I'd assume that any characters are valid).

classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,118 @@
...
+Expected User Id:         %s
+User Id:                  %s
+TXT;
+
+            throw new BadRequestHttpException(
+                sprintf(
+                    $msg,
+                    var_export($matches, true),
+                    var_export($encodedExpectedUserId, true),
+                    var_export($expectedUserId, true),
```
jpwhite4, line 73: Since the 'expected*' variables are our private data I don't think we should be outputting them in error execpetions. This seems like it could be used by a malicious actor to guess valid usernames.

classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,118 @@
...
+        // because I hate having variable declaration as a side effect of providing function arguments.
+        $matches = array();
+
+        // Check that the token is in a valid form
+        $match = preg_match(Tokens::TOKEN_FORMAT, $rawToken, $matches);
+        if ($match === 0 || $match === false || count($matches) !== 3) {
+            throw new Exception('A valid API token must be supplied to use this endpoint.');
+        }
+
+        $userId = html_entity_decode(base64_decode($matches[1]));
```
jpwhite4, line 113: I'm not sure that html_entity_decode() belongs here. Please explain why its needed and where the corresponding html_entity_encode() is done?

classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -3,6 +3,8 @@
 namespace Rest\Controllers;
 
 use DateTime;
+use Exception;
```
jpwhite4, line 6: How come this has been added? I can't see any new code that throws it (in any case we should prefer not to throw Exception and instead use one of the specfic ones.

classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -177,7 +180,7 @@ public function setupAssertions(Application $app, ControllerCollection $controll
      *                        defaults to true.
      * @return array of all of the required params and their values as supplied
      *                         by the Request object.
-     * @throws \Exception if not all of the $requiredParams were found.
+     * @throws Exception if not all of the $requiredParams were found.
```
jpwhite4, line 183: Don't think this change should be made.

classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -15,6 +17,7 @@
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
+use XDUser;
```
jpwhite4, line 20: How come we need to use XDUser here? I don't see any changes that use it

classes/Rest/Controllers/UserControllerProvider.php
```diff
@@ -113,12 +120,131 @@ public function updateCurrentUser(Request $request, Application $app)
...
+            }
+
+            $tokenData = $this->getCurrentAPITokenMetaData($user);
+
+            return $app->json(array(
+                    'success' => true,
+                    'data' => $tokenData
+                )
+            );
+        } catch (\Exception $e) {
```
jpwhite4, line 154: You don't need to do this - the rest stack already has code that traps unhandled exceptions and returns the message as json

classes/Rest/Controllers/UserControllerProvider.php
```diff
@@ -205,4 +331,140 @@ private function updateUser(XDUser $user, array $updatedProperties)
...
+        $hash = password_hash($password, PASSWORD_DEFAULT, array('cost' => 12));
+
+        // This token will be stored in the db and will be what's used to verify against $password for authentication / authorization.
+        $storedToken = sprintf('%s.%s', $encodedUserId, $hash);
+
+        // This token will be what's returned to the user.
+        $userToken = sprintf('%s.%s', $encodedUserId, $password);
+
+        $createdOn = date_create()->format('Y-m-d H:m:s');
+        $expirationInterval = \xd_utilities\getConfiguration('api_token', 'expiration_interval');
```
jpwhite4, line 426: Lots of extra spaces?

classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,118 @@
...
+        $row = $db->query('SELECT token, expires_on FROM moddb.user_tokens WHERE user_id = :user_id', array(':user_id' => $userId));
+
+        if (count($row) === 0) {
+            throw new BadRequestHttpException('Malformed token.');
+        }
+
+        $expectedToken = $row[0]['token'];
+        $expiresOn = $row[0]['expires_on'];
+
+        $match = preg_match(self::TOKEN_FORMAT, $expectedToken, $matches);
```
jpwhite4, line 49: Seems like it would be more computationally efficient to do this check before adding data to the database - not every time we read from the database.

Under what circumstances would we have corrupted token format in the database?

jpwhite4: Also why do a regular expression at all at this point? We know the username since it is a dedicated column in the database. Why not have one column for the username and one for the hash?

classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -800,4 +803,37 @@ protected function getTimestamp($date, $paramName = 'date', $format = 'Y-m-d')
...
+    {
+        // NOTE: header keys cannot contain '_' characters.
+        $rawToken = $request->headers->get('apitoken');
+
+        // if we weren't able to find a header called api_token then try getting it from one of the query parameters.
+        if (empty($rawToken)) {
+            $rawToken = $request->get('apitoken');
+            if (empty($rawToken)) {
+                throw new UnauthorizedHttpException(
+                    'Basic realm="Access to Data Analytics endpoints", charset="UTF-8"',
```
jpwhite4, line 824: Is this a cut-and-paste error?
```
Basic realm="Access to Data Analytics endpoints", charset="UTF-8"
```

jpwhite4 (changes requested):
comments added inline

jpwhite4:
Also the lookup for the token should check that moddb.Users.account_is_active is 1

classes/Models/Services/Tokens.php
ryanrath: This has been removed per our convo this morning.

classes/Models/Services/Tokens.php
ryanrath: Oh definitely, this was part of debugging that should have been removed.

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: It was added by IDE for the `@throws Exception` for the `authenticateToken` function.

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: Sure, this is just related to the addition of `use Exception;`. Since it's now been `use`'d the `\` can be removed.

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: Yeah basically. I've just replaced with with a `BadRequestHttpException`

classes/Rest/Controllers/UserControllerProvider.php
ryanrath: Yeah, I got into the habit when working on the the new rest stuff as I was holding off on a general catch all until I had worked out all the kinks. Obviously no reason to do it here.

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: This is the same as https://github.com/ubccr/xdmod/pull/1695#discussion_r1122315350

classes/Models/Services/Tokens.php
ryanrath: I've re-worked this per our convo this morning.

classes/Rest/Controllers/UserControllerProvider.php
```diff
@@ -205,4 +331,140 @@ private function updateUser(XDUser $user, array $updatedProperties)
...
+        // We need to, when presented with a token know which user it is for. To allow for this the tokens stored in the
+        // db will encode the Users.id value along with the hashed token value. This will mean that some pre-processing
+        // will need to occur when attempting to validate the token, but it alleviates the problem of having to attempt
+        // to match every token in the db to find which user it's meant to authenticate.
+        $password = bin2hex(random_bytes(32));
+        $encodedUserId = base64_encode($user->getUserID());
+        $hash = password_hash($password, PASSWORD_DEFAULT, array('cost' => 12));
+
+        // This token will be stored in the db and will be what's used to verify against $password for authentication / authorization.
+        $storedToken = sprintf('%s.%s', $encodedUserId, $hash);
```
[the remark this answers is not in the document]
ryanrath: This has been re-worked per our convo this morning.

classes/Models/Services/Tokens.php
ryanrath: Updated per our convo this morning.

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: So this function isn't actually being used anymore, I'm just going to remove it.

classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,87 @@
...
+            throw new BadRequestHttpException('The API Token has expired.');
+        }
+
+        // finally check that the provided token matches it's stored hash.
+        if (!password_verify($password, $expectedToken)) {
+            throw new AccessDeniedException('Invalid API token.');
+        }
+
+        // and if we've made it this far we can safely return the requested Users data.
+        return XDUser::getUserByID($userId);
```
jpwhite4, line 64: best to select the userId from the database and use that value in getUserByID call. The `$userId` variable contains untrused user supplied content.

ryanrath: Agreed, implemented in latest commits.

classes/Models/Services/Tokens.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    private static function getRawToken()
+    {
+        // Try to find the token in the `Authorization` header.
+        $headers = getallheaders();
+        if (!empty($headers['Authorization'])) {
+            $authorizationHeader = $headers['Authorization'];
+            if (is_string($authorizationHeader) && strpos($authorizationHeader, Tokens::HEADER_KEY) !== false) {
+                // The format for including the token in the header is slightly different then when included as a get or
+                // post parameter. Here the value will be in the form: `Bearer: <token>`
```
jpwhite4, line 130:
```suggestion
                // post parameter. Here the value will be in the form: `Bearer <token>`
```
I thought that there isn't a colon between the Bearer and the value - please confirm

jpwhite4 (approved):

PR #1696, opened 2023-03-09
Update version number to 10.0.1

jpwhite4:
Release prep for the 10.0.1 version. 

Note I added a CI build to test the 10.0.0 to 10.0.1 path and had to mess with the config a bit because the 10.0 docker has different version of the chromedriver used for the UI tests.


jpwhite4:
Note I manually checked the regression tests for a small number of charts - all were identical to the mk1 eyeball.

ryanrath (approved):

PR #1698, opened 2023-03-09
Prevent unrendered windows from attempting to sync shadows.

aaronweeden:
This prevents windows from attempting to sync their shadows if they have not been rendered.

## Motivation and Context

If the "My Profile" window had a "Role Delegation" tab and was opened, there was a console error caused by the window being told to sync its window shadow before the window had been rendered.

## Tests performed

1. Set up the test environment.
    1. New environment:
        1. Set up my port (9001) on `xdmod-dev` by running the script `update-open-xdmod-portal.php` with the branch `xdmod10.5-xsede`.
        1. Copy in the changed files from this Pull Request.
        1. Log in.
    1. Old environment:
        1. Spin up a Docker container from the image `tools-ext-01.ccr.xdmod.org/xdmod-10.0.0:centos7.9-0.6`.
        1. Upgrade to the latest Open XDMoD 10.5 version.
        1. Merge in the changed files.
        1. Run the `~/bin/services start` command and open a browser to `localhost:8080`.
        1. Log in using center director credentials.
1. Confirm that window shadows still appear properly and no console errors appear for each of the existing calls of `syncWindowShadow()` — for each of the fields in the following windows, erase the field and fill it back in:
      * "Contact Us" -> "Send Message"
      * "Sign In" -> "Don't have an account"
      * "Report Generator" -> "Save As"
      * "My Profile" -> "General"
      * (as center director) "My Profile" -> "Role Delegation" -> select staff member
      * (as admin) "Admin Dashboard" -> "User Management" -> "Create & Manage Users"

## Checklist:

### DIFF
--- a/html/gui/js/CCR.js
+++ b/html/gui/js/CCR.js
@@ -571,7 +571,9 @@ XDMoD.utils.trimOnBlur = function (thisField) {
 XDMoD.utils.syncWindowShadow = function (thisComponent) {
     thisComponent.bubble(function (currentComponent) {
         if (currentComponent instanceof Ext.Window) {
-            currentComponent.syncShadow();
+            if (currentComponent.rendered) {
+                currentComponent.syncShadow();
+            }
             return false;
         }
         return true;


html/gui/js/CCR.js
```diff
@@ -571,7 +571,9 @@ XDMoD.utils.trimOnBlur = function (thisField) {
 XDMoD.utils.syncWindowShadow = function (thisComponent) {
     thisComponent.bubble(function (currentComponent) {
         if (currentComponent instanceof Ext.Window) {
-            currentComponent.syncShadow();
+            if (currentComponent.rendered) {
```
jpwhite4, line 574: What happens if you use `currentComponent.isVisible()` rather than `.rendered`?

aaronweeden: It also works. Is it preferable to use `isVisible()`?

jpwhite4: Not sure what is best - usually you prefer to use the public api functions rather than member variables. But it looks like the rendered member variable is in the public documentation. It largely doesn't matter here because we are not going to be updating to a newer version of ExtJS anyway.

jpwhite4 (approved):

PR #1699, opened 2023-03-10
Update image hashing algorithm

jpwhite4:
Switch image hashing to use the average_hash algorithm. This should mean that the tests will not have to be regenerated every time the version number of xdmod changes since the EXiF metadata is not included in the hash function inputs.


jpwhite4:
Note this is basically a re-done version of these changes https://github.com/ubccr/xdmod/pull/1622. I created a new branch so we could have a clean commit history.

jpwhite4:
Note that I have an extra step in the circle ci build for rocky 8 to downgrade chrome - this step will be removed in the chrome 111 compativility updates are merged into 10.5

ryanrath (approved):

PR #1704, opened 2023-03-22
Add helper script that runs chromium.

jpwhite4:
The feature that we were previously using in chromium_headless v110 and earlier seems to have been removed in v111. This changes the code to use puppeteer-core to interact with chromium.

This change is intended to be a drop in replacement: The portal_settings config that specifies the path to chromium is still used and we package the puppeteer code so no need for post install steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,11 @@ jobs:
             - run:
                   name: install the composer dependencies
                   command: composer install
+            - run:
+                  name: update nodejs version
+                  command: |
+                    dnf module -y reset nodejs
+                    dnf module -y install nodejs:16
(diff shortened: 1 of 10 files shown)


ryanrath (commented):
So the only question I have is would this be the right time to add the regression tests back in?

jpwhite4:
> So the only question I have is would this be the right time to add the regression tests back in?

I've manually regression tested by running the old code, and this code in centos7 and rocky8 in three different dockers then exporting some charts. Then used an online image difference checker tool to confirm the output is the same (https://www.diffchecker.com/image-compare/)

image

ryanrath (approved):

PR #1709, opened 2023-03-30
Merge the 10.0 changes into 10.5

jpwhite4:
There were a couple of merge conflicts that were resolved manually.


jpwhite4:
Note the regression test error is caused by a tiny location change in one element (probably due to the chrome version update) 
image

so really a false positive.

PR #1717, opened 2023-04-19
Changing Core hours statistic to CPU Hours with CPU Hour

eiffel777:
Changing Core hours statistic to CPU Hours with CPU Hour to match the Jobs realm. This makes it so if you plot CPU Hours: Total in the Jobs and Cloud realm in the same chart only one axis is used since the units are the same.

## Tests performed
Tested in docker and on xdmod-dev

## Checklist:

### DIFF
--- a/configuration/datawarehouse.d/ref/Cloud-statistics.json
+++ b/configuration/datawarehouse.d/ref/Cloud-statistics.json
@@ -29,11 +29,11 @@
         "precision": 2
     },
     "cloud_core_time": {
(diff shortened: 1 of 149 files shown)


connersaeli (commented):
Looks good on dev. FYI looks like some of the tests are breaking because they're expecting the old "Core Hours" text.

connersaeli (approved):
There is still one test failing for the Centos 7 upgrade, but I trust that will get changed. The Rocky 8 failure looks like it might be a one-off.

connersaeli (dismissed):

connersaeli (approved):

PR #1721, opened 2023-04-20
Create REST endpoints for getting raw data.

aaronweeden:
## Description

This introduces two new REST endpoints, both of which require a valid API token to access:
- `/rest/warehouse/raw-data` — this endpoint responds to `GET` requests, accepts the following query parameters, and responds with a JSON object containing the relevant rows of raw data from the data warehouse and the list of corresponding field labels.
    - Required parameters:
        - `start_date` — start of date range for which to get data.
        - `end_date` — end of date range for which to get data.
        - `realm` — data realm for which to get data.
    - Optional parameters:
        - `fields` — list of aliases of fields to get. If not included, all fields are gotten.
        - `filters` — mapping of dimension names to their possible values. If included, results will only contain fields for which the values for each of the given dimensions match one of the corresponding given values.
        - `offset` — starting row index of data to get.
- `/rest/warehouse/raw-data/limit` — this endpoint responds to `GET` requests and responds with a JSON object containing the maximum number of rows that can be returned in a single response from the `raw-data` endpoint. This value is specified in `portal_settings.ini` under the `datawarehouse` section as `rest_raw_row_limit` (default `10000`).

This changes `BatchDataset` to be able to be constructed with specified fields, a limit, and an offset.

This also deletes some unused variables.

## Motivation and Context

These endpoints are used by the [Python data analytics framework](https://github.com/ubccr/xdmod-python).

## Tests performed

This adds integration tests for the new endpoints.

To test the documentation changes, I copied the files `docs/xdmod-rest-schema.json` and `docs/rest.html` to a web server and confirmed they look correct.

## Checklist:


classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -2181,4 +2191,265 @@ private function getUserStore(\XDUser $user, $realm)
...
+        $className = $realmManager->getRawDataQueryClass($params['realm']);
+        $query = new $className(
+            [
+                'start_date' => $params['start_date'],
+                'end_date' => $params['end_date']
+            ],
+            'batch'
+        );
+        $query = $this->setRawDataQueryFilters($query, $params);
+        if ('SUPREMM' === $params['realm']) {
```
jpwhite4, line 2338: Ideally we shouldn't hardcode optional module information into the core XDMoD. The rawstatistics file can be used if we need a realm-specficit configuation.

jpwhite4 (changes requested):

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -730,7 +741,7 @@ public function getRealms(Request $request, Application $app)
         $user = $this->authorize($request);
 
         // Get parameters.
-        $queryGroup = $this->getStringParam($request, 'querygroup', false, self::_DEFAULT_QUERY_GROUP);
+        $this->getStringParam($request, 'querygroup', false, self::_DEFAULT_QUERY_GROUP);
```
jpwhite4, line 744: This seems odd to me. Please can you check this change is what you intended to do.

jpwhite4 (changes requested):

classes/Rest/Controllers/WarehouseControllerProvider.php
aaronweeden: phpcs caught that `$queryGroup` was an unused variable; it appears it became unused in the commit https://github.com/ubccr/xdmod/commit/b72a5875f6218fd229303145c7e60919b5a73a80 and was never removed. Since there seem to be no useful side effects to this call to `getStringParam()`, I am going to remove the line all together.

jpwhite4 (approved):

PR #1728, opened 2023-05-18
Fixes for warnings generated with the supremm module tests

jpwhite4:
The xdmod-supremm module build has been updated to check for errors/warnings. Needless to say I found some.

The default value for the font size must be a numeric string - and not a text string. Also an error in the updateHistory rest endpoint. (these two changes have tests in the supremm module that check them).

Add some checks for data validity and hoy out an execption if the checks fail.

The corresponding pull request in the supremm module is https://github.com/ubccr/xdmod-supremm/pull/342

### DIFF
--- a/classes/DataWarehouse/Access/Common.php
+++ b/classes/DataWarehouse/Access/Common.php
@@ -45,6 +45,12 @@ class Common
 
     protected function checkDateParameters()
     {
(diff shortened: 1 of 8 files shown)


aaronweeden (commented):
Might be good to also have additional integration tests in the xdmod repository for these changes.

PR #1729, opened 2023-05-19
Integrate Plotly JavaScript for Job Viewer Tab

aestoltm:
## Description

**Timeseries Charts:**
- Timeseries charts are now rendered Plotly js charts instead of HighCharts

**Analytic Charts:**
- Analytic charts are now rendered with Plotly js instead of Highcharts
- Color brightness is slightly different than Highcharts. 

**Peer Charts (Gantt Charts):**

- No official support for gantt charts in Plotly js therefore rendered charts with Plotly shapes and added an underlying scatter plot due to shapes not supporting Plotly events and tooltips. Two other implementations I explored where using horizontal bar charts and box plot charts however the current implementation matches the highcharts version the best.

**Timeseries Drilldown:**
- Adjusted `plotly_click` in `ChartPanel.js` to grab the nodeid of the trace's data point selected for drilldown redirect.
- Adjusted how drilldown works in `GanttChart.js` to include the click event information in each data point and then when `plotly_click` fires `action: 'show'` is added to the click event info and is encoded for the drilldown redirect.

**Export:**
- Updated selector in `chrome-helper.js` to the div in `plotly_template.html`
- Updated layout for export chart to Plotly equivalent in `chartImageResponse()` in `classes/Rest/Controller/WarehouseControllerProvider.php` 
- Updated `charting.php` to grab file content from `plotly_template.html`
- Created `plotly_template.html`. Uses layout and data given from `charting.php` to draw chart a div. 
 
**Known Issues/TODO**
- Fix text cutoff on error text for when there are more than four analytic charts. Plotly does not support any word wrapping, therefore, we can add a [wordwrap](https://github.com/75lb/wordwrapjs) library as this most likely will be needed for when updating other modules with Plotly js.
- Optimize `plotly_hover` and `plotly_unhover` that mimics highcharts point hover events. Currently, not included with the PR but during testing the events crashed the webpage if hovering over charts with many overlapping traces. **Note:** this issue may be included in a future PR as it does not impact functionality

## Motivation and Context

Currently, Highcharts requires a license to be purchased which is solved by using Plotly js an open-source library.

## Tests performed

- Added test cases in `xdmod-supremm` for changes made to Job Viewer
- Performed GUI testing by visually comparing production and dev environment charts.

**TODO**
- Add test cases for exporting Plotly charts
- Add a button on Job Viewer Tab to redirect the reviewer to a random job with timeseries data. 

## Checklist:


background_scripts/chrome-helper/chrome-helper.js
```diff
@@ -20,7 +20,7 @@ const args = require('yargs').argv;
 
     await page.goto('file://' + args['input-file']);
 
-    const innerHtml = await page.evaluate(() => document.querySelector('.highcharts-container').innerHTML);
+    const innerHtml = await page.evaluate(() => document.querySelector('.plotly').innerHTML);
```
jpwhite4, line 23: You can't change this - it will break image export for the rest of XDMoD.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1984,74 +1984,61 @@ private function chartImageResponse($data, $type, $settings)
...
-                    'fontSize' => $mainTitleFontSize
+                    'size' => $mainTitleFontSize
                 ),
 
                 'text' => $settings['show_title'] ? $data['schema']['description'] : null
             )
         );
 
-        if (strpos($data['schema']['units'], '%') !== false) {
+        /*if (strpos($data['schema']['units'], '%') !== false) {
```
jpwhite4, line 2039: Please don't add commented code - I'm sure the linter has already complained about this!

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
```diff
@@ -195,11 +248,39 @@ XDMoD.Module.JobViewer.AnalyticChartPanel = Ext.extend(Ext.Panel, {
...
+		    "align": "left",
+                    "xref": "paper",
+                    "yref": "paper",
+                    "sizex": 0.4,
+                    "sizey": 0.4,
+                    "x": 0,
+                    "y": 1.2
+                }
+            ]
+	    console.log("new annontation");
```
jpwhite4, line 263: Please don't add console.log() statements (by all means use them when debugging - but not in the final code that gets reviewed).

html/gui/js/modules/job_viewer/ChartPanel.js
```diff
@@ -172,18 +92,18 @@ XDMoD.Module.JobViewer.ChartPanel = Ext.extend(Ext.Panel, {
...
-                this.chart.setSize(adjWidth, adjHeight);
+            if (panel.chart) {
+                Plotly.relayout(panel.id, {width: adjWidth, height: adjHeight});
             }
-            this.options.chart.width = adjWidth;
-            this.options.chart.height = adjHeight;
         }, // resize
 
         destroy: function () {
+/*
```
jpwhite4, line 101: What is the new object lifecycle? Before we explicitly called destroy on highcharts charts to free up memory when the tab was closed. How does it work now in your code?

html/gui/js/modules/job_viewer/ChartTab.js
```diff
@@ -99,8 +99,9 @@ XDMoD.Module.JobViewer.ChartTab = Ext.extend(Ext.Panel, {
 
             var chartOptions = jQuery.extend(true, {}, defaultChartSettings, self.chartSettings);
 
-            self.chart = new Highcharts.Chart(chartOptions);
-            self.chart.showLoading();
+            //self.chart = new Highcharts.Chart(chartOptions);
+            //self.chart.showLoading();
+            Plotly.newPlot(this.id, [], {});
```
jpwhite4, line 104: How come you create an empty Plotly plot here?

html/gui/js/modules/job_viewer/ChartPanel.js
aestoltm: `Plotly.purge(graphDiv)` is supposed to be called here. I will add that to destroy.

html/gui/js/modules/job_viewer/ChartTab.js
aestoltm: This is not needed here and will be removed. This was added when I was trying to understand ChartTab.js and GanttChart.js

html/gui/js/modules/job_viewer/JobViewer.js
```diff
@@ -965,6 +965,10 @@ XDMoD.Module.JobViewer = Ext.extend(XDMoD.PortalModule, {
         export_option_selected: function (exportParams) {
             var chartPanel = this.getActiveJobSubPanel();
             if (chartPanel) {
+		var test = document.getElementsByClassName("main-svg");
+                console.log("SVG");
+                console.log(test);
+                console.log(test[test.length-1]);
```
eiffel777, line 971: Is this section just some debugging code that wasn't removed before committing? It looks like the only changes to this file are basically console.log statements.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1984,74 +1984,61 @@ private function chartImageResponse($data, $type, $settings)
...
-                'enabled' => false
+            'annotations' => array(
+                'text' => $data['schema']['source'] . '. Powered by XDMoD/Plotly',
+		'xref' => 'paper',
+		'yref' => 'paper',
+		'x' => 1,
+		'xanchor' => 'left',
+		'y' => 0,
+		'yanchor' => 'top',
+		'showarrow' => false
```
eiffel777, line 2027: This is very nitpicky but it looks like there is differing indentation in a couple of spots in this file. It would be better if there was uniform indentation. The linter might be picking this up but I'm not sure.

html/gui/js/modules/job_viewer/GanttChart.js
```diff
@@ -34,48 +34,138 @@ XDMoD.Module.JobViewer.GanttChart = Ext.extend(XDMoD.Module.JobViewer.ChartTab,
                 return;
             }
             var record = store.getAt(0);
-
+	    console.log("record");
+  	    console.log(record);
```
eiffel777, line 38: More console.log statements that should be removed.

eiffel777 (changes requested):

background_scripts/chrome-helper/chrome-helper.js
```diff
@@ -20,7 +20,15 @@ const args = require('yargs').argv;
 
     await page.goto('file://' + args['input-file']);
 
-    const innerHtml = await page.evaluate(() => document.querySelector('.highcharts-container').innerHTML);
+    const highchartInnerHtml = await page.evaluate(() => document.querySelector('.highcharts-container').innerHTML);
```
jpwhite4, line 23: Suggest that you make the selector string a command line parameter. Something like '--selector' then the helper will not have to hardcode the container string name,

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
```diff
@@ -190,16 +218,43 @@ XDMoD.Module.JobViewer.AnalyticChartPanel = Ext.extend(Ext.Panel, {
...
+                    align: 'left',
+                    xref: 'paper',
+                    yref: 'paper',
+                    sizex: 0.4,
+                    sizey: 0.4,
+                    x: 0,
+                    y: 1.2
+                }
+            ];
+            this._DEFAULT_CONFIG.layout.images = errorImage;
```
ryanrath, line 237: I'm still making my way though the rest of the PR, but does the altering of `this._DEFAULT_CONFIG` have an effect on other instantiations of AnalyticChartPanel? Was that intended? I was just checking out what we did before which kind of looks like: 
- `Ext.apply(this.chartOptions, this._DEFAULT_CONFIG.chartOptions);` 
- `this.colorSteps = Ext.apply(this.colorSteps || [], this._DEFAULT_CONFIG.colorSteps );`

So we were essentially using it to provide the defaults and then further updating `this.chartOptions` or `this.colorSteps` if needed before using. 

Instead of marking each of the lines for this one I'll just list them here: 
- 252 -> 257
- 272
- 289

ryanrath (commented):
Just a couple of questions

html/gui/js/libraries/PlotlyUtilities.js
```diff
@@ -0,0 +1,152 @@
...
+                size: 0.1,
+                color: color
+            },
+            line: {
+                width: 2,
+                color: color
+            },
+            text: qtip,
+            hovertemplate: '%{x|%A, %b %e, %H:%M:%S.%L} ' + tz + '<br> <span style=color:'
+            + color + ';>●</span> ' + record.data.series[sid].name + ': <b>%{y:}</b> <extra></extra>',
```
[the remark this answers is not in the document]
aestoltm: I will fix that. I have not been able to find anything yet that states Plotly JS html encodes the hovertemplate beforehand.

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
aestoltm: This was not intended and should be changed to how it worked before. Fixing this should also clean up the resize handler for Analytic charts because the plot will have independent layouts.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -1964,106 +1964,27 @@ private function chartDataResponse($data)
...
-        foreach ($data['series'] as $series) {
-            // if the series array contains any data array with exactly one element, enable markers
-            // (a dot for each series' element) so that the plotted data can be seen:
-            if (isset($series['data']) && count($series['data']) == 1) {
-                $markerEnabled = true;
-                break;
-            }
-        }
-
         $axisTitleFontSize = ($settings['font_size'] + 12) . 'px';
```
jpwhite4, line 1967: How come this code was left in but the rest of the setting transformation code was moved? Is there something different about the axis font sizes?

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
```diff
@@ -116,17 +112,8 @@ XDMoD.Module.JobViewer.AnalyticChartPanel = Ext.extend(Ext.Panel, {
...
-
-            this.chartOptions.chart.renderTo = this.id;
-
-            if (this.chart) {
-                this.chart.destroy();
-            }
-            this.chart = new Highcharts.Chart(this.chartOptions);
-
+        render: function () {
+            this.chart = Plotly.newPlot(this.id, [], this._DEFAULT_CONFIG.layout, this._DEFAULT_CONFIG.config);
```
jpwhite4, line 116: Does it cause problems passing the same configuration objects to multiple plotly instances? I naively would have done a deep copy on the DEFAULT_CONFIG objects and then passed the copy into the plotly instance.

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
```diff
@@ -190,16 +218,43 @@ XDMoD.Module.JobViewer.AnalyticChartPanel = Ext.extend(Ext.Panel, {
...
+                    yref: 'paper',
+                    font: {
+                        size: 11
+                    },
+                    x: 0.05,
+                    y: 1.2,
+                    showarrow: false
+                }
+            ];
+            this._DEFAULT_CONFIG.layout.annotations = errorText;
```
jpwhite4, line 252: I don't think it is a good idea to be editing the DEFAULT_CONFIG object. it can hardly be a portal-wide default if it changes accordiong to the state of an individual chart

html/gui/js/modules/job_viewer/AnalyticChartPanel.js
```diff
@@ -210,35 +265,33 @@ XDMoD.Module.JobViewer.AnalyticChartPanel = Ext.extend(Ext.Panel, {
...
-            color = this._getDataColor(data.value);
-            nColor = new Highcharts.Color(color).brighten(brightFactor);
-            this.chart.options.chart.plotBackgroundColor = 'rgba(' + nColor.rgba + ')';
-
-            this.chart.addSeries({
+    _updateData: function (data) {
+        var trace = {};
+        if (data.error === '') {
+            var chartColor = this._getDataColor(data.value);
+            this._DEFAULT_CONFIG.layout.plot_bgcolor = chartColor.bg_color;
```
jpwhite4, line 272: Same comment as above - not a good idea to edit the default that is shared among all charts.

ryanrath (commented):
LGTM. I"m good to merge once Joe has a look.

libraries/charting.php
```diff
@@ -41,16 +41,19 @@ function exportHighchart(
...
-    $fileMetadata = null
+    $fileMetadata = null,
+    $isPlotly = false
 ) {
     $effectiveWidth = (int)($width*$scale);
     $effectiveHeight = (int)($height*$scale);
 
     $html_dir = __DIR__ . "/../html";
     $template = file_get_contents($html_dir . "/highchart_template.html");
+    if ($isPlotly){
```
jpwhite4, line 52: Very minor point: if the chart is plotly then the code will open the file called highchart_template.html then allocate memory, then copy the contents of the file into memory, then delete the memory. Needless to say there is no point using cpu cycles to read the highchart_template file when in plotly mode.

Don't need to change this for this pull request. But should address this in the next set of code changes.

jpwhite4 (approved):

PR #1732, opened 2023-05-31
Adding default values for modw_cloud.instance_type_union table

eiffel777:
Adding default values for modw_cloud.instance_type_union table prevents the following warning

```
[warning] Warning 1364 Field 'instance_type_id' doesn't have a default value
```
## Tests performed
Tested in docker

## Checklist:

### DIFF
--- a/configuration/etl/etl_tables.d/cloud_common/instance_type_union.json
+++ b/configuration/etl/etl_tables.d/cloud_common/instance_type_union.json
@@ -16,8 +16,7 @@
             {
                 "name": "instance_type_id",
                 "type": "int(11)",
-                "nullable": false,
-                "default": -1,
+                "nullable": true,
                 "comment": "Resource to which this type belongs"
             },
             {
--- a/configuration/etl/etl_tables.d/cloud_common/instance_type_union.json
+++ b/configuration/etl/etl_tables.d/cloud_common/instance_type_union.json
@@ -10,24 +10,28 @@
                 "name": "resource_id",
                 "type": "int(11)",
                 "nullable": false,
+                "default": -1,
                 "comment": "Resource to which this type belongs"
             },
             {
                 "name": "instance_type_id",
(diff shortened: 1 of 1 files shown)


configuration/etl/etl_tables.d/cloud_common/instance_type_union.json
```diff
@@ -10,24 +10,27 @@
...
                 "type": "int(11)",
                 "nullable": false,
+                "default": -1,
                 "comment": "Resource to which this type belongs"
             },
             {
                 "name": "instance_type_id",
                 "type": "int(11)",
-                "nullable": false,
+                "nullable": true,
```
jpwhite4, line 19: Is this a foreign key to a dimension table? If so then this shouldn't be nullable, instead the default value can be the row index in the dimension table corresponding to 'Unknown' or 'NA' or whatever is the appropriate name.

eiffel777: @jpwhite4 On this table I would say it is not really a foreign key. This table contains a union of rows from modw_cloud.instance_type, that have already been ingested and have an `instance_type_id`, and `modw_cloud.raw_instance_type`, which has not been ingested and have not been assigned an `instance_type_id`. These rows are then passed to the `CloudInstanceTypeIngestor` which updates the end_time for any existing instance type configuration and makes a new row for any new instance type configuration. They are then inserted into `modw_cloud.instance_type`. Having rows that are new instance types have `NULL` allows them to get the auto incremented ID on the `modw.instance_type` table when they are inserted. Setting it to the unknown value wouldn't work. If you want it to not be `NULL` I'd have to designate another value that is not the unknown value, and then I would have to change `CloudStateInstanceTypeIngestor` to recognize that value.

See #1650 for original PR for this functionality.

jpwhite4 (approved):

PR #1735, opened 2023-06-15
Fix "My Profile" window loading bugs.

aaronweeden:
## Description

This fixes bugs in the loading of the "My Profile" window. The bugs are noticeable when the portal has a lot of data to load. 

One of the bugs is that the "API Token" window sometimes does not advance past its "Loading..." message, particularly if you click the "My Profile" button very quickly after the page initially starts loading.

Another category of bugs involves clicking the "My Profile" window multiple times before the portal data has finished loading. For example, on `xdmod-dev`, while logged in as a Program Officer / Manager / Developer, reloading the portal uncached in the browser, and clicking the "My Profile" button multiple times in rapid succession can lead to bugs.

One of the bugs in this category is that once the window loads, if you switch to another tab in the window, the contents of that tab get appended to the end of the current tab rather than showing in their own tab. An example screenshot is below, showing the contents of the "API Token" tab underneath the contents of the "General" tab.

2023-06-23-my-profile-bug-0

Another bug in this category is that if you move the window, its shadow does not follow it:

2023-06-23-my-profile-bug-1

Another bug in this category is that if you close the window, you can no longer interact with the portal; it is grayed out.

The reason why these bugs occur is that each click of the button causes a new window to be initialized with its own shadow and modal background; however, only one window is displayed and can be interacted with. Closing that one window does not cause the other instances of the modal background to close, and moving that one window does not cause the shadows of the other windows to move.

The API token "Loading..." bug is caused by a race condition. When the "My Profile" button is clicked, requests are made to obtain data for all of the window's tabs. Once the "General" tab has its response, the window appears, and the DOM elements are filled. However, if the "API Token" tab receives its data first, it will try to fill the DOM elements without waiting for the window to load, and it will fail to do so.

## Solution
The solution implemented here is to immediately open the window as soon as the "My Profile" button is clicked, which places the button behind the modal window such that it cannot be clicked again until the window is closed.

Instead of having the window fill its DOM elements as soon as it opens, the DOM elements are initially set to be disabled and have "Loading" messages; then, once the data is received, the DOM elements are filled. However, if the window has closed before the data arrives, the DOM elements are not filled, since those elements no longer exist. If the data arrives before the window has opened, the DOM elements are filled once the window opens.

## Motivation and Context

The Data Analytics Framework involves generating an API token, and if there is enough data being loaded that there is a delay in the "My Profile" window opening, users may naturally click the button again if it doesn't respond right away.

## Tests performed

Manual tests were performed on my port on `xdmod-dev` (9001) logged in as a Program Officer / Manager / Developer, which has a lot of dashboard data to load, while checking browser console for errors, refreshing between each test, and doing each test once looking at the General tab and once looking at the API Token tab:
- Test 1:
    - Click the My Profile button.
    - Wait for the data from the dashboard and the My Profile window to load.
    - Close the My Profile window.
- Test 2:
    - Click the My Profile button.
    - Immediately close the My Profile button.
    - Wait for the data from the dashboard to load.
    - Click the My Profile button.
    - Wait for the data from the My Profile window to load.
    - Close the My Profile window.
- Test 3:
    - Click the My Profile button.
    - Immediately close the My Profile button.
    - Immediately click the My Profile button.
    - Wait for the data from the dashboard and the My Profile window to load.
    - Close the window.

## Checklist:


html/gui/js/profile_editor/ProfileGeneralSettings.js
```diff
@@ -163,6 +166,10 @@ XDMoD.ProfileGeneralSettings = Ext.extend(Ext.form.FormPanel, {
         // ------------------------------------------------
 
         this.cbProfile = function (options, success, response) {
+            // If the window is not open, do not attempt to manipulate the DOM.
+            if (!self.parentWindow.isVisible()) {
+                return;
+            }
```
aaronweeden, line 172: May be better to wrap the code below in an window loaded event handler.

html/gui/js/profile_editor/ProfileEditor.js
```diff
@@ -14,7 +14,7 @@ XDMoD.ProfileEditorConstants = {
 
 XDMoD.ProfileEditor = Ext.extend(Ext.Window, {
     id: 'xdmod-profile-editor',
-    width: 375,
+    width: 400,
```
aaronweeden, line 17: This change was added because it was noticed during testing that the window is not wide enough to fit all five tabs at once.

html/gui/js/profile_editor/ProfileGeneralSettings.js
```diff
@@ -315,7 +350,7 @@ XDMoD.ProfileGeneralSettings = Ext.extend(Ext.form.FormPanel, {
                 user_profile_lastname,
                 user_profile_email_addr,
 
-                lblRole
+                user_profile_most_privileged_role
```
aaronweeden, line 353: This variable was renamed to be consistent with others.

html/gui/css/ProfileEditor.css
```diff
@@ -44,6 +44,11 @@
    width: 125px !important;
 }
 
+#user_profile_most_privileged_role {
+    font-weight: bold;
+    padding-left: 3px;
+}
+
```
aaronweeden, line 51: This was added to replace inline CSS used in `html/gui/js/profile_editor/ProfileGeneralSettings.js`.

tests/ui/webdriverHelpers/waitForAllInvisible.js
```diff
@@ -2,7 +2,7 @@
  *  Wait for all elements that match a selector to be invisible.
  *
  *  @param {string} selector - elements to check
- *  @param {Number} [ms=9000] - Milliseconds to wait for elements to be invisible
+ *  @param {Number} [ms=18000] - Milliseconds to wait for elements to be invisible
```
aaronweeden, line 5: During debugging I noticed this comment was not consistent with the code below.

jpwhite4 (dismissed):

jpwhite4 (approved):

PR #1739, opened 2023-07-07
Update styling of error message to match highcharts version.

jpwhite4:
Changes the icon to better match the original one remove the bold font and set the font size and face to match the original.
This does not address the word wrap (which will be in a different pull request).

HighCharts original version:
image

Ploty version with this code change:

image

Original plotly version:

image


aestoltm (commented):
LGTM. I see that this addresses the error text and error icon scrunching together when the window is resized to a smaller viewport.

PR #1740, opened 2023-07-07
Add word wrap to job viewer analytics error messages

jpwhite4:
Before:
image

After (this also includes the styling changes in #1739 
image

Note this depends on a pull request in the xdmod-supremm repo that changes the error output to not add html entitires.


jpwhite4:
Note that the text does not get re-wrapped on a window size change. The fix for that would be a bit more intrusive.

aestoltm (approved):

PR #1749, opened 2023-08-03
Handle case where filters contain a quote character

jpwhite4:
## Description
Update the filter parsing code to handle case where a filter value has a quote character in it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,16 +108,16 @@ jobs:
           command: |
             echo "export COMPOSER=composer.json" >> $BASH_ENV
       - run:
-          name: Setup & Run QA Tests
-          command: |
-            ./tests/ci/scripts/qa-test-setup.sh
-      - run:
           name: Setup Configuration Files for Integration Tests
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -107,17 +107,17 @@ jobs:
...
-          command: |
-            ./tests/ci/scripts/qa-test-setup.sh
       - run:
           name: Setup Configuration Files for Integration Tests
           command: |
             mv ./configuration/organization.json ./configuration/organization.json.old
             mv ./configuration/portal_settings.ini ./configuration/portal_settings.ini.old
             cp /etc/xdmod/portal_settings.ini ./configuration/portal_settings.ini
             cp /etc/xdmod/organization.json ./configuration/organization.json
+      - run:
```
jpwhite4, line 117: This has to run after the configs are copied over because one of the so-called 'unit' tests is actually a component test that needs the PDO data base handle to exist

ryanrath (approved):

PR #1755, opened 2023-08-09
Add information about older non-supported XDMoD versions [skip ci]

jpwhite4:


aaronweeden:
I would suggest the headings for "Older versions" and "Other versions" be changed to "Older supported versions" and "Other supported versions."

PR #1756, opened 2023-08-09
Remove jQuery dependency

aestoltm:
## Description

Removing jQuery dependency by replacing `jQuery.extend()` with solutions provided with modern browsers. This PR is a continuation of PR #1542.

Dependent on PR #1785 so qa-tests will not receive a parsing error on arrow syntax. 

## Motivation and Context

jQuery is a large dependency that we use very sparingly. 

## Checklist:

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204876410658973


html/gui/js/HighChartWrapper.js
```diff
@@ -2,6 +2,36 @@
 
 Ext.namespace('XDMoD.utils');
 
+// Taken from: https://github.com/HubSpot/YouMightNotNeedjQuery
```
jpwhite4, line 5: Please can you add the license info about this function into the comments too - I think you said it was MIT license?

aestoltm: Updated license information in NOTICE, notices.md, and the comment mentioned here.

html/gui/js/HighChartWrapper.js
```diff
@@ -2,6 +2,37 @@
 
 Ext.namespace('XDMoD.utils');
 
+// Taken from: https://github.com/HubSpot/YouMightNotNeedjQuery, which
```
jpwhite4, line 5: I think this function belongs with the other utility functions in CCR.js. Seems to me a file called 'HighChartWrapper' should just contain functions related to wrapping highcharts.

html/gui/js/AddDataPanel.js
```diff
@@ -305,7 +305,7 @@ Ext.extend(CCR.xdmod.ui.AddDataPanel, Ext.Panel, {
             ])
         });
         if (this.record.data.filters) {
-            var currentFilters = jQuery.extend({}, this.record.data.filters);
+            var currentFilters = { ...{}, ...this.record.data.filters }; // eslint-disable-line
```
jpwhite4, line 308: Please don't disable eslint for one line - either the line is good and the rule can be added to eslint or the line is bad in which case don't do it!

jpwhite4 (changes requested):

html/gui/js/HighChartWrapper.js
aestoltm: Moved to CCR.js with the other utils functions

html/gui/js/AddDataPanel.js
aestoltm: With updates to xdmod-qa these lines should now be parsed correctly without needing to disable them.

html/gui/js/modules/Summary.js
```diff
@@ -638,8 +638,7 @@ Ext.extend(XDMoD.Module.Summary, XDMoD.PortalModule, {
                             t: 22,
                             b: 0
                         }
-                    },
-                    data: []
```
jpwhite4, line 642: This looks like a mistake or a code change for a different purpose. Please can you double check.

html/gui/js/modules/Usage.js
```diff
@@ -2718,14 +2717,14 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
-                            jQuery.extend(true, chartOptions, baseChartOptions);
+                            chartOptions = XDMoD.utils.deepExtend({}, baseChartOptions, chartOptions);
                             chartOptions.exporting.enabled = false;
                             chartOptions.credits.enabled = true;
 
                             this.chart = XDMoD.utils.createChart(chartOptions);
                             this.chartId = id;
                             this.chartOptions = chartOptions;
-                            var chartDiv = document.getElementById(baseChartOptions.renderTo);
+                            var chartDiv = document.getElementById(chartOptions.renderTo);
```
jpwhite4, line 2727: Please can you double check that this should be changed here.

html/gui/js/modules/Usage.js
```diff
@@ -1510,7 +1510,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                                 if (chartRecords.length > 0) {
                                     chartOptions = chartRecords[0].get('hc_jsonstore');
                                 }
-                                jQuery.extend(true, chartOptions, baseChartOptions);
+                                chartOptions = XDMoD.utils.deepExtend({}, baseChartOptions, chartOptions);
```
jpwhite4, line 1513: The function arguments are in a different order here. Please can you check this is intended (elsewhere you have kept the same order).

html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -756,9 +756,9 @@ Ext.apply(XDMoD.Module.MetricExplorer, {
                                 if (datasetCount === 1) {
                                     instance.filtersStore.add(new instance.filtersStore.recordType(filter));
                                 } else if (datasetCount > 1) {
-                                    var filters = jQuery.extend(true, {}, record.get('filters')),
-                                        found = false;
-                                    for (var i = 0; i < filters.length; i++) {
+                                    var filters = XDMoD.utils.deepExtend({}, record.get('filters'));
+                                    var found = false;
```
jpwhite4, line 760:
```suggestion
                                    let found = false;
```

jpwhite4 (changes requested):

html/gui/js/modules/Summary.js
aestoltm: This was leftover from the initial implementation of the Plotly conversion when I was trying to figure out a timing issue with the relayout events. jQuery the empty data array would not overwrite the data field in the target object. However, the native JS approach will overwrite the data loaded in from `chartOptions = chartRecords[0].get('hc_jsonstore'); ` when called with the original ordering  `chartOptions = XDMoD.utils.deepExtend({}, chartOptions, baseChartOptions);`. We want to keep the original ordering but remove the unnecessary empty data array here and in `Usage.js.`

jpwhite4 (approved):

PR #1759, opened 2023-08-18
Fix raw data filtering to use fact tables rather than aggregate tables.

aaronweeden:
Abandoned in favor of fixing raw data filtering to use aggregate tables.


aaronweeden:
Abandoned in favor of fixing raw data filtering to use aggregate tables.

PR #1762, opened 2023-08-23
Fix zoom resetting to incorrect initial range values

aestoltm:
## Description

When double clicking to reset the zoom level on peer jobs Plotly would reset the range for the axis to values different than the ones set manually. This is due to the chart being created in `ChartTab.js` initially and then being updated in `GanttChart.js`. The Plotly zoom reset functionality checks for the initial range created but we set it based on the data given in `GanttChart.js,`. This change fixes this by recreating the original plot when we receive data instead of updating it.

There is another [solution ](https://community.plotly.com/t/reset-axis-after-layout-update/4335) by manipulating the chart div during a `doubleClick` event to set the initial range correctly. This could be a better solution due to `Plotly.newPlot` being slower than using `Plotly.react`. 

NOTE: This behavior isn't observed in the timeseries charts because we have our chart settings configured before the graph is created for the first time.

## Motivation and Context

The original implementation of the peer jobs had the xaxis range set to the current peer job walltime. Plotly auto-sizing the range during zoom reset breaks this original implementation 

## Tests performed

Tested on xdmod-dev

## Checklist:


html/gui/js/modules/job_viewer/GanttChart.js
```diff
@@ -146,7 +146,7 @@ XDMoD.Module.JobViewer.GanttChart = Ext.extend(XDMoD.Module.JobViewer.ChartTab,
             };
 
             layout.shapes = rect;
-            Plotly.react(this.id + '_hc', data, layout, { displayModeBar: false, doubleClick: 'reset' });
+            Plotly.newPlot(this.id + '_hc', data, layout, { displayModeBar: false, doubleClick: 'reset' });
```
jpwhite4, line 149: Just want to confirm that there isn't a memory leak when you do this. Is it necessary to call a destroy() function (or equivalent) on the old plot? or does the plotly library handle that?

aestoltm: Plotly library handles destroying the existing plot before recreating it.  As seen here https://github.com/plotly/plotly.js/blob/630b092a3878942a61f8fd7b08ac1b98a1abe86e/src/plot_api/plot_api.js#L564

jpwhite4 (approved):

PR #1706, opened 2023-03-23
Refactor some of the integration tests.

aaronweeden:
## Description

This refactors some of the integration tests to make the code simpler, less duplicated, and more consistent. Specifically,
- Change `TokenHelper` to be a test class called `TokenAuthTest` that can be extended to run a suite of token authentication tests on an endpoint: empty token, malformed token, invalid token, valid token (one or more tests), expired token, and revoked token. This improves the test coverage of the `Tokens::authenticateToken()` method.
- Move the `makeRequest()` method from `TokenHelper` into a new `requestAndValidateJson()` method in `BaseTest`. This provides a method that any test can reuse.
- When expiring tokens, set a time in the past rather than the current time. This fixes a bug in which a test could run fast enough that the difference between the current and past times was too insignificant to be detected.
- Move API token endpoint tests from `UserControllerProviderTest` into the test classes for the controllers that implement those endpoints, `MetricExplorerTest` and `WarehouseExportControllerProviderTest`. This keeps tests closer to the endpoints they are testing.
- Use the existing `Configuration` functionality to parse JSON test artifact files, adding an `IntegrationTestConfiguration` class that allows files to refer to paths relative to the integration test artifacts root directory and allow JSON schemas to have references to keys in the same document (e.g., `"$ref": "#/foo"`).
- Add a `dataProvider` to `BaseTest` that simply provides the list of roles, which is sufficient for many tests.
- Move JSON file loading and validation into a `validateJsonAgainstFile()` method since the tests that called the `validateJson` method all involved loading a file first. This reduces duplicated code.
- Add the [`swaggest/php-json-schema`](https://github.com/swaggest/php-json-schema) library to enable validating JSON using JSON Schema Draft 7. The existing [`justinrainbow/json-schema`](https://github.com/justinrainbow/json-schema) library only supports up through JSON Schema Draft 4. For example, the `contains` array property can now be validated; it was introduced in JSON Schema Draft 6.
- Make the expected "unauthorized" error code be only 401 since el7 does not actually produce 400.
- Move `getTestFiles()` into `BaseTest` to avoid multiple files reimplementing it.
- Make the `getRealms()` method of `BaseTest` be `protected` instead of `public` since it doesn't need to be called outside of its class hierarchy.
- Rename classes to match the names of the classes they are testing (e.g., `WarehouseExportControllerTest` becomes `WarehouseExportControllerProviderTest`).
- Fix warnings reported by `phpcs`.

## Motivation and Context

[...]

PR #1763, opened 2023-08-24
Refactor some of the integration tests.

aaronweeden:
## Description

This PR refactors some of the existing integration tests (a further refactoring from https://github.com/ubccr/xdmod/pull/1706) to make integration tests easier to write and maintain. Specifically, it moves test data out of JSON test artifact files and into PHP associative arrays in the integration test code. This makes the test data simpler for developers to find, read, write, reuse, and debug, and it also speeds up the validation.

This PR also changes the API token authentication tests to only run for a single public user and a single non-public user, resulting in the integration test framework finishing faster.

## Motivation and Context
There are many integration tests that involve making HTTP requests to an endpoint and validating the JSON response. This PR organizes similar code into functions to avoid repetition.

This PR serves as a base for https://github.com/ubccr/xdmod/pull/1788, https://github.com/ubccr/xdmod/pull/1765, https://github.com/ubccr/xdmod-appkernels/pull/95, and https://github.com/ubccr/xdmod-supremm/pull/350.

## Tests performed

In a Docker container, as I refactored tests, I manually changed the covered code to make sure the test still tests that code.

Also, in a Docker container running `tools-ext-01.ccr.xdmod.org/xdmod-10.5.0-x86_64:rockylinux8.5-0.3`:
1. Run the following commands:
    ```
    export COMPOSER_ALLOW_SUPERUSER=1
    export XDMOD_REALMS='jobs,storage,cloud'
    export XDMOD_IS_CORE=yes
    export XDMOD_INSTALL_DIR=/xdmod
    export XDMOD_TEST_MODE=upgrade
    openssl genrsa -rand /proc/cpuinfo:/proc/dma:/proc/filesystems:/proc/interrupts:/proc/ioports:/proc/uptime 2048 > /etc/pki/tls/private/localhost.key
    /usr/bin/openssl req -new -key /etc/pki/tls/private/localhost.key -x509 -sha256 -days 365 -set_serial $RANDOM -extensions v3_req -out /etc/pki/tls/certs/localhost.crt -subj "/C=XX/L=Default City/O=Default Company Ltd"
    git clone {https://github.com/ubccr,}/xdmod
    cd /xdmod
    composer install
    mkdir ~/phpunit
    mkdir /tmp/screenshots
    ~/bin/buildrpm xdmod
    ./tests/ci/bootstrap.sh
    ./tests/ci/validate.sh
    composer install --no-progress
    mv ./configuration/organization.json ./configuration/organization.json.old
    mv ./configuration/portal_settings.ini ./configuration/portal_settings.ini.old
    cp /etc/xdmod/portal_settings.ini ./configuration/portal_settings.ini
    cp /etc/xdmod/organization.json ./configuration/organization.json
    ./tests/ci/scripts/qa-test-setup.sh
    sed -i 's/\$phpunit /$phpunit --debug /' /xdmod/tests/unit/runtests.sh
    /xdmod/tests/unit/runtests.sh > /unit-before.txt
    sed -i 's/\$phpunit /$phpunit --debug /' /xdmod/tests/component/runtests.sh
    /xdmod/tests/component/runtests.sh > /component-before.txt
    sed -i 's/\$phpunit /$phpunit --debug /' /xdmod/tests/regression/runtests.sh
    /xdmod/tests/regression/runtests.sh > /regression-before.txt
    cd /xdmod/tests/integration
    ../../vendor/bin/phpunit --testsuite default --group UserAdminTest.createUsers --debug > /integration-before.txt
    ../../vendor/bin/phpunit --testsuite default --exclude-group UserAdminTest.createUsers --debug >> /integration-before.txt
    git clone https://github.com/aaronweeden/xdmod -b refactor-integration-tests /xdmod-new
    unalias rm
    rm -r /xdmod/tests/integration/lib
    cp -r /xdmod{-new,}/tests/integration/lib
    /xdmod/tests/unit/runtests.sh > /unit-after.txt
    /xdmod/tests/component/runtests.sh > /component-after.txt
    /xdmod/tests/regression/runtests.sh > /regression-after.txt
    cd /xdmod/tests/integration
    ../../vendor/bin/phpunit --testsuite default --group UserAdminTest.createUsers --debug > /integration-after.txt
    ../../vendor/bin/phpunit --testsuite default --exclude-group UserAdminTest.createUsers --debug >> /integration-after.txt
    unalias mv
    for file in /regression-{before,after}.txt; do
        for test in {testChartSettings,testChartFilters,testCsvExport}; do
            for realm in {Jobs,Cloud,Storage}; do
                sed -i "s/$test with data set.*$realm.*/$test $realm/" $file;
            done;
        done;
        sort $file > $file.tmp;
        mv $file.tmp $file;
    done
    ```
1. Compare `/unit-{before,after}.txt` and make sure the only differences are user hashes and amount of time the tests ran.
1. Compare `/component-{before,after}.txt` and make sure the only differences are test name hashes and amount of time the tests ran.
1. Compare `/regression-{before,after}.txt` and make sure the only differences are the progress dots at the top and the amount of time the tests ran.
1. Compare `/integration-{before,after}.txt` and make sure the differences are correct:
    1. First set of tests: 497 assertions -> 2972 assertions.
    1. Different random values for `ControllerTest::testEnumTargetAddresses`.
    1. In `/integration-after.txt`, only `usr` tests for:
        * `MetricExplorerTest::testGetDwDescripterTokenAuth`.
        * `MetricExplorerTest::testGetDimensionFilters` — and also a single `valid_token` test each for `cd`, `cs`, `usr`, `pi`, and `mgr`.
        * `UsageExplorerTest::testGetDataTokenAuth`.
        * `WarehouseControllerProviderTest::testGetRawData`.
        * `WarehouseControllerProviderTest::testGetRawDataLimit`.
        * `WarehouseExportControllerProviderTest::testGetRealmsTokenAuth`.
    1. Different test arguments for `UserControllerProviderTest::testGetCurrentUser` and `WarehouseControllerProviderTest::testGetRawData`.
    1. Last group of tests takes about 30 sec less, ~60 fewer MB, 813 tests -> 722 tests, 6469 assertions -> 570956 assertions.

## Checklist:


tests/integration/lib/BaseTest.php
```diff
@@ -298,27 +275,316 @@ protected static function assertRequiredKeys(
...
+    /**
+     * Validate ISO 8601 dates.
+     *
+     * @param string $date
+     * @return null
+     */
+    protected function validateDate($date)
+    {
+        parent::assertRegExp(
+            '/[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/',
```
ryanrath, line 550: minor thing, but I'd recommend pulling this out to a const.

tests/integration/lib/BaseTest.php
```diff
@@ -371,4 +637,146 @@ private static function truncateStr($str, $numChars)
...
+    ) {
+        $types = [
+            'int_params' => 'integer',
+            'date_params' => 'ISO 8601 Date'
+        ];
+        foreach ($types as $key => $type) {
+            if (array_key_exists($key, $options)) {
+                foreach ($options[$key] as $param) {
+                    $input = $validInputWithAdditionalParams;
+                    $input[$paramSource][$param] = 'foo';
```
ryanrath, line 770: Was this meant to be here?

tests/integration/lib/TokenAuthTest.php
```diff
@@ -37,114 +37,64 @@ abstract class TokenAuthTest extends BaseTest
...
-     *                         if $tokenType is 'valid_token', the output
-     *                         artifact that describes how to validate the HTTP
-     *                         response body.
-     * @param string|null $testKey if the test artifact files contain multiple
-     *                             keys, a single key must be provided to
-     *                             indicate which test to run, and this key
-     *                             must be present in both the input and output
-     *                             test artifacts.
+     *                          (@see provideTokenAuthTestData()).
+     * @param array input @see BaseTest::requestAndValidateJson(). Must have
```
ryanrath, line 61: missing a `$` in front of `input`

tests/integration/lib/TokenAuthTest.php
```diff
@@ -37,114 +37,64 @@ abstract class TokenAuthTest extends BaseTest
...
+     *                    is either 'controller' or 'rest', and
+     *                    'authentication_type', whose value is either
+     *                    'token_optional' or 'token_required'.
+     * @param array $output @see BaseTest::requestAndValidateJson(). If
+     *                      $tokenType is not 'valid_token', this will be
+     *                      ignored and instead the relevant
+     *                      authentication/authorization error testing will be
+     *                      performed.
      * @return mixed the decoded JSON response body of the request.
      * @throws Exception if the role is unrecognized or there is an error
```
ryanrath, line 72: This line won't have an affect on the running of tests, but you'll want to either change this to `\Exception` or `use Exception;` at the top of the file ( Unless you have a new class `IntegrationTests\Exception` ).

tests/integration/lib/TokenAuthTest.php
```diff
@@ -153,9 +103,9 @@ public function runTokenAuthTest(
                     'SUBDATE(NOW(), 1)'
```
ryanrath, line 103: This makes me feel dirty inside, lol... The dynamic SQL doesn't really matter since it's only available in test but still.

tests/integration/lib/TokenAuthTest.php
```diff
@@ -174,9 +124,15 @@ public function runTokenAuthTest(
...
+                        'body_validator' => $this->validateErrorResponseBody(
+                            'Session Expired',
+                            2
+                        )
+                    ];
                 } elseif ('rest' === $input['endpoint_type']) {
-                    $testKey = 'authentication_error';
+                    $output = parent::validateAuthorizationErrorResponse(401);
                 } else {
                     throw new Exception(
```
ryanrath, line 137: Ah, so here's where you might run into trouble if you don't `use Exception` above. Again, unless I'm missing an `IntegrationTests\Exception` class.

tests/integration/lib/Rest/WarehouseExportControllerProviderTest.php
```diff
@@ -192,8 +192,58 @@ public function testGetRealmsTokenAuth($role, $tokenType) {
...
+                'method' => 'get',
+                'params' => null,
+                'data' => null,
+                'endpoint_type' => 'rest',
+                'authentication_type' => 'token_optional'
+            ],
+            parent::validateSuccessResponse(function ($body, $assertMessage) {
+                $this->assertSame(2, $body['total'], $assertMessage);
+                $index = 0;
+                foreach (['Jobs', 'Cloud'] as $realmName) {
```
ryanrath, line 206: Do we have any Storage realm data to test?

tests/integration/lib/TokenAuthTest.php
```diff
@@ -186,32 +142,26 @@ public function runTokenAuthTest(
...
+                ];
+                $output = [
+                    'status_code' => 401,
+                    'body_validator' => $this->validateErrorResponseBody(
+                        $messages[$tokenType],
+                        0
+                    )
+                ];
             } else {
                 throw new Exception(
```
ryanrath, line 160: 'nother instance of `Exception` like above.

tests/integration/lib/TokenAuthTest.php
ryanrath: Just something else I noticed, you may want to note that the function can throw an exception if an invalid value for `endpoint_type`, `$tokenType`, or `authentication_type` is provided.

tests/integration/lib/BaseTest.php
```diff
@@ -5,9 +5,9 @@
 use CCR\Json;
 use Exception;
 use Swaggest\JsonSchema\Schema;
-use TestHarness\IntegrationTestConfiguration;
 use TestHarness\TestFiles;
 use TestHarness\Utilities;
+use TestHarness\XdmodTestHelper;
 
 abstract class BaseTest extends \PHPUnit_Framework_TestCase
```
ryanrath, line 12: Could we get some documentation on how the class is meant to be used? Something like a simple example test case?

ryanrath (commented):
Alrighty, just some comments with no actual show stoppers. It's mainly documentation & naming related w/ a sprinkling of possible gotchas ( some Exception namespace stuff ).

tests/integration/lib/Rest/WarehouseExportControllerProviderTest.php
aaronweeden: No, the Storage raw data export is not implemented.

tests/integration/lib/BaseTest.php
ryanrath: Ah yes, provideRestEndpoint**Invalid**ParamTests. Just so I'm sure I'm reading this one right... 
If the provided `$options` contains [int|date]_params, then it will add invalid string tests for those params? Does it make sense to add invalid tests for strings? Or is that kidna moot because everything is coerced into a string?

tests/integration/lib/Rest/WarehouseExportControllerProviderTest.php
ryanrath: Ah ok, I thought it was tests for the Data Export Tab. ( And now that I look at it, we don't have Storage available for export anyway ). No worries.

tests/integration/lib/BaseTest.php
aaronweeden: Yes, you are reading it right, and yes, arrays parameters aren't coerced to strings, and validating them is currently broken; this is to be fixed in https://github.com/ubccr/xdmod/pull/1768, which includes tests of string parameter validation.

ryanrath: Sweet, sounds good!

tests/integration/lib/BaseTest.php
aaronweeden: There are a number of examples in https://github.com/ubccr/xdmod/pull/1765 (which is based on this PR. Perhaps best viewed as a diff between them: https://github.com/aaronweeden/xdmod/compare/refactor-integration-tests...add-int-param-tests).

I can add a comment to the top of this class to say more about what it does and give examples.

tests/integration/lib/BaseTest.php
```diff
@@ -298,27 +275,316 @@ protected static function assertRequiredKeys(
...
+        $validInputWithAdditionalParams = $validInput;
+        if (array_key_exists('additional_params', $options)) {
+            $validInputWithAdditionalParams = self::mergeParams(
+                $validInputWithAdditionalParams,
+                $paramSource,
+                $options['additional_params']
+            );
+        }
+        $tests = [];
+        $this->provideRestEndpointAuthenticationTests(
```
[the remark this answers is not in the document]
aaronweeden: I agree with these points and have folded the sub-functions into the overall function to address them. SonarCloud says it increases the cognitive complexity of the function, but I think the new comments help, and it seems worth it to avoid the argument mutation of which I'm also not a fan in general.

I'll add a further update to make sure the comment lines with deep indentations are wider. In general I like 80 characters since it tends to keep everything in a tight column that my eyes can scan faster and doesn't wrap as much when doing side-by-side diffs, but when the lines are indented this much the effective column width is too squished.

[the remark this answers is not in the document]
aaronweeden: Updated the comment widths.

tests/integration/lib/BaseTest.php
aaronweeden: Done.

tests/integration/lib/TokenAuthTest.php
aaronweeden: Fixed.

tests/integration/lib/TokenAuthTest.php
aaronweeden: Done.

tests/integration/lib/TokenAuthTest.php
aaronweeden: Me too. Is there a better way to do it?

tests/integration/lib/TokenAuthTest.php
aaronweeden: Fixed.

tests/integration/lib/TokenAuthTest.php
aaronweeden: Fixed.

tests/integration/lib/BaseTest.php
aaronweeden: Added a comment to the top of the class.

tests/integration/lib/TokenAuthTest.php
ryanrath: Hrmm... good question... So one possibility is it could look something like: 

```php
/**
 * Expire the API token for the user identified by $userId.
 *
 * @param string $userId the id of the user who will have their API token expired.
 * @return int the number of rows updated. It's expected that a successful update will result in 1 row being updated.
 * @throws Exception if there is a problem constructing the database connection
 * @throws Exception if there is a problem executing the SQL statement to expire the users API token.
 */
function expireToken($userId)
{
    $db = DB::factory('databse');
    $query = "UPDATE moddb.user_tokens SET expires_on = SUBDATE(NOW(), 1) WHERE user_id = :user_id";
    $params = [':user_id' => $userId];
    return $db->execute($query, $params);
}

/**
 * Un-Expire the API token for the user identified by $userId.
 *
 * @param string $userId the id of the user who will have their API token un-eexpired.
 * @return int the number of rows updated. It's expected that a successful update will result in 1 row being updated.
 * @throws Exception if there is a problem constructing the database connection
 * @throws Exception if there is a problem executing the SQL statement to un-expire the users API token.
 */
function unexpireToken($userId)
{
    $db = DB::factory('databse');
    $query = "UPDATE moddb.user_tokens SET expires_on = ADDDATE(NOW(), 1) WHERE user_id = :user_id";
    $params = [':user_id' => $userId];
    return $db->execute($query, $params);
}
```

That way we explicitly know what will happen when we call each function ( we can also check the returned int value to determine if it was successful )

ryanrath:
Lemme know what you think about the `expireToken` function suggestions and then just re-request the review and I'll go ahead and approve.

tests/integration/lib/TokenAuthTest.php
aaronweeden: I agree a call like `expireToken($role)` is better, but I also like having a function to reduce all the duplicated code. I've committed a compromise that keeps `updateTokenExpirationDate()` but comments that it's only intended to be called by `expireToken()` and `unexpireToken()`.

ryanrath (approved):

PR #1773, opened 2023-09-12
Removing XSEDE specific GAQ

ryanrath:
We no longer have an XSEDE specific Google Analytics files.

## Checklist:


jpwhite4:
Is this file necessary - seems odd to  have a php file whose only job is to require_once another php file. Why not just require_once the second file and get rid of this one completely.

jpwhite4 (approved):

PR #1778, opened 2023-09-13
Add check to sanitize job timeseries data of NAN values

aestoltm:
## Description

Add check to sanitize job timeseries data of NAN values due to Symfony throwing double NAN error when json encoding. Added helper function which replaces any NAN values in the series data with 0. 

Note: This may need to be addressed in other areas. Need to find where else we are possibly letting NAN/null values through.

## Motivation and Context

Charts with NAN data would return a null record.

## Tests performed

Observed changes on xdmod-dev

## Checklist:

### DIFF
--- a/classes/Rest/Controllers/WarehouseControllerProvider.php
+++ b/classes/Rest/Controllers/WarehouseControllerProvider.php
@@ -2501,5 +2501,4 @@ class WarehouseControllerProvider extends BaseControllerProvider
         }
         return $ret;
     }
-
 }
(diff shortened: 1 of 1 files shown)


jpwhite4 (changes requested):
I really would prefer you change the JobMetadata->getJobTimeseriesData() implementation so that it does not return invalid values. - rather than having it still return nan data and then fudging it afterwards.

aestoltm:
Closed. Refer to PR: https://github.com/ubccr/xdmod-supremm/pull/353

PR #1780, opened 2023-09-20
Optimize raw data job queries.

aaronweeden:
## Description

This PR optimizes raw data queries in the Jobs realm by using `WHERE` conditions on the `end_day_id` field of `modw.job_tasks` instead of the `end_time_ts` field. This PR also removes the `DISTINCT` modifier for single-day queries.

There is also a PR for https://github.com/ubccr/xdmod-xsede/pull/433.

## Motivation and Context

The new query is much faster (<10 seconds per single-day query compared to ~200 seconds), and in conjunction with https://github.com/ubccr/xdmod/pull/1792 and https://github.com/ubccr/xdmod-data/pull/19, it makes the data analytics framework more responsive.

## Tests performed

In addition to ensuring the regression tests from https://github.com/ubccr/xdmod/pull/1788 still pass:
1. Run the script from https://github.com/ubccr/xdmod-xsede/pull/436 with the `--time` argument.
    1. Set `'new': 'https://xdmod-dev.ccr.xdmod.org'` since https://github.com/ubccr/xdmod-xsede/pull/433 is already merged.
    1. Set `'old': 'https://xdmod-dev.ccr.xdmod.org:9001'` with the version of `classes/DataWarehouse/Query/Jobs/JobDataset.php` from https://github.com/ubccr/xdmod/pull/1779.
1. Run the script again with the `--time` argument and the variable `NUM_DAYS_PER_REQUEST` set to 1:
    1. Confirm the new Jobs realm queries take <10 seconds compared to ~200 seconds for the old queries.

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Query/Jobs/JobDataset.php
+++ b/classes/DataWarehouse/Query/Jobs/JobDataset.php
@@ -94,8 +94,28 @@ class JobDataset extends \DataWarehouse\Query\RawQuery
                 throw new Exception('invalid "end_date" query parameter');
             }
 
-            $this->addPdoWhereCondition(new WhereCondition(new TableField($factTable, 'end_time_ts'), ">=", $startDateTs));
-            $this->addPdoWhereCondition(new WhereCondition(new TableField($factTable, 'end_time_ts'), "<=", $endDateTs));
+            $startDayId = (
+                date('Y', $startDateTs) * 100000
+                + date('z', $startDateTs) + 1
+            );
+            $endDayId = (
+                date('Y', $endDateTs) * 100000
+                + date('z', $endDateTs) + 1
+            );
+            $factEndIdField = new TableField($factTable, 'end_day_id');
+            $this->addWhereCondition(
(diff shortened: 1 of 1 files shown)


classes/DataWarehouse/Query/Jobs/JobDataset.php
```diff
@@ -94,8 +94,49 @@ public function __construct(
...
+                    )
+                );
+                $this->addPdoWhereCondition(
+                    new WhereCondition(
+                        new TableField($factTable, 'end_time_ts'),
+                        '<=',
+                        $endDateTs
+                    )
+                );
+            }
```
eiffel777, line 139: Is there a reason why either the `day_id` or `end_day_id` field wouldn't be better for multiple day queries as well?

aaronweeden: Good idea, I have changed it to use `end_day_id`.

eiffel777 (approved):

PR #1793, opened 2023-11-30
Add slightly less confusing error message.

jpwhite4:
## Description

Attempt to make the error message displayed when there is a database connection problem slightly less confusing.

We still report the execption code so that an admin could potentially work out what is wrong.

image


html/index.php
```diff
@@ -76,7 +77,12 @@ function isReferrer($referrer)
         $_SESSION['public_session_token'] = 'public-' . microtime(true) . '-' . uniqid();
     }
 
-    $user = XDUser::getPublicUser();
+    try {
+        $user = XDUser::getPublicUser();
+    } catch (PDOException $e) {
+        xd_web_message\displayMessage('XDMoD is currently under maintenance and access is temporarily disabled.', 'Status: code=' . $e->getCode(), true);
```
aaronweeden, line 83:
```suggestion
        xd_web_message\displayMessage('XDMoD is currently experiencing a temporary outage.', 'Status: session=false code=' . $e->getCode(), true);
```

html/index.php
```diff
@@ -41,11 +41,12 @@
         // TODO: Refactor generic catch block below to handle specific exceptions,
         //       which would allow this block to be removed.
         throw $see;
+    } catch (PDOException $e) {
+        xd_web_message\displayMessage('XDMoD is currently under maintenance and access is temporarily disabled.', 'Status: session=true code=' . $e->getCode(), true);
```
aaronweeden, line 45:
```suggestion
        xd_web_message\displayMessage('XDMoD is currently experiencing a temporary outage.', 'Status: session=true code=' . $e->getCode(), true);
```

aaronweeden (commented):
Couple suggested changes since people might take "maintenance" to mean we intentionally brought it down, and "access" might be misunderstood as "ACCESS" (I'm less worried about that one).

aaronweeden (approved):

PR #1801, opened 2023-12-15
Fix utilization metrics for Jobs realm

eiffel777:
This change to the utilization metric for the jobs realm greatly improves the reliability and performance of the metric. It uses the resource specs aggregate table to get the core hours for a time period.

The aggregate tests changed because the new statistic gets a resources' core hours from the earliest day a job exists in the xdmod database rather than the first day of a resource specification. The robertson and pozidriv resource has specification that start on 12/12 and 12/21 and the earliest date of a job in the database is 12/22.

Some timeseries values changed because the new value is the fewer decimal places, 4 instead of 10. This seems the be because the old utilization metric had a division operation which cause the extra decimal places. The new statistic does not have a division operation. Adding `/ 1` to the new metric will produce the same value, to the same number of decimal places, as the old metric. The value in the metric explorer and usage tab is the still the same since it rounds to 2 decimal places, it is the export that goes to the full number of decimal places.

## Tests performed
Tested in docker

## Checklist:


configuration/datawarehouse.d/ref/Jobs-statistics.json
```diff
@@ -263,15 +263,11 @@
         "unit": "Hour"
     },
     "utilization": {
-        "aggregate_formula": {
-            "$include": "datawarehouse.d/include/Jobs-util-agg.sql"
```
aaronweeden, line 267: It looks like `configuration/datawarehouse.d/include/Jobs-util-time.sql` and `configuration/datawarehouse.d/include/Jobs-util-agg.sql` aren't being used anymore and can be removed.

aaronweeden (changes requested):

aaronweeden (approved):

PR #1803, opened 2023-12-20
Change host for Ext JS dependency

connersaeli:
## Description

There appear to be multiple files with the same name in https://cdn.sencha.com/ext/gpl/ext-3.4.1.1-gpl.zip, which causes an issue when doing a composer install. Composer uses `unzip` under the hood when installing zipped packages like these. This breaks `composer install` when `unzip` prompts to replace the files with the same name.

There is a corresponding PR [here](https://tools-int-01.ccr.xdmod.org/git/UBCCR/xms-puppet/pulls/136) to make the necessary infrastructure updates to support this. This change will not work until the fixed zip file is being served.

## Motivation and Context

Doing a `composer install` with the old zip file would not complete due to the files with the same name. This also appears to silently break the CI tests (look closely at the "install composer dependencies" step).

## Tests performed

```
cd xdmod
$ composer -V
Composer version 2.6.6 2023-12-08 18:32:26
$ composer install
```
Should now work instead of throwing an error when unzip fails.

Tested when setting up xdmod-dev. This was also tested against Composer version 2.4.4 (the same version on the current xdmod-dev)

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,9 +46,9 @@
                 "license": "GPL-3.0",
                 "homepage": "https://www.sencha.com/products/extjs",
                 "dist": {
-                    "url": "https://cdn.sencha.com/ext/gpl/ext-3.4.1.1-gpl.zip",
+                    "url": "https://tools-int-01.ccr.xdmod.org/ext/gpl/ext-3.4.1.1-gpl.zip",
                     "type": "zip",
-                    "shasum": "26734b47eae909ff7f8cd7de4cadfb3531bd3cdc"
+                    "shasum": "b0cd8f5524a0076e78b9c230dc7ec4150908c74f"
                 },
                 "require": {
                     "composer/installers": "~1.0"
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
                 "license": "GPL-3.0",
                 "homepage": "https://www.sencha.com/products/extjs",
                 "dist": {
-                    "url": "https://tools-int-01.ccr.xdmod.org/ext/gpl/ext-3.4.1.1-gpl.zip",
+                    "url": "https://tools-ext-01.ccr.xdmod.org/ext/gpl/ext-3.4.1.1-gpl.zip",
                     "type": "zip",
                     "shasum": "b0cd8f5524a0076e78b9c230dc7ec4150908c74f"
                 },
--- a/composer.lock
+++ b/composer.lock
@@ -1852,8 +1852,8 @@
             "version": "3.4.1.1",
             "dist": {
                 "type": "zip",
-                "url": "https://cdn.sencha.com/ext/gpl/ext-3.4.1.1-gpl.zip",
-                "shasum": "26734b47eae909ff7f8cd7de4cadfb3531bd3cdc"
+                "url": "https://tools-int-01.ccr.xdmod.org/ext/gpl/ext-3.4.1.1-gpl.zip",
+                "shasum": "b0cd8f5524a0076e78b9c230dc7ec4150908c74f"
             },
             "require": {
                 "composer/installers": "~1.0"
--- a/composer.lock
+++ b/composer.lock
@@ -1852,7 +1852,7 @@
             "version": "3.4.1.1",
             "dist": {
                 "type": "zip",
-                "url": "https://tools-int-01.ccr.xdmod.org/ext/gpl/ext-3.4.1.1-gpl.zip",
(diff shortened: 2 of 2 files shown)


aaronweeden:
What are the steps to reproduce the problem?

It looks like from `unzip -l ext-3.4.1.1-gpl.zip` that there are indeed duplicates in the zip file.

It looks from the CI output that unzip is acting as if the user entered `N`; i.e., not replacing any files; i.e., using the older version of each file.

Confirming this in a Docker container running `tools-ext-01.ccr.xdmod.org/xdmod-10.5.0-x86_64:rockylinux8.5-0.3`:
```
git clone https://github.com/ubccr/xdmod
cd xdmod
composer install
cd /tmp
wget https://cdn.sencha.com/ext/gpl/ext-3.4.1.1-gpl.zip
unzip ext-3.4.1.1-gpl.zip -d 1
# (Enter N)
diff -r 1/ext-3.4.1/ /xdmod/html/gui/lib/extjs/
# (No differences)
```
Comparing the result to instead entering `A` to replace all the files (i.e., using the newer version of each file):
```
unzip ext-3.4.1.1-gpl.zip -d 2
# (Enter A)
diff -r 2/ext-3.4.1/ /xdmod/html/gui/lib/extjs/
```
The differences appear to be all comment headers. Confirming this:
```
diff -r 2/ext-3.4.1/ /xdmod/html/gui/lib/extjs/ | sort | uniq | less
```
So the files are missing their comment headers, but otherwise the code is the same.

connersaeli:
Aaron,

Steps to reproduce (on `mms-dev-ws`):
```
git clone git@github.com:ubccr/xdmod.git
```
Get the new script to update the portal. The diff is for `update-open-xdmod-portal.php` is [here](https://tools-int-01.ccr.xdmod.org/git/connersa/ccr-private-xdmod/commit/b6bad8d027535c5dfb96b90c30074789ac20745f) and just gets rid of old `composer-el7.*`.
```
git clone git@tools-int-01.ccr.xdmod.org:connersa/ccr-private-xdmod.git \
    --branch xdmod-dev-rocky8
    --single-branch
```
```
cd ccr-private-xdmod
```
```
./scripts/deployment/update-open-xdmod-portal.php \
    --xsede \
    --prefix /data/www/$USER \
    --source-dir /home/$USER/xdmod \
    --conf-repository $PWD
    --skip-user-manual
```

aaronweeden:
@connersaeli: Thanks, I was able to reproduce the problem.

It looks like `mms-dev-ws` doesn't have `php-zip` installed (`php -m` doesn't list `zip`) so composer doesn't fall back to using the `ZipArchive` class, whereas in the Docker container it does. In the Docker container, I tried `dnf remove php-zip -y` followed by `composer install`, and it also fails.

So, if we don't want to host this file ourselves, we could add build instructions saying that `php-zip` is required.

Incidentally, I see that we already have a section in the `xdmod` README titled "Why is Composer failing to unzip Ext JS?" https://github.com/ubccr/xdmod?tab=readme-ov-file#why-is-composer-failing-to-unzip-ext-js

Here's where I was looking in the composer source code that led me to the conclusion above:
1. Error message mentions ZipDownloader.php line 185, which is here:
https://github.com/composer/composer/blob/8e62977cb5de315d5190c021a2b2bb4e59ee2f4f/src/Composer/Downloader/ZipDownloader.php#L185C29-L185C40
1. `$tryFallback` is defined here; it looks like it is entering the if block on lines 143–145: https://github.com/composer/composer/blob/8e62977cb5de315d5190c021a2b2bb4e59ee2f4f/src/Composer/Downloader/ZipDownloader.php#L142
1. `$isLastChance` is defined here:
https://github.com/composer/composer/blob/8e62977cb5de315d5190c021a2b2bb4e59ee2f4f/src/Composer/Downloader/ZipDownloader.php#L115
1. `$hasZipArchive` is defined here: https://github.com/composer/composer/blob/8e62977cb5de315d5190c021a2b2bb4e59ee2f4f/src/Composer/Downloader/ZipDownloader.php#L69

connersaeli:
@aaronweeden I think it will be a lot less onerous to just install `php-zip` on that machine and document it. I have a set of documentation changes for the setup of `mms-dev-ws`, so I will add it to that. I will also update the README. I think the question is still relevant, but it has a different answer now.

I'm just going to test that out and see if it works on `mms-dev-ws` now. Thanks for your help with this!

PR #1810, opened 2024-01-17
Adding OpenID Connect Instructions for SSO

ryanrath:
## Description
Added a new documentation page that describes the steps that need to be taken to setup OpenID Connect as an Identity Provider for XDMoD. 

## Motivation and Context
We had a support ticket asking for how to set this up and realized that while we use OpenID Connect as an IDP in one of our instances, we didn't have any documentation for how to set it up. This rectifies that problem.

## Checklist:


.circleci/config.yml
```diff
@@ -38,16 +38,6 @@ jobs:
...
-      - restore_cache:
-          keys:
-            # "composer.lock" can be used if it is committed to the repo
-            - v1-el8-dependencies-{{ checksum "composer.json" }}
-            # fallback to using the latest cache if no exact match is found
-            - v1-el8-dependencies-
-      - save_cache:
-          key: v1-el8-dependencies-{{ checksum "composer.json" }}
-          paths:
-            - ./vendor
```
eiffel777, line 50: If this is just a documentation PR why is this being removed? If this needs to be removed, should it be done in a different PR?

ryanrath: Needed to make it so that the ci would run.

eiffel777 (approved):

PR #1813, opened 2024-01-25
Plotly Conversion for Summary, Usage and Metric Explorer

aestoltm:
## Description

These changes are for the Summary, Usage and Metric Explorer Tabs.

 **When finding a bug please compare with an instance running Highcharts** so we know if this is an existing bug or not.

**Features not in Plotly JS yet**
- Hover events styling marker and line sizes
- Shadow

## Motivation and Context

Plotly JS has no license requirement which is desirable for Open XDMoD.

## Tests performed

Tested on docker and dev-port. Will be converting the Playwright tests to work with Plotly JS.

## Checklist:


classes/DataWarehouse/Access/MetricExplorer.php
```diff
@@ -104,7 +104,7 @@ public function get_data($user)
         $filename = substr($filename, 0, 250);
 
         $all_data_series = $this->getDataSeries();
-
+
```
connersaeli, line 107:
```suggestion
```

ryanrath:
Apologies if this has already been covered, but does the date / footer always have to be included? It looks like it might be contributing to some "squishage" that's going on: 
image

aestoltm:
> Apologies if this has already been covered, but does the date / footer always have to be included? It looks like it might be contributing to some "squishage" that's going on: image
> 
> There might also be some instances of the legend being cut off?

The footer does not belong there so it will need to be removed. The legend shouldn't be cutt off, Plotly creates scrollbar for the other entries

aestoltm:
Moved to PR #1822 because of history fix for Highcharts2.php and an attempt to fix the linter issues with a clean PR

PR #1816, opened 2024-02-02
Allow `@` in remote user format string when parsing web server log files.

aaronweeden:
## Description

This PR updates the web server log file parser to allow `@` in remote user values.

## Motivation and Context

Remote user values are logged by some identity providers in the form `<user>@<domain>`. This PR makes sure that the log parser does not skip these lines. This is relevant for the `ondemand` module and was noticed when testing the upcoming OnDemand realm in ACCESS XDMoD.

## Tests performed

I tested this by running `xdmod-ondemand-ingestor` on my port on `xdmod-dev` with log files containing user values containing `@`. Before the changes in this PR, the lines are skipped; after, they are ingested.

## Checklist:

### DIFF
--- a/classes/ETL/DataEndpoint/WebServerLogFile.php
+++ b/classes/ETL/DataEndpoint/WebServerLogFile.php
@@ -47,6 +47,12 @@ class WebServerLogFile extends aStructuredFile implements iStructuredFile
         parent::__construct($options, $logger);
 
         $this->web_parser = new \Kassner\LogParser\LogParser();
+
+        // Allow "at" sign in remote user format string.
+        // This can be removed if Kassner LogParser is updated to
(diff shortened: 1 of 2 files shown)


jpwhite4:
It seem this change will not be needed with the next (as yet unreleased) version of LogParser. To help us keep the code clean in future I suggest that you add a comment to the composer.json next to the LogParser dependency metadata that reminds us to remove the remote user format string override if we change the version number. (I think you could use the "description" property to do this).

aaronweeden:
@jpwhite4 I have added a comment above the code and a comment to `composer.json` using `"extra"`. I have also set myself up for GitHub notifications of new releases from `kassner/log-parser`.

eiffel777 (approved):

PR #1824, opened 2024-02-21
Parse request method, URL, and protocol separately in web server logs.

aaronweeden:
## Description

This PR updates the web server log parser to split the request line (i.e., `%r`, e.g., `GET /index.html HTTP/1.1`) into its three parts (i.e., `%m %U %H`) so these can be parsed separately.

## Motivation and Context

This enables https://github.com/ubccr/xdmod-ondemand/pull/50.

## Tests performed

I added unit tests to make sure a log is correctly parsed whether the format is `%r` or `%m %U %H`. I also tested ingesting logs on my port on `xdmod-dev` with the changes from https://github.com/ubccr/xdmod-ondemand/pull/50 and the PRs on which it depends.

## Checklist:

### DIFF
--- a/classes/ETL/DataEndpoint/WebServerLogFile.php
+++ b/classes/ETL/DataEndpoint/WebServerLogFile.php
@@ -54,6 +54,13 @@ class WebServerLogFile extends aStructuredFile implements iStructuredFile
         $this->web_parser->addPattern('%u', '(?P<user>(?:-|[\w\-\.@]+))');
 
         if (isset($options->log_format)) {
+            // Replace `%r` with `%m %U %H` so the request method, URL, and
+            // protocol can be parsed separately.
+            $options->log_format = str_replace(
+                '%r',
+                '%m %U %H',
+                $options->log_format
+            );
             $this->web_parser->setFormat($options->log_format);
         }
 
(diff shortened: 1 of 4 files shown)


jpwhite4:
The webserver log parser already supports requestMethod (%m) and url === requestPath (%U) format strings. 

Why don't we use these?

classes/ETL/DataEndpoint/WebServerLogFile.php
```diff
@@ -109,6 +109,10 @@ protected function decodeRecord($data)
         try {
             $decoded = $this->web_parser->parse($data);
 
+            $requestParts = explode(' ', $decoded->request);
```
jpwhite4, line 112: In general can't guarantee that request is in the decoded output (e.g. could use "%m %U %H" instead of "%r") [in fact we probably want to do this in the ondemand module]

jpwhite4: If you really do want to do this I recommend a `property_exists` guard around it.

jpwhite4 (changes requested):

aaronweeden:
> The webserver log parser already supports requestMethod (%m) and url === requestPath (%U) format strings.
> 
> Why don't we use these?

@jpwhite4 Thanks, I have refactored it to use `%m`, `%U`, and `%H`; added unit tests; and updated https://github.com/ubccr/xdmod-ondemand/pull/50.

jpwhite4 (approved):

PR #1838, opened 2024-05-08
Allow multiple columns in `RegexTransformIngestor`.

aaronweeden:
_NOTE: This PR should not be merged unless https://github.com/ubccr/xdmod-ondemand/pull/50 is also merged; otherwise `xdmod-ondemand` ingestion will break._
## Description

This PR allows the `RegexTransformIngestor` to transform multiple columns.

## Motivation and Context

This enables https://github.com/ubccr/xdmod-ondemand/pull/50, specifically the ability to transform `app`, `request_path`, `reverse_proxy_host`, and `reverse_proxy_port` in the same pipeline action (`ondemand.log-ingestion.normalize-staging`).

## Tests performed

I tested ingesting logs on my port on `xdmod-dev` with the changes from https://github.com/ubccr/xdmod-ondemand/pull/50 and the PRs on which it depends.

## Checklist:

### DIFF
--- a/classes/ETL/Ingestor/RegexTransformIngestor.php
+++ b/classes/ETL/Ingestor/RegexTransformIngestor.php
@@ -2,26 +2,31 @@
 /* ==========================================================================================
  * RegexTransform. This ingestor transforms values using the preg_filter() function.
  *
- * The regular expressions, name of the source to split and the destination column
+ * The regular expressions, names of the sources to split and the destination columns
  * to populate are specified using configuration properties. All other source
- * columns are passed unmodified. If the regular expression does not match then
+ * columns are passed unmodified. If no regular expressions match then
  * the row is not passed by the ingestor.
  *
  * Configuration properties:
  *
- * - regex_column: defines the column in the source table to read and column name in the
+ * - regex_column: defines the columns in the source table to read and column names in the
  *                 output to use for the transformed data. For example:
- *                 { "source": "dest" } would read the data in column named "source" and
- *                 the transformed content of "source" would be written to "dest".
- * - regex_config: a json formatted string that contains regular expression and output
- *                 patterns. The the regex format is the one used by preg_filter().
+ *                 { "dest1": "source1", "dest2": "source2" } would read the data in column named
+ *                 "source1" and the transformed content of "source1" would be written to "dest1",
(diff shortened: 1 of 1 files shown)


classes/ETL/Ingestor/RegexTransformIngestor.php
```diff
@@ -54,38 +66,79 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Logg
...
     }
 
     /**
      * @see ETL\Ingestor\pdoIngestor::transform()
      */
-    protected function transform(array $srcRecord, $orderId)
+    protected function transform(array $srcRecord, &$orderId)
     {
         $transformedRecord = array();
+        $outdata = $srcRecord;
```
jpwhite4, line 81: I think that this function is called for each record in the data: You should avoid opening and parsing config files in this function instead do the configuration setup once in the constructor.

aaronweeden: I have updated it to do the opening and parsing in the constructor.

classes/ETL/Ingestor/RegexTransformIngestor.php
```diff
@@ -54,38 +46,41 @@ public function __construct(aOptions $options, EtlConfiguration $etlConfig, Logg
...
-
-        if ($res !== null) {
-            $outdata = $srcRecord;
-            $outdata[$this->destKey] = $res;
-
+            if (!is_null($res)) {
+                $outdata[$dest] = $res;
+                $ignored = false;
+            } else {
+                $this->logger->debug("$dest: Ignore $srcColumn");
```
jpwhite4, line 78: Should this debug statement be in the `else` part of the if statement on line 81? If you have multiple regexes then one can match and another not match and the row will not be ignored. It will only be ignored if all of the regexes fail to match.

jpwhite4 (changes requested):

classes/ETL/Ingestor/RegexTransformIngestor.php
aaronweeden: Thanks, I have updated it.

jpwhite4 (approved):

PR #1851, opened 2024-05-13
Add visualization of null values in Job Viewer charts

aestoltm:
## Description

Added dashed line for null data similar to other tabs. Also fixed bug with `if (args)` always evaluating to true which would cause undefined values for some chart element sizes.

This PR is an addition to the fix in [xdmod-supremm PR #353](https://github.com/ubccr/xdmod-supremm/pull/353)

## Motivation and Context

Users can visualize missing data instead of gaps in line segments.

## Tests performed

Tested on my devport so far.

## Checklist:


html/gui/js/libraries/PlotlyUtilities.js
```diff
@@ -52,19 +52,18 @@ function getNoDataErrorConfig() { // eslint-disable-line no-unused-vars
  * @param  {Object} Record containing chart data
  *
  */
-function generateChartOptions(record, params) { // eslint-disable-line no-unused-vars
-    var args = params || {};
+function generateChartOptions(record, params = null) { // eslint-disable-line no-unused-vars
```
jpwhite4, line 55: is this eslint override needed now that you are using the params variable?

aestoltm: No, because the `no-unused-vars` error is thrown on the function itself. ESLint thinks the function is unused based on its [criteria](https://eslint.org/docs/latest/rules/no-unused-vars#rule-details) for determining unused variables/functions. It looks like I can label the function as 'exported' as shown [here](https://eslint.org/docs/latest/rules/no-unused-vars#exported) without ignoring `no-unused-vars`. This is probably a better way so that we don't disable any rules.

jpwhite4: I suggest using the 'exported' label, - because it is!

jpwhite4 (approved):

PR #1859, opened 2024-05-24
Add Metric Explorer copy to clipboard

aestoltm:
## Description

Add copy to clipboard functionality when generating a link to the current chart in the Metric Explorer tab. Also added changelog information about copy to clipboard for both the Metric Explorer and the Job Viewer.

## Motivation and Context

Much easier to copy shareable link and removes possibility of copying a partial link.

## Tests performed

Tested on dev

## Checklist:


html/gui/js/modules/metric_explorer/MetricExplorer.js
```diff
@@ -6255,14 +6255,20 @@ Ext.extend(XDMoD.Module.MetricExplorer, XDMoD.PortalModule, {
...
-            var encodedData = window.btoa(JSON.stringify(this.getConfig()));
-            var link = window.location.protocol + '//' + window.location.host + '/#main_tab_panel:metric_explorer?config=' + encodedData;
-            var msg = 'Use the following link to share the current chart. Note that the link does not override the access controls. So if you send the link to someone who does not have access to the data, they will still not be able to see the data. <br> We recommend using Chrome or Firefox if the link does not work in Internet Explorer.<br><b>' + link + '</b>';
+            const encodedData = window.btoa(JSON.stringify(this.getConfig()));
+            const link = `${window.location.protocol}//${window.location.host}/#main_tab_panel:metric_explorer?config=${encodedData}`;
+            const msg = `Use the following link to share the current chart. Note that the link does not override the access controls. So if you send the link to someone who does not have access to the data, they will still not be able to see the data. <br> We recommend using Chrome or Firefox if the link does not work in Internet Explorer.<br><b>${link}</b>`;
```
aaronweeden, line 6260: Do we still need the recommendation for Internet Explorer?

aestoltm: I have not tested on Edge but I think that would be the alternative. Although it is the continuation of Internet Explorer. I wonder if we can get rid of the statement about the web browser now that we only support modern browsers.

jpwhite4: I'd get rid of the statement recommending any browser

aaronweeden (approved):

PR #1865, opened 2024-06-11
jQuery partial revert

aestoltm:
## Description

Bring back jQuery 3.7.1 for only `deepExtend` calls due to issues found with `youmightnotneedjquery` implementation.

## Motivation and Context

Job Viewer had errors due to cyclical data structure on an extend call.

## Tests performed

Tested affected job on xdmod-dev

## Checklist:


html/internal_dashboard/index.php
```diff
@@ -52,6 +52,7 @@
 
   <script type="text/javascript" src="../gui/js/Error.js.php"></script>
   <script type="text/javascript" src="../gui/js/globals.js"></script>
+  <script type="text/javascript" src="_html_dir_/gui/lib/jquery/jquery-3.7.1.min.js"></script>
```
jpwhite4, line 55: This looks wrong to me. Shouldn't it be ../gui like all the other script includes?  I don't think we have code that autosubstitutes the _html_dir_ string.

jpwhite4 (changes requested):

html/internal_dashboard/index.php
aestoltm: It should be `../gui`. Updating now

jpwhite4 (approved):

PR #1869, opened 2024-06-13
Fix Metric Explorer title when set to an empty string

aestoltm:
## Description

`classes/Datawarehouse/Access/Common.php` sets empty strings for the title to `null`. This will cause the annotation for the title to display 'new text' which is typically displayed when a chart text component crashes.

## Motivation and Context

Bugfix for empty string passed as Metric Explorer title.

## Tests performed

Tested on dev.

## Checklist:


classes/DataWarehouse/Access/Common.php
```diff
@@ -198,9 +198,9 @@ protected function getFontSize()
     protected function getTitle()
     {
         return
-            isset($this->request['title']) && $this->request['title'] != ''
+            isset($this->request['title'])
```
jpwhite4, line 201: Is it necessary to change this? I'm concerned there might be an inintended side effect somewhere else.

aestoltm: Moved to logic for handling empty titles to `html/gui/js/PlotlyChartWrapper.js`.

jpwhite4 (approved):

PR #1875, opened 2024-06-25
Fixing misnamed variable

ryanrath:
## Description
This variable was misnamed which causes an Exception to be thrown at runtime.

## Checklist:

### DIFF
--- a/classes/DB/PDODBMultiIngestor.php
+++ b/classes/DB/PDODBMultiIngestor.php
@@ -452,7 +452,7 @@ class PDODBMultiIngestor implements Ingestor
     protected function loadProcessingConfig($configFilePath)
     {
         if (! is_file($configFilePath)) {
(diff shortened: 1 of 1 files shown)


classes/DataWarehouse/Query/ResourceSpecifications/RawData.php
```diff
@@ -23,7 +23,8 @@ public function __construct(
         $endDate,
         $groupById = null,
         $statisticId = null,
-        array $parameters = array()
+        array $parameters = array(),
+        LoggerInterface $logger = null
```
aaronweeden, line 27: This should be passed into the parent constructor, right?

ryanrath: Addressed in the latest commit

aaronweeden (approved):

PR #1876, opened 2024-06-25
Reducing PHP Warnings

ryanrath:
## Description
This bit of  code is being told to process not-files on PHP 7.4 AppKernels. So adding a quick check to make sure that the fd actually points to a file before processing will make it so those error messages aren't generated.

## Checklist:


ryanrath:
Turns out this breaks the upgrade path. I'll continue looking into this w/ the PHP 8.2 branch.

PR #1879, opened 2024-06-27
Add additional information for 11.0 upgrade

ryanrath:
## Description
Added additional documentation on how to update the xdmod DB user to work when the XDMoD webserver is migrated to a new host. 

## Motivation and Context
It's important for installations that host the XDMoD DB on a different server than the XDMoD Web Server to ensure that the database user XDMoD uses is updated to reflect a change in the XDMoD Web Server's hostname. Failure to do this will result in XDMoD not being able to funciton due to permission denied errors.

## Checklist:

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -50,7 +50,7 @@ section which starts at `Server: EL8, XDMoD: 10.5, PHP: 7.2`.
 4. Import the CentOS 7 exported database files into the Rocky 8 server's database.
     1. `mysql < backup.sql`
 5. **NOTE:** MariaDB / MySQL users are defined as `'username'@'hostname'` so if the hostname of the new Rocky 8 web server is different than the hostname of the old CentOS 7 web server, you will need to make sure that this change is reflected in the database.
-    1. Run the following from an account that has db admin privileges to ensure the XDMoD user is correct: `mysql -e "UPDATE mysql.user SET Host = '<insert new XDMoD web server hostname here>' WHERE username = 'xdmod';"`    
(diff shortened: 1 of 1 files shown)


docs/upgrade.md
```diff
@@ -49,8 +49,10 @@ section which starts at `Server: EL8, XDMoD: 10.5, PHP: 7.2`.
           ``sed -i 's|DEFINER=`xdmod`@`localhost`||g' backup.sql``
 4. Import the CentOS 7 exported database files into the Rocky 8 server's database.
     1. `mysql < backup.sql`
-5. Restart the web server / database on the Rocky 8 server and confirm that everything is working as expected.
-6. Next, follow the upgrade process detailed below on the Rocky 8 Server.
+5. **NOTE:** MariaDB / MySQL users are defined as `'username'@'hostname'` so if the hostname of the server that the user is connecting from changes, you will need to make sure that this change is reflected in the database.
```
aaronweeden, line 52:
```suggestion
5. **NOTE:** MariaDB / MySQL users are defined as `'username'@'hostname'` so if the hostname of the Rocky 8 web server is different than the hostname of the CentOS 7 web server, you will need to make sure that this change is reflected in the database.
```

docs/upgrade.md
aestoltm, line 1: I believe we want to specify the mongo version here. From the notes when you were helping with staging it looks like specifically `$ yes '' | pecl install mongodb-1.18.1`

aestoltm (changes requested):
From my notes during metrics-staging upgrade

> For the upgrade documentation, we may want to also include reminders for copying over more than just /etc/xdmod/, also copying over the Apache config, sendmail config, etc

ryanrath:
@aestoltm do you happen to have a list of the configs that should be transferred?

aestoltm (approved):

PR #1883, opened 2024-07-15
Fix for charts with `side-by-side` grouping (Bar/Area plots)

aestoltm:
Risk: MEDIUM
Notes on Risk: These changes affect Area/Bar plots due to both of these making use of the stacking feature. Minor changes to trace ordering which affects every plot. The change on trace ordering should be lower risk because the  `zIndex` is not being used due to `layering` being currently disabled.

Severity: MEDIUM/HIGH
Notes on Severity: Depending on the units of the data and if the data sources are from the same realm or not the plot has misplaced traces, due to wrongful stacking (instead of side-by-side). See attached plots.

## Description

Side by side bar charts were broken when comparing time-series data from two different realms.

Before Fix:
Screenshot from 2024-07-15 12-05-24
After Fix:
Screenshot from 2024-07-15 12-05-39

Found an additional bug with legend ordering w/ trendline and error bars enabled:
Before: 
Screenshot from 2024-07-15 16-19-28
After:
Screenshot from 2024-07-15 16-19-41

## Motivation and Context

Fixes time-series grouped (side by side) bar charts.

## Tests performed

Tested on dev and tested grouped bar charts with data from the same realm to check for regressions

## Checklist:


classes/DataWarehouse/Access/Usage.php
```diff
@@ -848,7 +845,7 @@ function ($drillTarget) {
                             && $chartSortedByValue
                             && $usageGroupBy !== 'none'
                         ) {
-                            $rank = $meDataSeries['legendrank']+1;
+                            $rank = $meDataSeries['legendrank'] / 2;
```
jpwhite4, line 848: We've had problems in the past using floats because their representation changes between php and js versions.

Is `$rank` supposed to be an int? is it guaranteed to be an int?

aestoltm: I need to take another look at this. It doesn't seem to be guaranteed to be an int at the moment.

aestoltm: `$rank` here is guaranteed to be an `int` after taking a closer look. This is only run when the series is a part of a `timeseries` plot and it is based off a variable we control, `$legendRank`. 

During my investigation I noticed I was incrementing the `$legendRank` off by one. However, it was not noticeable because any series with a "tied" `$legendRank` would be broken by the fact that one of the traces came later in the data array passed with the plot. Therefore, I reverted the refactor to the Aggregate chart `$legendRank` I did last week and I put the timeseries `$legendRank` to increment by 3 (because we have the primary trace and two possible supplemental traces).

html/gui/js/PlotlyChartWrapper.js
```diff
@@ -97,10 +97,9 @@ XDMoD.utils.createChart = function (chartOptions, extraHandlers) {
         // for comparison idea
         baseChartOptions.data.sort((trace1, trace2) => {
             if (baseChartOptions.layout.barmode !== 'group') {
-                return Math.sign(trace2.zIndex - trace1.zIndex) || Math.sign(trace2.traceorder - trace1.traceorder);
+                return Math.sign(trace2.traceorder - trace1.traceorder);
```
jpwhite4, line 100: Please change this if to explicitly check for a column bar chart (rather than not == group which will match scatter plots and pie charts too).

aestoltm: This should be addressed.

For the trendline code that contained the ternaries, it seems that having the valid check was unnecessary. Nothing breaks when the log scale and std error bars are enabled with the trendline series. I need to start working on a few refactor PRs for Plotly because I saw a few cases where we can refactor `barmode` (only need for bar charts shouldn't be default on every chart) and allowing log scale with error bars (haven't found any reason why it shouldn't be allowed, I can look back on the git blame and try to find a reason)

aestoltm:
Correcting the order of the trendline with the other series (primary and error) exposes the out-of-order timeseries bug that was introduced in PR #1899 . Therefore, I may need to get PR #1960 to merged first to fix that issue before merging this PR.

PR #1885, opened 2024-07-15
Fix pie chart data lables

aestoltm:
Risk: LOW
Severity: MEDIUM (heavily affects readability of some pie charts)
## Description

These changes fix the pie chart margin for when very long data labels are showing.

Also discovered that pie chart margin for some of the interactive charts have clipping with the data labels. This fix is non-trivial and will require some more effort in a separate PR. This is because the margin for pie charts may need to be re-worked entirely due to interactions with the subtitle, and different legend locations.

## Motivation and Context

Pie charts with long data labels were barely visible

## Tests performed

Tested on xdmod-dev

## Checklist:


classes/DataWarehouse/Visualization/AggregateChart.php
```diff
@@ -983,13 +983,20 @@ public function configure(
...
                     $labelLimit = 12;
                     $labelsAllocated = 0;
                     $pieSum = array_sum($yValues);
                     for ($i = 0; $i < count($xValues); $i++) {
-                        if ($isThumbnail || (($labelsAllocated < $labelLimit) && (($yValues[$i] / $pieSum) * 100) >= 2.0)) {
-                            $text[] = '<b>' . $xValues[$i] . '</b><br>' . number_format($yValues[$i], $decimals, '.', ',');
+                        if ($isThumbnail || ($labelsAllocated < $labelLimit && (($yValues[$i] / $pieSum) * 100) >= 2.0)) {
+                            $label = $xValues[$i];
+                            // Truncate long data labels to improve visibility.
+                            if (strlen($xValues[$i]) >= 70) {
```
jpwhite4, line 996:
```suggestion
                            if (mb_strlen($label) >= 70) {
```

jpwhite4: Also why only keep 40 characters but have the threshold set to 70 bytes?

aestoltm: I'm not sure why I had those set so differently. Should be addressed now.

jpwhite4 (approved):

PR #1888, opened 2024-07-19
Added Usage tab link loading for statistics link

t-scholtz:
Added capability so that when you load a chart link from the usage tab ie https://xdmod.access-ci.org/index.php#tg_usage?node=statistic_Jobs_nodecount_utilization the usage tab will load that chart, instead of loading the default chart configuration

Note - this does not handle the case for group charts ie https://xdmod.access-ci.org/index.php#tg_usage?node=group_by_Jobs_nodecount which are handled like the old case and just load the default chart instead.

## Motivation and Context
This allows users to share usage tab charts, as well as adds the production effort in creating an XDMod Chatbot

### DIFF
--- a/html/gui/js/modules/Usage.js
+++ b/html/gui/js/modules/Usage.js
@@ -965,6 +965,57 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
 
             var root = tree.getRootNode();
 
+             //Try grab url and load chart accordingly, otherwise load default
+            const url = window.location.href;
+            if (url) {
+                 const substringToRemove = /https:\/\/[^\/]+\/#tg_usage\?node=/;
+                 let chartSettings = url.replace(substringToRemove, '');
+                 if( chartSettings.includes('statistic_')){
+                 const realmsMetric = root.childNodes;
+                 let selectedRealmMetric = null;
+
+                 for (let realm of realmsMetric) {
+                     let transformedId = realm.id.replace('group_by_', 'statistic_');
+                     if (chartSettings.includes(transformedId)) {
+                         selectedRealmMetric = realm;
+                         chartSettings = chartSettings.replace(transformedId+'_', '');
+                         break;
+                     }
+                }
+                if (selectedRealmMetric){
+                    tree.expandPath(selectedRealmMetric.getPath(), null, function (success) {
+                            if (success) {
+                            var jobCountNode = selectedRealmMetric.findChild("statistic", chartSettings);
+                            if (jobCountNode && !jobCountNode.disabled) {
+                                tree.getSelectionModel().select(jobCountNode);
+                                return;
+                            }
+                        }
+                    });
+                }
+                }
+                else{
+                    defaultSelectFirstNode();
+                }
+            } else if (root.hasChildNodes()) {
+
+                defaultSelectFirstNode();
+
+            } //if (root.hasChildNodes())
+
+        } //selectFirstNode
+
+        // ---------------------------------------------------------
+
+        function defaultSelectFirstNode() {
+            updateDisabledMenus.call(this, true);
+
+            var node = tree.getSelectionModel().getSelectedNode();
+
+            if (node != null) return;
+
+            var root = tree.getRootNode();
+
             if (root.hasChildNodes()) {
 
                 var child = root.findChildBy(function (n) {
@@ -1004,8 +1055,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                 } //if(child)
 
             } //if (root.hasChildNodes())
-
-        } //selectFirstNode
+        } //defaultSelectFirstNode
 
         // ---------------------------------------------------------
 


html/gui/js/modules/Usage.js
```diff
@@ -1003,7 +1002,56 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
+                        return true;
+                    }
+                    return false;
+                });
+
+                    if (selectedRealmMetric) {
+                        tree.expandPath(selectedRealmMetric.getPath(), null, (success) => {
+                                if (success) {
+                                const jobCountNode = selectedRealmMetric.findChild('statistic', chartSettings);
+                                if (jobCountNode && !jobCountNode.disabled) {
```
aestoltm, line 1043: Should add in 'else' clause to revert to the default node if this evaluates to false. Currently it will just spin indefinitely if it cannot find the child node.

html/gui/js/modules/Usage.js
```diff
@@ -1003,7 +1002,56 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
+                                    tree.getSelectionModel().select(jobCountNode);
+                                }
+                            }
+                        });
+                    }
```
aestoltm, line 1048: Need to revert to default node if `selectedRealmMetric` evaluates to false.

html/gui/js/modules/Usage.js
```diff
@@ -1003,7 +1002,56 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
+            if (url) {
+                const substringToRemove = /https:\/\/[^/]+\/#tg_usage\?node=/;
+                let chartSettings = url.replace(substringToRemove, '');
+                if (chartSettings.includes('statistic_')) {
+                const realmsMetric = root.childNodes;
+                let selectedRealmMetric = null;
+
+                realmsMetric.some((realm) => {
+                    const transformedId = realm.id.replace('group_by_', 'statistic_');
+                    if (chartSettings.includes(transformedId)) {
```
aestoltm, line 1031: This can throw a false positive on scenarios where the dimension contains a partial match. For example, if you were trying to reload 'Resource Specifications by Resource Institution Country' you would actually match 'Resource Specification by Resource" first. 

We will need more rigorous pattern matching or we need to alter the URL to denote the Realm - Group By - Statistic so that we can more easily pull that information from the URL.

html/gui/js/modules/Usage.js
```diff
@@ -1003,7 +1002,56 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
 
                 } //if(child)
 
-            } //if (root.hasChildNodes())
+            } // if (root.hasChildNodes())
+        } // defaultSelectFirstNode
+
+        // ---------------------------------------------------------
+
+        function selectFirstNode() {
```
aestoltm, line 1010: This function has some indentation issues, and I believe other linter issues that need to be addressed before merging.

html/gui/js/modules/Usage.js
```diff
@@ -1015,7 +1063,6 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
         });
 
         tree.on('expandnode', function (node) {
-
```
aestoltm, line 1018: Please add this line back so we do not include unnecessary changes

tests/ui/test/specs/xdmod/usageTab.js
aestoltm, line 1: I will take a look at this and see how we can alter the existing test cases to pass correctly.

aestoltm (changes requested):

html/gui/js/modules/Usage.js
```diff
@@ -1003,19 +1002,82 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
+                        });
+                    } else {
+                        defaultSelectFirstNode();
+                    }
+                } else {
+                    defaultSelectFirstNode();
+                }
+            } else if (root.hasChildNodes()) {
+                defaultSelectFirstNode();
+            }
```
aestoltm, line 1069:
```suggestion
            const substringToRemove = /https:\/\/[^/]+\/#tg_usage\?node=/;
            let treeNode = url.replace(substringToRemove, '');
            let selectedDimension = null;
            const dimensions = root.childNodes;
            dimensions.sort((a, b) => b.id.length - a.id.length);
            dimensions.some((dimension) => {
                if (treeNode.startsWith('group_by_') && treeNode === dimension.id) {
                    return tree.getSelectionModel().select(dimension);
                }
                const transformedId = dimension.id.replace('group_by_', 'statistic_');
                if (treeNode.includes(transformedId)) {
                    selectedDimension = dimension;
                    return true;
                }
            });
            if (selectedDimension) {
                tree.expandPath(selectedDimension.getPath(), null, (success) => {
                    if (success) {
                        let jobCountNode = selectedDimension.childNodes.find((elem) => elem.attributes.id === treeNode);
                        if (jobCountNode && !jobCountNode.disabled) {
                            return tree.getSelectionModel().select(jobCountNode);
                        }
                        defaultSelectFirstNode();
                    } else {
                        defaultSelectFirstNode();
                    }
                });
            } else {
                defaultSelectFirstNode();
            }
```

aestoltm (changes requested):
Left a suggestions for a minor refactor when we know what dimension the url is referring to. There are also a few changes to some of the variable names so they are more consistent with the XDMoD language. I also included a small change so that this can work with the 'Summary Overview' charts. Even though we most likely will not support those charts with the chat bot I think it would be nice to support in the portal.

Please let me know if you have any questions on the suggested changes.

html/gui/js/modules/Usage.js
```diff
@@ -1003,19 +1002,77 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
...
+                });
+            } else {
+                defaultSelectFirstNode();
+            }
         } //selectFirstNode
 
         // ---------------------------------------------------------
 
         tree.loader.on('load', selectFirstNode, this, {
+
```
aestoltm, line 1070:
```suggestion
```

html/gui/js/modules/Usage.js
```diff
@@ -1015,7 +1071,6 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
         });
 
         tree.on('expandnode', function (node) {
```
aestoltm, line 1073:
```suggestion
        tree.on('expandnode', function (node) {

```

html/gui/js/modules/Usage.js
```diff
@@ -955,8 +955,7 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
 
         // ---------------------------------------------------------
 
-        function selectFirstNode() {
-
+        function defaultSelectFirstNode() {
```
aestoltm, line 958:
```suggestion
        function defaultSelectFirstNode() {
        
```

aestoltm:
The regex is throwing a sonar warning. Need to confirm if it needs to be adjusted.

jpwhite4:
> The regex is throwing a sonar warning. Need to confirm if it needs to be adjusted.

We shouldn't be using a regex in this code. We already have a library function that will parse the location string and get what we need out of it - since thats what the code already does!

aestoltm:
Tests will fail until I update some of the expected files for integration tests

aaronweeden (approved):

PR #1889, opened 2024-07-29
Changes to support splitting CI build into multiple containers.

ShixinWu16:
### Changes to support XDMoD database on an external host
- Additional documentation needed: 
  - Note that when upgrading to this version, after installation, but before running xdmod-upgrade, they need to run xdmod-setup and go through the database setup section. 

#### Added a check for if the user was created
- classes/CCR/DB/MySQLHelper.php
- classes/OpenXdmod/Shared/DatabaseHelper.php

#### Added a question to get XDMoD host
- classes/CCR/DB/MySQLHelper.php
- classes/OpenXdmod/Setup/DatabaseSetup.php

#### Changed from using bootstrap to entrypoint (in create docker images repo)
- tests/ci/bootstrap.sh

~~#### Updated collation from utf8_general_ci to utf8_unicode_ci~~
- ~~configuration/etl/etl_tables.d/*~~
- ~~tests/artifacts/xdmod/etlv2/configuration/input/*~~
- ~~tests/artifacts/xdmod/etlv2/configuration/output/*~~
- ~~tests/unit/lib/ETL/DbModel/*~~
*Update ryanrath: The above changes were merged in a different PR and as such have been removed from this PR*

## Motivation and Context
We wanted to split up the docker containers so that each module has their own container.
This would allow us to share our docker images without having to share everything.

## Tests performed
Ran all automated tests, changed tests to match new collation sets

## Checklist:
- [ X ] The pull request description is suitable for a Changelog entry
- [ X ] The milestone is set correctly on the pull request
- [ X ] The appropriate labels have been added to the pull request


jpwhite4:
As per our conversation: Since the mysql gzip files are output artifacts from the build we should not add them to the source repository. The plan wil be to modify the docker image creation script that is in ccr-private so that it generates two images (1) the image that contains the current release of XDMoD with webserver, php, but not the database and (2) the image that contains the database.

This pull request will then be modified to use these two images in the CI build.

PR #1890, opened 2024-07-29
API changes to allow for public user.

nsafwan:
## Description

 
Allows users without an authentication token to access the raw data API. Migrating the dw_descripter to the Rest stack through the route `/rest/v1/warehouse/search/dw_descripter`. Added new test cases for the new dw_descripter endpoint and modifying the test for the raw data endpoint to now pass for requests with empty token in accordance to the modifications. 

Two new exception classes were added for token authentication:
- `classes/Rest/Exceptions/BadTokenException.php`
- `classes/Rest/Exceptions/EmptyTokenException.php`

## Motivation and Context

Since the raw data endpoint is used for the data catalog, this would allow for public/non logged in users to view the entries available to them without a token. Migrating the dw_descripter endpoint allows for this same behavior and places it in the same place as the rest of the API functionality. Adding the new exception classes separates `UnauthorizedHttpException` for more nuanced behavior depending on empty or invalid token. 

## Tests performed

Automated tests. Modified tests to test for the changed functionality and added new test for the migrated endpoint. 

## Checklist:

### DIFF
--- a/classes/Rest/Controllers/WarehouseControllerProvider.php
+++ b/classes/Rest/Controllers/WarehouseControllerProvider.php
@@ -5,6 +5,7 @@ namespace Rest\Controllers;
 use CCR\DB;
 use CCR\Log;
 use Configuration\XdmodConfiguration;
+use DataWarehouse;
 use DataWarehouse\Data\BatchDataset;
 use DataWarehouse\Export\RealmManager;
 use DataWarehouse\Query\Exceptions\AccessDeniedException;
@@ -336,6 +337,9 @@ class WarehouseControllerProvider extends BaseControllerProvider
 
         $controller
             ->get("$root/raw-data", "$current::getRawData");
+
+        $controller
+            ->get("$root/search/dw_descripter", "$current::getDwDescipter");
     }
 
     /**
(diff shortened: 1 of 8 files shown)


classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -336,6 +340,9 @@ public function setupRoutes(Application $app, ControllerCollection $controller)
 
         $controller
             ->get("$root/raw-data", "$current::getRawData");
+
+        $controller
+            ->get("$root/search/dw_descripter", "$current::getDwDescipter");
```
ryanrath, line 345: Looks like there's a missing an `r` here.

classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -2494,4 +2501,143 @@ private function validateRawDataFilterParam(
         $filterValuesArray = explode(',', $filterValuesStr);
         return $filterValuesArray;
     }
+
+
+    public function getDwDescipter(Request $request, Application $app){
```
ryanrath, line 2506: 'nother missing `r`. 

I'm also guessing that the duplicate code warning is due to this being pulled from `get_dw_descripter.php` ?

tests/integration/lib/Rest/WarehouseExportControllerProviderTest.php
```diff
@@ -200,7 +200,8 @@ public function testGetRealmsTokenAuth($role, $tokenType) {
                 'params' => null,
                 'data' => null,
                 'endpoint_type' => 'rest',
-                'authentication_type' => 'token_optional'
+                'authentication_type' => 'token_optional',
+                'wantPublicUser' => true
```
ryanrath, line 204: I'd try and keep with the snake_case of the other params here.

tests/integration/lib/TokenAuthTest.php
```diff
@@ -150,7 +150,14 @@ public function runTokenAuthTest(
                         )
                     ];
                 } elseif ('rest' === $input['endpoint_type']) {
-                    $output = parent::validateAuthorizationErrorResponse(401);
+                    // If token is empty and we want Public user, test that it returns a success response.
```
ryanrath, line 153: Same as above, snake case.

ryanrath (changes requested):
Just a few

classes/Rest/Controllers/WarehouseControllerProvider.php
nsafwan: Yes

nsafwan (commented):
Missing 'r' and snake_case for want_public_user have been fixed. And yes, that check is failing because of pulled code from `get_dw_descripter.php`.

PR #1891, opened 2024-08-01
Resolve Dependabot Alerts

ryanrath:
**Note: The upgrade won't pass because the patch files have lines that have trailing whitespace in them. But if the whitespace is removed the patches no longer work soooo... yeah, that's a thing.** 

## Description
- `.circleci/config.yml`:
  - Updated the pecl install for `mongodb` to specifically install version `1.19.3`, which is the current version, as we've been bitten more than once by not explicitly specifying which version we want. A point of discussion for this could be do we want to install the latest dependencies? If so, *where* do we want to install these dependencies? i.e. in the CI build? In the production build? etc. 
- `BaseControllerProvider.php`: the namespace for the `ControllerProviderInterface;` interface changed in the latest version of Silex ( from `Silex\ControllerProviderInterface` to `Silex\Api\ControllerProviderInterface`). 
- `XdmodApplicationFactory.php`: 
  - `UrlGeneratorServiceProvider` has been changed to `RoutingServiceProvider`. 
  - there is no more `app->share` function, you just use an anonymous function.
  - the `Request $request` argument has been added to the anonymous function called by the `$app->error` helper      
    function. 
- `composer.json`: 
  - Updated silex to the latest ( last ) version `v2.3.0`, this resolves the following dependabot alerts:
    - https://github.com/ubccr/xdmod/security/dependabot/10 (CVE-2022-24894)
      - Here's the commit with the fix: https://github.com/symfony/symfony/commit/d2f6322af9444ac5cd1ef3ac6f280dbef7f9d1fb
      - And here is where you can find the updated file: `vendor/symfony/http-kernel/HttpCache/Store.php`. Specifically lines `49-51`, and `228-230`.
    - https://github.com/ubccr/xdmod/security/dependabot/11 (CVE-2019-18888)
      - this is fixed as the current version `v4.4.49` is >= `3.4.35`. 
      - Here's the commit w/ the fix: https://github.com/symfony/symfony/commit/691486e43ce0e4893cd703e221bafc10a871f365
      - And here's the updated file: `vendor/symfony/http-foundation/File/MimeType/FileBinaryMimeTypeGuesser.php`
    - https://github.com/ubccr/xdmod/security/dependabot/12 (CVE-2019-10913)
      - Here's the commit w/ the fix: https://github.com/symfony/symfony/commit/944e60f083c3bffbc6a0b5112db127a10a66a8ec
      - Here's the updated file: `vendor/symfony/http-foundation/Request.php`
[...]

PR #1893, opened 2024-08-06
Patch for symfony/http-foundation: CVE-2019-10913

ryanrath:
This patch sets up the patch pipeline in `build.json` and updates a few files that will cause issues during the CI / QA tests / build process. These changes originated in https://github.com/ubccr/xdmod/pull/1891 ( which will be merged in 11.5 ).

## Description
- `open_xdmod/modules/xdmod/assets/setup.sh`:
  - Removed the Composer Install section as this + the dangling `cd` was causing the composer dependencies of `qa` to be installed in the xdmod directory. This probably wasn't a problem or caught before because we were not previously patching files.
  - Changed the `cd $xdmod_dir`to a `pushd | popd` so that we reset the cwd after the script runs.
- `open_xdmod/modules/xdmod/build.json`
  - added entries to support patching files. This necessitated adding the `rm -rf vendor/` and `composer install` due to the build step occurring twice in the CI process. 
- `tests/ci/scripts/qa-test-setup.sh`:
  - Taking care of the other part of the **Case of the Mysterious Disappearing Dependencies**. Just made sure to save the existing `$COMPOSER` env variable so that we can restore it after we're done installing the qa dependencies. 

## Motivation and Context
Less holes == more good. 

## Tests performed
All Automated Tests passed. 

## Checklist:

### DIFF
--- a/open_xdmod/modules/xdmod/assets/setup.sh
+++ b/open_xdmod/modules/xdmod/assets/setup.sh
@@ -7,10 +7,8 @@ assets_dir="$(
 module_dir="$assets_dir/.."
 xdmod_dir="$module_dir/../../.."
 
-echo Installing composer managed dependencies
-cd $xdmod_dir
-composer install --no-dev
-
+pushd "$xdmod_dir" || exit
 echo Installing npm managed dependencies
 npm install --production --prefix etl/js
 npm install --production --prefix background_scripts/chrome-helper
+popd || exit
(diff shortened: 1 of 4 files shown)


open_xdmod/modules/xdmod/assets/setup.sh
```diff
@@ -7,10 +7,8 @@ assets_dir="$(
 module_dir="$assets_dir/.."
 xdmod_dir="$module_dir/../../.."
 
-echo Installing composer managed dependencies
-cd $xdmod_dir
-composer install --no-dev
-
+pushd "$xdmod_dir" || exit
```
aaronweeden, line 10: Should this exit with an error code?

ryanrath: Looks like it defaults to 1, are you thinking we need a different code?

aaronweeden: Nope, I forgot exit by default returns the code of the last command that ran.

aaronweeden (approved):

PR #1895, opened 2024-08-07
Refactor REST warehouse raw data realm validation.

aaronweeden:
## Description

This PR refactors the validation of the `realm` parameter for REST requests to the warehouse raw data endpoint.

Now, instead of responding with a message of `Invalid realm`, the message provides more information depending on the reason for the error:
* `No realm exists with the requested name.`
* `The requested realm is not configured to provide raw data.`
* `Your user account does not have permission to get raw data from the requested realm.`

In the last example above, it now responds with `403 Forbidden` instead of `400 Bad Request`.

## Tests performed

In addition to updating the expected message in an integration test, I also updated my port on `xdmod-dev` and made sure the following tests passed with different roles and different realms, making the following request (replacing `<token>` with the user's token and `<realm>` with the realm):
```
GET https://xdmod-dev.ccr.xdmod.org:9001/rest/warehouse/raw-data?realm=<realm>&start_date=2024-01-01&end_date=2024-01-01&Bearer=<token>
```
| Top Role | Realm | Result |
| - | - | - |
| User | foo | No realm exists with the requested name. |
| User | Jobs | Success |
| User | Cloud | Success |
| User | Accounts | Success |
| User | Allocations | Success |
| User | ResourceAllocations | Your user account does not have permission to get raw data from the requested realm. |
| User | Requests | The requested realm is not configured to provide raw data. |
| User | Gateways | The requested realm is not configured to provide raw data. |
| User | SUPREMM | Success |
| User | JobEfficiency | The requested realm is not configured to provide raw data. |
| User | ResourceSpecifications | Success |
| User | Introspection | The requested realm is not configured to provide raw data. |
| User | PCPArchives | The requested realm is not configured to provide raw data. |
| User | OnDemand | The requested realm is not configured to provide raw data. |
| Program Officer | foo | No realm exists with the requested name. |
| Program Officer | Jobs | Success |
| Program Officer | Cloud | Success |
| Program Officer | Accounts | Success |
| Program Officer | Allocations | Success |
| Program Officer | ResourceAllocations | Success |
| Program Officer | Requests | The requested realm is not configured to provide raw data. |
| Program Officer | Gateways | The requested realm is not configured to provide raw data. |
| Program Officer | SUPREMM | Success |
| Program Officer | JobEfficiency | The requested realm is not configured to provide raw data. |
| Program Officer | ResourceSpecifications | Success |
| Program Officer | Introspection | The requested realm is not configured to provide raw data. |
| Program Officer | PCPArchives | The requested realm is not configured to provide raw data. |
| Program Officer | OnDemand | The requested realm is not configured to provide raw data. |

## Checklist:

### DIFF
--- a/classes/Rest/Controllers/WarehouseControllerProvider.php
+++ b/classes/Rest/Controllers/WarehouseControllerProvider.php
@@ -2126,9 +2126,8 @@ class WarehouseControllerProvider extends BaseControllerProvider
      *                                 not included; if an invalid start date,
      *                                 end date, realm, field alias, or filter
      *                                 key is provided; if the end date is
-     *                                 before the start date; if the offset is
-     *                                 negative; or if the requested realm is
-     *                                 not configured to provide raw data.
+     *                                 before the start date; or if the offset
+     *                                 is negative.
      */
     public function getRawData(Request $request, Application $app)
     {
@@ -2136,9 +2135,6 @@ class WarehouseControllerProvider extends BaseControllerProvider
         $params = $this->validateRawDataParams($request, $user);
         $realmManager = new RealmManager();
         $queryClass = $realmManager->getRawDataQueryClass($params['realm']);
-        if (!class_exists($queryClass)) {
-            throw new BadRequestHttpException('The requested realm is not configured to provide raw data.');
-        }
         $logger = $this->getRawDataLogger();
         $streamCallback = function () use (
             $user,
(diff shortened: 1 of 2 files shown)


classes/Rest/Controllers/WarehouseControllerProvider.php
```diff
@@ -2213,9 +2213,27 @@ private function validateRawDataParams($request, $user)
...
+        );
+        if (!in_array($params['realm'], $allBatchExportRealms)) {
+            throw new BadRequestHttpException(
+                'The requested realm is not configured to provide raw data.'
+            );
+        }
         $queryDescripters = Acls::getQueryDescripters($user, $params['realm']);
         if (empty($queryDescripters)) {
-            throw new BadRequestHttpException('Invalid realm.');
+            throw new AccessDeniedException(
```
ryanrath, line 2233: Is there a test that exercises this code path? If not could you add one?

ryanrath (changes requested):
Just the one note re: adding a test that exercises the new `AccessDeniedException` path in `validateRawDataParams`.

classes/Rest/Controllers/WarehouseControllerProvider.php
aaronweeden: I don't think I can since none of the realms in the Docker container are restricted to particular roles.

ryanrath: We should probably add a dummy realm that is restricted to some subset of users. Doesn't have to be for this PR of course, but we should certainly strive to cover more code paths that we know can be reached. Because if the code path can't be reached then or then we should just remove the code ( or modify the code so that it can ).

ryanrath (approved):

PR #1896, opened 2024-08-09
Adding logic to catch invalid datetime values

ryanrath:
## Description
Additional Validation of both ISO8601 Date & Unix Timestamp parameters REST parameters have been added. 

## Motivation and Context
Aaron noted in https://app.asana.com/0/159049597309611/1208001546005418/f that it seemed like invalid values ( ex. `2024-01-99`) were being accepted for date parameters provided to rest endpoints.

These changes add additional validation to both ISO8601 Date & Unix Timestamp parameters where `DateTime::getLastErrors()` is checked to see if there are any warnings or errors before possibly returning a value.

New tests were added per conversation with Aaron.

## Tests performed
All automated tests were run / passed. 

## Checklist:

### DIFF
--- a/classes/Rest/Controllers/BaseControllerProvider.php
+++ b/classes/Rest/Controllers/BaseControllerProvider.php
@@ -561,11 +561,18 @@ abstract class BaseControllerProvider implements ControllerProviderInterface
                     $value_dt = \DateTime::createFromFormat('Y-m-d', $value);
 
                     $lastErrors = DateTime::getLastErrors();
-                    if ($lastErrors['warning_count'] > 0 || $lastErrors['error_count'] > 0) {
-                        throw new BadRequestHttpException("Invalid value for $name. Must be a(n) ISO 8601 Date.");
+
+                    // Depending on the version of PHP the value returned is different. This makes one less change needed
+                    // to support PHP 8.* in the future.
+                    // https://www.php.net/manual/en/datetimeimmutable.getlasterrors.php#refsect1-datetimeimmutable.getlasterrors-changelog
+                    if (version_compare(PHP_VERSION, '8.2.0') >= 0)
+                    {
+                        $hasWarningsOrErrors = $lastErrors;
+                    } else {
+                        $hasWarningsOrErrors = $lastErrors['warning_count'] > 0;
                     }
 
-                    if ($value_dt === false) {
+                    if ($value_dt === false || $hasWarningsOrErrors) {
                         return null;
                     }
                     return $value_dt;
--- a/classes/Rest/Controllers/BaseControllerProvider.php
+++ b/classes/Rest/Controllers/BaseControllerProvider.php
@@ -814,9 +814,9 @@ abstract class BaseControllerProvider implements ControllerProviderInterface
      */
     private static function filterDate(string $value, string $format = 'Y-m-d'): ?DateTime
     {
-        $dateTime = \DateTime::createFromFormat($format, $value);
+        $dateTime = DateTime::createFromFormat($format, $value);
 
-        $lastErrors = \DateTime::getLastErrors();
+        $lastErrors = DateTime::getLastErrors();
 
         /* For PHP versions less than 8.2.0 $lastErrors will always be an array w/ the properties:
          * warning_count, warnings, error_count, and errors. For versions >= 8.2.0, it will return false if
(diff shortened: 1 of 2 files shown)


classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -505,8 +505,14 @@ protected function getDateTimeFromUnixParam(Request $request, $name, $mandatory
             $default,
             FILTER_CALLBACK,
             array(
-                "options" => function ($value) {
+                "options" => function ($value) use ($name) {
                     $value_dt = DateTime::createFromFormat('U', $value);
+
+                    $lastErrors = DateTime::getLastErrors();
+                    if ($lastErrors['warning_count'] > 0 || $lastErrors['error_count'] > 0) {
```
aaronweeden, line 512: If there were no errors, will this statement cause `Warning: Trying to access array offset on value of type bool`?

tests/integration/lib/BaseTest.php
```diff
@@ -579,31 +579,37 @@ protected function provideRestEndpointTests(
...
-                                    $type,
-                                    $errorBodyValidator
-                                )
-                            );
-                            $tests[] = $testData;
+                        // We can skip tests:
+                        //   - Strings can be strings, so skip that test.
+                        //   - Invalid dates should only be tested on date params.
+                        if (('invalid_date' === $id && 'date_params' !== $key) ||
+                            ('string' === $id && 'string_params' === $key)) {
```
aaronweeden, line 594: For readability, I would swap the two conditions to match the order of the comments above.

classes/Rest/Controllers/BaseControllerProvider.php
aaronweeden, line 512: Is `$lastErrors['error_count'] > 0` necessary; i.e., is there a way for it to be true and for `$value_dt` not to be false?

aaronweeden (changes requested):

classes/Rest/Controllers/BaseControllerProvider.php
ryanrath: Doesn't seem to, here's an example w/ no errors showing that the property `error_count` is still present. 
```php
$value = '2024-09-99';
$datetime = \DateTime::createFromFormat('Y-m-d', $value);
print_r(DateTime::getLastErrors());

Output:
Array
(
    [warning_count] => 1
    [warnings] => Array
        (
            [10] => The parsed date was invalid
        )

    [error_count] => 0
    [errors] => Array
        (
        )

)
```

ryanrath: No idea, but since it was always present I added it to the `if`, I can remove it. I did find an interesting update to this, apparently the return value changes depending on what version of PHP is used. I've made the necessary updates to future proof the code in an upcoming commit.

tests/integration/lib/BaseTest.php
ryanrath: sure

classes/Rest/Controllers/BaseControllerProvider.php
```diff
@@ -505,8 +505,14 @@ protected function getDateTimeFromUnixParam(Request $request, $name, $mandatory
...
+                        throw new BadRequestHttpException("Invalid value for $name. Must be a(n) Unix timestamp.");
```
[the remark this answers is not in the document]
ryanrath: sure

classes/Rest/Controllers/BaseControllerProvider.php
aaronweeden: I was referring to the case where the value is a valid date; as you note below, this will be `false` in PHP ≥8.2.0.

aaronweeden: Looks like this code needs to be updated to match the code below.

aaronweeden: Maybe best to define a function since it will be duplicated code. Something like:
```php
"options" => function ($value) {
    return self::filterDate('U', $value);
}
// ...
"options" => function ($value) {
    return self::filterDate('Y-m-d', $value);
}
```
```php
private static function filterDate(string $format, string $value)
    $value_dt = DateTime::createFromFormat($format, $value);
    $lastErrors = DateTime::getLastErrors();
    if ($value_dt === false || (is_array($lastErrors) && $lastErrors['warning_count'] > 0)) {
        return null;
    }
    return $value_dt;
}
```

aaronweeden (dismissed):
Since there is a `use DateTime`, the backslashes in front of `\DateTime` aren't necessary, but it's up to you if you want to change those or merge as is.

aaronweeden (approved):

PR #1923, opened 2024-10-09
Cache Bust attempt

jpwhite4:
## Description
The update from 10.5 to 11.0 had substantial changes to various javascript files mostly due to the plotly updates. However, the filenames themselves didn't change. On our production systems, the document lifetime is set to a year, so Safari and Edge browsers  are caching those files and not checking to see if they have been updated (we have etags enabled, but it seems the browsers don't check the etag values, instead they just load from cache).

## Motivation and Context
This happens when you browse to XDMoD 11.0 with cached versions of the 10.5 javascript:

image

Fix is to add a query parameter with the XDMoD version number to all javascript file loads. This means that the browsers see the file from 10.5 as different from 11.0 (e.g. in 10.5 the browser requests "Summary.js" and with this change in 11.0 it will request "Summary.js?version=11.0.0". Which is a different filename and will cause the browser to request it since its not cached. The query parameters are ignored by the webserver when serving a normal file so the file contents are not changed by this request structure.

## Tests performed

Installed on my port on dev and confirmed that it will pull it the first time:

image

and then cache it on subsequent pulls:

image

### DIFF
--- a/html/index.php
+++ b/html/index.php
@@ -114,7 +114,7 @@ $page_title = xd_utilities\getConfiguration('general', 'title');
     <title><?php print $page_title; ?></title>
 
     <link rel="shortcut icon" href="gui/icons/favicon_static.ico"/>
-    <script type="text/javascript" src="gui/lib/internet-explorer-polyfills.js"></script>
+    <script type="text/javascript" src="gui/lib/internet-explorer-polyfills.js?version=<?php print OPEN_XDMOD_VERSION; ?>"></script>
     <?php if (!$userLoggedIn): ?>
         <script type="text/javascript">
             /**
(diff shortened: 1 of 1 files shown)


aaronweeden:
Do these also need to be updated?
* xdmod: `html/plotly_template.html`
* xdmod-appkernels: `html/appkernel/index.php`

aaronweeden (approved):

aaronweeden:
Does `classes/ExtJS/ExtJS.php` also need to be edited?

aaronweeden (changes requested):
Do these still need to be updated?
- xdmod: `classes/ExtJS/ExtJS.php`
- xdmod: `configuration/assets.json`
- xdmod-supremm: `configuration/assets.d/supremm.json`
- xdmod-appkernels: `configuration/assets.d/appkernels.json`
- xdmod-xsede: `configuration/assets.d/xsede.json`

PR #1931, opened 2024-10-24
Update software requirements docs to expunge Centos 7

jpwhite4:
Also fix some other errors about ssl config and clarify timezone setting reqs


aaronweeden (dismissed):

ryanrath (dismissed):
Just those two previous comments, but otherwise looks good.

PR #1933, opened 2024-10-25
Expunge more references to CentOS 7 in the docs.

aaronweeden:


jpwhite4:
I think this should go in the xdmod11.0 branch and then we'll upadrte the 11.0 docs. Its currently pointing at 11.5

eiffel777 (approved):

PR #1936, opened 2024-10-28
Adding upgrade pipeline for Person and PI Organization for cloud realm

eiffel777:
This adds the migration pipeline for the columns added in #1958. The `cloud-state-pipeline` etl pipeline has been added to the migration to re-aggregate the cloud realm so these columns are added and the columns are populated. These columns are used in the xsede module at this point in time.

The XDMoD version number also needed to be updated to 11.0.1 in various places.

## Tests performed
Tested in docker.

## Checklist:


classes/OpenXdmod/Migration/Version1050To1100/DatabasesMigration.php
```diff
@@ -24,7 +24,7 @@ public function execute()
 
         if ($mysql_helper->tableExists('modw_cloud.event')) {
             Utilities::runEtlPipeline(
-                ['cloud-migration-10-5-0_11-0-0'],
+                ['cloud-migration-10-5-0_11-0-0','cloud-state-pipeline'],
```
jpwhite4, line 27: Should this be changed to an 11.0.0 -> 11.0.1 migration? We can't be modifying the 10.5 to 11.0 migration anymore since the 11.0.0 is released

jpwhite4 (changes requested):

aaronweeden:
It looks like this should be based on `xdmod11.0` branch instead of `main`.

eiffel777:
This is replaced by #1966

PR #1941, opened 2024-11-04
Fix marker opacity of mostly null data series

aestoltm:
## Description

@ryanrath noticed that data series with hidden markers (due to length of date range) do not show markers when there is only one non-null value. This fix adds an additional check to show the markers if only one non-null data point exists (regardless of the data range). The additional check only looks for one data point because if there are multiple data points within a mostly null dataset there will be "dotted" lines showing the data.

The aggregate charts needed a counter to address this issue.

## Motivation and Context

Charts that have one non null value are not viewable (for long date ranges).

## Tests performed

Tested on xdmod-dev

## Checklist:


aestoltm:
> You set this to the 11.0.1 milestone, but to merge into main - which would be for 11.5. Please confirm which is correct and fix either the milestone or the merge branch.

Updated all my PRs that are set to 11.0.1 milestone to merge into branch `xdmod11.0`.

classes/DataWarehouse/Visualization/TimeseriesChart.php
```diff
@@ -468,7 +468,8 @@ public function configure(
                         // Hide markers for 32 points or greater, except when there are multiple traces then hide markers starting at 21 points.
                         // Need check for chart types that this applies to otherwise bar, scatter, and pie charts will be hidden.
                         $hideMarker = in_array($data_description->display_type, array('line', 'spline', 'area', 'areaspline'))
-                            && ($values_count >= 32 || (count($yAxisDataObjectsArray) > 1 && $values_count >= 21));
+                            && ($values_count >= 32 || (count($yAxisDataObjectsArray) > 1 && $values_count >= 21))
+                            && $y_values_count > 1;
```
jpwhite4, line 472: The pull request comments talk about not null values, but this change does not appear to be related to not null. Please explain?

aestoltm: Refactored the code we discussed.  The logic for the marker visibility threshold is the same as during `10.5` which should have been the implementation from the start. The only difference is we did not track non-null (visible) data points for Aggregate charts in `10.5`.

jpwhite4 (approved):

PR #1947, opened 2024-11-22
Fix various bugs found when testing strict SQL modes

jpwhite4:
This pul request contains fixes for four different issues found when testing strict SQL mode. Each one is a different commit (for ease of review). The full changes needed to support strict SQL modes is a bit more involved - these are just the trivial ones.

## Fix storage aggregator.

    The physical_usage column in the fact table is nullable, but the one in the
    aggregates table is not. This updates the load SQL to coalesce the null to 0 if necessary.

## Fix SQL update string for XDUser

    You can't call SQL function in PDO parameters. This code changes it so the that call to the MySQL
    NOW() function is done good and proper like.

##  Update default values for XDUser's fks

    Null is not an allowed value for the not null foreign keys to the organization
    and person tables. This change initializes them to a valid value (-1 == Unknown)

##   Fix filter in Report Generator

    The filter string was filtering out valid values for the Report Generator
    delivery mode. This results in the valid string "E-mail" being converted to
    a null which then getes converted to "E-mail" on the sql load if the
    sql_mode is set empty. _Sigh_


classes/DataWarehouse/Access/ReportGenerator.php
```diff
@@ -10,7 +10,7 @@ class ReportGenerator extends Common
     const REPORT_DATE_REGEX =  '/^[0-9]{4}(-[0-9]{2}){2}$/';
     const REPORT_FORMATS_REGEX = '/^doc|pdf$/';
     const REPORT_SCHEDULE_REGEX = '/^Once|Daily|Weekly|Monthly|Quarterly|Semi-annually|Annually$/';
-    const REPORT_DELIVERY_REGEX = '/^E-Mail$/';
+    const REPORT_DELIVERY_REGEX = '/^E-mail$/';
```
aestoltm, line 13: Just to confirm for my understanding, this is used to match the "Delivery Method" when saving a report? And the "Delievery Method: E-Mail" on the portal is converted to "E-mail" on the sql load?

aestoltm: That clears up the confusion I had. I was missing the javascript that hardcodes the string to `E-mail`.

aestoltm (approved):

PR #1964, opened 2025-01-16
Add Cloud statistics

jpwhite4:
image


eiffel777:
This is being replaced by https://github.com/ubccr/xdmod/pull/1976 right? That PR adds the same statistics and adds all the tests.

PR #1991, opened 2025-02-26
Add ability to add additional database migration classes.

aaronweeden:
## Description

This PR adds the ability to add additional database migration classes by defining them in the same namespace and ending their names with `DatabasesMigration`.

This is useful for, e.g., https://github.com/ubccr/xdmod-ondemand/pull/84.

## Tests performed

I tested https://github.com/ubccr/xdmod-ondemand/pull/84 in tandem with this PR.

### DIFF
--- a/classes/OpenXdmod/Migration/MigrationFactory.php
+++ b/classes/OpenXdmod/Migration/MigrationFactory.php
@@ -65,7 +65,7 @@ class MigrationFactory
                     $toVersion
                 );
 
-                // Run each of the other DatabasesMigration classes in the
+                // Add each of the other DatabasesMigration classes in the
                 // namespace.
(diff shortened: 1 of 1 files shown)


jpwhite4:
I don't thinkit is a good idea to hardcode submodule names into the core. How about just have the code detect any database migrations classes and run them automatically. (run the additional migrations in alphabetical order by classname so that migrations are deterministic)

aaronweeden:
> I don't thinkit is a good idea to hardcode submodule names into the core. How about just have the code detect any database migrations classes and run them automatically. (run the additional migrations in alphabetical order by classname so that migrations are deterministic)

@jpwhite4 I have updated this.

jpwhite4 (approved):

PR #1992, opened 2025-02-27
Refactor timeseries ticklabels

aestoltm:
## Description

Compute the placement and format of ticklabels using `tickvals` and `ticktext`. This gives us full control of which tick labels are rendered and their values. This is necessary to have the first and last label of the data always showing, along with not having any labels outside of the selected time range. These changes result in not needing to remove any extra tick labels because we dictate what is rendered.

## Motivation and Context

Avoid any confusion about data existing outside the selected time range.

## Tests performed

Tested on my port.

## Checklist:


classes/DataWarehouse/Access/Usage.php
```diff
@@ -756,6 +756,7 @@ public function getCharts(XDUser $user, $chartsKey = 'data') {
                     (isset($meChart['layout']['xaxis']['ticktext']) || isset($meChart['layout']['yaxis']['ticktext']))
                     && $chartSortedByValue
                     && $usageGroupBy !== 'none'
+                    && !$meRequestIsTimeseries
```
aestoltm, line 759: Need this so that we do not rank our timeseries ticklabels. Was not a concern before because we did not make use of `ticktext` and `tickvals` for timeseries charts.

aaronweeden (approved):

PR #1993, opened 2025-02-28
Use specific mongodb version

aestoltm:
## Description

Downloading the current mongodb version with pecl has issues, `yes '' | pecl install mongodb || true` results in:
```
PHP Warning:  PHP Startup: Unable to load dynamic library 'mongodb.so' (tried: /usr/lib64/php/modules/mongodb.so (/usr/lib64/php/modules/mongodb.so: undefined symbol: _zval_ptr_dtor), /usr/lib64/php/modules/mongodb.so.so (/usr/lib64/php/modules/mongodb.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
WARNING: channel "pecl.php.net" has updated its protocols, use "pecl channel-update pecl.php.net" to update
pecl/mongodb requires PHP (version >= 8.1.0, version <= 8.99.99), installed version is 7.4.33
No valid packages found
install failed
```

Our 11.0 [docs](https://open.xdmod.org/11.0/upgrade.html) specify the mongodb version to `1.18.1` which does work. However, pecl php docs show that the latest supported version is 1.20.1

## Motivation and Context

Can't build xdmod without php mongodb package

## Tests performed

Tested in docker container

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
           command: dnf install -y php libzip-devel php-pear php-devel
       - run:
           name: Install MongoDB Pear module
-          command: yes '' | pecl install mongodb-1.18.1 || true
+          command: yes '' | pecl install mongodb-1.20.1 || true
       - run:
             name: install the composer dependencies
             command: composer install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
           command: dnf install -y php libzip-devel php-pear php-devel
       - run:
           name: Install MongoDB Pear module
-          command: yes '' | pecl install mongodb || true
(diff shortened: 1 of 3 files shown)


ryanrath (approved):

jpwhite4:
Is that version of mongo still supported? If not is there a more recent compatible version?

aestoltm:
> Is that version of mongo still supported? If not is there a more recent compatible version?

I believe it is still supported. The pecl php [site](https://pecl.php.net/package/mongodb) shows active maintainers for v1.18.1. However, it does show that the latest version compatible with PHP7.4 is v1.20.1. Therefore, we could update to that version.

jpwhite4:
Please can you also updated the docs/software-requirements.md file to list 1.20 mongo library version.

aaronweeden (approved):

PR #1997, opened 2025-03-06
Use xdmod10.5 logic for enabling markers

aestoltm:
## Description

@aaronweeden found examples where series incorrectly had markers rendered. I looked at the changes make in #1941 and noticed it did not use the same marker logic as in xdmod10.5 (highcharts implementation) which matches our desired behavior. This PR corrects the logic that enables markers to be the same as in 10.5 [timeseries](https://github.com/ubccr/xdmod/blob/b2f8b3e629423a5656532bb578f0871d099e8e4a/classes/DataWarehouse/Visualization/HighChartTimeseries2.php#L451) and [aggregate ](https://github.com/ubccr/xdmod/blob/b2f8b3e629423a5656532bb578f0871d099e8e4a/classes/DataWarehouse/Visualization/HighChart2.php#L1124) charts

## Motivation and Context

Improve the clarity of charts. The incorrectly rendered markers made it difficult to see other series on the same date.

## Tests performed

Tested on my dev port.

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Visualization/TimeseriesChart.php
+++ b/classes/DataWarehouse/Visualization/TimeseriesChart.php
@@ -460,7 +460,7 @@ class TimeseriesChart extends AggregateChart
                         // hidden.
                         // Need check for chart types that this applies to otherwise bar, scatter, and pie charts will be hidden.
                         $showMarker = in_array($data_description->display_type, array('scatter', 'pie', 'bar', 'h_bar', 'column'))
-                            || ($visiblePoints < 21 && !$isThumbnail)
+                            || (count($values) < 21 && !$isThumbnail)
                             || $visiblePoints == 1;
 
                         $isRemainder = $yAxisDataObject->getGroupId() === TimeseriesDataset::SUMMARY_GROUP_ID;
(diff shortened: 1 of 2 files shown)


classes/DataWarehouse/Visualization/TimeseriesChart.php
```diff
@@ -448,6 +448,9 @@ public function configure(
                                 $allNull = false;
                                 $visiblePoints++;
                             }
+                            if ($visiblePoints > 1) {
+                                break;
+                            }
                         }
                         if($allNull)
```
aaronweeden, line 455: This could be `if ($visiblePoints > 0)` and get rid of `$allNull` variable.

aaronweeden (approved):

classes/DataWarehouse/Visualization/TimeseriesChart.php
aaronweeden, line 455:
```suggestion
                        if($visiblePoints > 0)
```
and can get rid of `$allNull` variable.

aestoltm: I think it should be `if($visiblePoints == 0)` so that we know its "allNull" and can skip.

aaronweeden: You're right!

aaronweeden (approved):

PR #2000, opened 2025-03-07
Add explicit transactions to the StructuredFile ingestor

eiffel777:
This adds an explicit transaction for the StructuredFile Ingestor. By default, PDO is set to auto-commit mode meaning that every query that run has its own implicit transaction. If the files you are ingesting has a lot rows there can be significant performance hit because it is writing to the commit log for each insert as well as the database. By setting the transaction before the inserts start auto commit is turned off and when commit is called it will write to the commit log once, providing significant performance improvement of ingestion. 

Many of the changes here are existing code moving to a different line or indenting of existing code because of the addition of a try catch statement.

## Motivation and Context
This change provides significant performance improvements. In testing with some cloud logs the insert time for a file with ~20,000 records went from ~2 minutes to ~10-20 seconds.

## Tests performed
Tested in docker

## Checklist:

### DIFF
--- a/classes/ETL/Ingestor/StructuredFileIngestor.php
+++ b/classes/ETL/Ingestor/StructuredFileIngestor.php
@@ -298,51 +298,59 @@ class StructuredFileIngestor extends aIngestor implements iAction
 
         $warnings = array();
 
-        $this->destinationHandle->beginTransaction();
-
-        foreach ( $this->sourceEndpoint as $sourceRecord ) {
-
-            // The same source record may be used in multiple tables.
-
-            foreach ( $this->destinationFieldMappings as $etlTableKey => $destFieldToSourceFieldMap ) {
-
-                $parameters = $this->generateParametersFromSourceRecord(
-                    $sourceRecord,
-                    $destinationFieldIdToSourceFieldMap[$etlTableKey],
-                    $sourceFieldToValueMapTemplate[$etlTableKey],
-                    $simpleSourceFields[$etlTableKey],
-                    $complexSourceFields[$etlTableKey]
-                );
-
-                try {
-                    // Some values of parameters are complex objects and cannot be! This is due to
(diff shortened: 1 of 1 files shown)


classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -298,43 +298,60 @@ function ($k, $v) {
...
             }
-            $numRecords++;
+
+            $this->destinationHandle->commit();
+        } catch (PDOException $e) {
+            $this->logAndThrowException(
+                "Error committing transaction. Rolling back transactions.",
+                array('exception' => $e, 'endpoint' => $this)
+            );
+            $this->destinationHandle->rollback();
```
aaronweeden, line 354: Is this line reachable if the exception was just thrown on the line above?

classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -298,43 +298,60 @@ function ($k, $v) {
...
 
-                $warning = $this->destinationHandle->query("SHOW WARNINGS");
-
-                if ( count($warning) > 0 ) {
-                    $warnings[$etlTableKey] = array_key_exists($etlTableKey, $warnings) ? array_merge($warnings[$etlTableKey], $warning) : $warning;
+                    try {
+                        // Some values of parameters are complex objects and cannot be! This is due to
+                        // the auto-generated field map.
+                        $insertStatements[$etlTableKey]->execute($parameters);
+                    } catch (PDOException $e) {
```
aaronweeden, line 327: Does the rollback need to happen here, too?

classes/ETL/Ingestor/StructuredFileIngestor.php
eiffel777: It is not. I've changed this in #2003

PR #2003, opened 2025-03-10
Improving transaction error handling in StructuredFileIngestor

eiffel777:
The call to rollback needs to be before the `logAndThrowException` call and it should catch all exceptions not just PDOException so that it can catch exceptions thrown on line 329.

## Tests performed
Tested in docker by adding a `throw new PDOException` in the try catch on line 323 and in the for loop on line 313

## Checklist:

### DIFF
--- a/classes/ETL/Ingestor/StructuredFileIngestor.php
+++ b/classes/ETL/Ingestor/StructuredFileIngestor.php
@@ -346,12 +346,12 @@ class StructuredFileIngestor extends aIngestor implements iAction
             }
 
             $this->destinationHandle->commit();
-        } catch (PDOException $e) {
+        } catch (Exception $e) {
+            $this->destinationHandle->rollback();
             $this->logAndThrowException(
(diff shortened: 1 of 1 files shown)


classes/ETL/Ingestor/StructuredFileIngestor.php
```diff
@@ -346,12 +346,12 @@ function ($k, $v) {
             }
 
             $this->destinationHandle->commit();
-        } catch (PDOException $e) {
+        } catch (Exception $e) {
+            $this->destinationHandle->rollback();
             $this->logAndThrowException(
                 "Error committing transaction. Rolling back transactions.",
```
aaronweeden, line 352:
```suggestion
                "Error committing transaction. Rolling back transaction.",
```
It's just one transaction that gets rolled back, right?

eiffel777: @aaronweeden Yeah, it is.

aaronweeden (approved):

PR #2017, opened 2025-03-26
Update RPM install/upgrade instructions.

aaronweeden:
This PR updates the docs as follows:
* Explain that when upgrading via RPM, the GitHub URL can be directly installed via `dnf`, and if modules are installed, their RPMs need to be installed in the same `dnf install` line.
* Mention to be aware of `.rpmnew` and `.rpmsave` files.
* Make sure the source upgrade instructions link to the module upgrade pages instead of their install pages.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -121,7 +121,7 @@ Note that if you have installed any of the optional modules for Open XDMoD, you
 should also include their new RPM file(s) on the same `dnf install` command
 line below that you use to install the new Open XDMoD RPM file. The upgrade
 guides for each of the optional modules are linked below; these each contain a
-link to the GitHub page for the module release, which has the link for their
+link to the GitHub page for the module release, which has the link to their
 RPM file.
 
 - [Application Kernels](https://appkernels.xdmod.org/{{ page.version }}/ak-upgrade.html)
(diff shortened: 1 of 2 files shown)


docs/upgrade.md
```diff
@@ -115,24 +115,26 @@ You may now continue with the standard upgrade steps below.
...
-### Install the RPM
+If you have installed any of the optional modules for Open XDMoD:
 
-    # dnf install xdmod-{{ page.rpm_version }}.el8.noarch.rpm
+1. Download their RPMs, too; download links and additional notes are available
+   in their upgrade guides:
+    - [Application Kernels](https://appkernels.xdmod.org/{{ page.version }}/ak-upgrade.html)
+    - [Job Performance (SUPReMM)](https://supremm.xdmod.org/{{ page.version }}/supremm-upgrade.html)
+    - [OnDemand](https://ondemand.xdmod.org/{{ page.version }}/upgrade.html)
+1. Add the module RPMs to the end of the command below.
```
eiffel777, line 129: Should this be a 2. at the begging of this line, not a 1?

docs/upgrade.md
```diff
@@ -115,24 +115,26 @@ You may now continue with the standard upgrade steps below.
...
 
-If you have installed any of the optional modules for Open XDMoD, download and
-install their RPMs, too:
-- [Application Kernels](https://appkernels.xdmod.org/{{ page.version }}/ak-install-rpm.html)
-- [Job Performance (SUPReMM)](https://supremm.xdmod.org/{{ page.version }}/supremm-install.html)
-- [OnDemand](https://ondemand.xdmod.org/{{ page.version }}/install.html)
+```
+# dnf install xdmod-{{ page.rpm_version }}.el8.noarch.rpm
```
eiffel777, line 132: Would it be good to have an example `dnf` command with multiple modules?

docs/upgrade.md
jpwhite4: The markdown processor turns `1.` into a `<ol><li></li></ol>`

docs/upgrade.md
jpwhite4: Also dnf can download from https anyway, so you don't actually need to separately download and dnf install the packages - only necessary if you have an air-gapped system.

aaronweeden: Thanks, I have added these.

eiffel777 (approved):

PR #2027, opened 2025-04-08
Playwright updates for xdmod-supremm

ryanrath:
## Motivation and Context
These minor changes were found while getting the playwright tests working for xdmod-supremm.

## Checklist:

### DIFF
--- a/tests/playwright/lib/base.page.ts
+++ b/tests/playwright/lib/base.page.ts
@@ -1,4 +1,5 @@
 import {expect, Locator, Page} from "@playwright/test";
+import config from "../playwright.config";
 
 export class BasePage {
     readonly page: Page;
@@ -10,6 +11,9 @@ export class BasePage {
         this.page = page;
         this.maskSelector = '.ext-el-mask-msg';
         this.mask = page.locator(this.maskSelector);
+        if (baseUrl == undefined || baseUrl.length === 0) {
+            baseUrl = config.use.baseURL
+        }
         this.baseUrl = baseUrl;
     }
 
--- a/tests/playwright/lib/login.page.ts
+++ b/tests/playwright/lib/login.page.ts
@@ -27,7 +27,7 @@ export class LoginPage extends BasePage implements LoginInterface {
 
   readonly sso:boolean;
 
-  constructor(page: Page, baseUrl: string, sso:boolean) {
+  constructor(page: Page, baseUrl: string, sso:boolean=false) {
     super(page, baseUrl);
     this.sso = sso;
     this.logo = page.locator('#logo');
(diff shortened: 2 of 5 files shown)


tests/playwright/lib/base.page.ts
```diff
@@ -10,6 +11,9 @@ export class BasePage {
         this.page = page;
         this.maskSelector = '.ext-el-mask-msg';
         this.mask = page.locator(this.maskSelector);
+        if (baseUrl == undefined || baseUrl.length === 0) {
```
jpwhite4, line 14: The type hinting says that baseUrl is a string - doesn't that mean it can't be undefined? I'm not 110% sure how the typescript type hinting functions

jpwhite4: image

jpwhite4: looks like you could never make baseUrl undefined or null so that check is done by the compiler and you don't need to do it in code.

tests/playwright/lib/metricExplorer.page.ts
```diff
@@ -80,7 +80,6 @@ class MetricExplorer extends BasePage{
         await this.page.click(this.newChart.modalDialog.ok());
         await expect(this.page.locator(this.newChart.modalDialog.box)).toBeHidden();
         await expect(this.page.locator(this.newChart.modalDialog.noDataMessage)).toBeVisible();
-        await expect(this.page.locator(this.mask)).toBeHidden();
```
jpwhite4, line 83: How come you removed this check?

ryanrath: So I'm pretty sure I removed it because I misread it to be `await this.page.locator(this.mask).isHidden();` which doesn't do anything. But it ends up that having it or not having it doesn't seem to affect whether or not tests that call this function pass.

tests/playwright/lib/base.page.ts
ryanrath: Yup, I think I put this in as a precaution while developing. It'll be removed.

tests/playwright/lib/metricExplorer.page.ts
ryanrath: @jpwhite4 After some research I replaced the `toBeHidden` to `toHaveCount(0)` to capture that the mask is actually removed from the dom when it disappears unlike some of our menus which are just made hidden. So it was failing before because `toBeHidden` expects the element to exist within the DOM.

tests/playwright/lib/metricExplorer.page.ts
```diff
@@ -106,7 +106,7 @@ class MetricExplorer extends BasePage{
         await this.endDateLocator.click();
         await this.page.fill(this.endDate, end);
         await this.page.click(this.toolbar.buttonByName('Refresh'));
-        await this.page.locator(this.mask).isHidden();
+        await expect(this.page.locator(this.mask)).toHaveCount(0);
```
ryanrath, line 109: Same reason as above for this change.

tests/playwright/lib/login.page.ts
```diff
@@ -27,7 +27,7 @@ export class LoginPage extends BasePage implements LoginInterface {
 
   readonly sso:boolean;
 
-  constructor(page: Page, baseUrl: string, sso:boolean) {
+  constructor(page: Page, baseUrl: string, sso:boolean=false) {
```
jpwhite4, line 30: How come this changed?

jpwhite4 (approved):

PR #2029, opened 2025-04-22
Updating playwright docker image version

ryanrath:
## Description
Just updating the version number so tests don't break.

This change was made following @aaronweeden finding that it was a playwright version mismatch between what was in `package.json` and the version installed by `npm install -g @playwright/test`. 

## Checklist:

### DIFF
--- a/tests/playwright/Docker/docker-compose.yml
+++ b/tests/playwright/Docker/docker-compose.yml
@@ -16,7 +16,7 @@ services:
         tty: true
         command: sleep infinity
     playwright:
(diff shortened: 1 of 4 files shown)


aestoltm:
Just one question,

Should `package-lock.json` in `tests/playwright' be updated with the version change?

ryanrath:
@aestoltm Yup, a downside to the current way I test locally ( copying the working directory in to the docker container). Just committed the updated file.

aestoltm (dismissed):

aestoltm (approved):
Originally, I thought that `fsevents` was an external package, but it is an optional package used by playwright

PR #2033, opened 2025-04-28
Update Sphinx to latest

jpwhite4:
Keep the user manual builder upto date with latest dependencies.


jpwhite4:
This supercedes some half-baked bot pull requests that tried to update the dependency versions without actually knowing what needed to change.

PR #2039, opened 2025-05-08
Moved 11.5 etl migration actions into migration etl file instead of referencing them

eiffel777:
The `xdmod-xsede` module has a different `ingest_resources.json` file that does not have the same actions so even though the migration is not run for ACCESS XDMoD it still checks this file. When it does the references are either the wrong etl actions or do not exist. To prevent this, I just copied the necessary etl actions into the migration file.

## Tests performed
Tested in docker

## Checklist:

### DIFF
--- a/configuration/etl/etl.d/xdmod-migration-11_0_0-11_5_0.json
+++ b/configuration/etl/etl.d/xdmod-migration-11_0_0-11_5_0.json
@@ -64,28 +64,188 @@
             ]
         },
         {
-            "$ref": "${local_config_dir}/ingest_resources.json#/ingest-resources/0"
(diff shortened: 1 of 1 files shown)


aaronweeden:
The file in `xdmod-xsede` doesn't appear to be used anywhere; I wonder if it would be better to remove it?

eiffel777:
@aaronweeden I would be fine with that. How about we merge this so it doesn't affect our dev systems for now and then we check to make sure it isn't being used anywhere and remove it. After that, this file can get changed back.

aaronweeden:
@eiffel777 sounds good.

PR #2050, opened 2025-06-03
[TRIVIAL] Fix documentation links pointing to .md files

einonm:
Some documentation links are pointing to the source .md files, resulting in 404 errors - fix this by changing the extension to .html to match other documentation links.


jpwhite4:
Hi,

we can't reproduce the 404 problem described in this pull request. The two scenarios are viewing the (rendered) pages on the main documentation site and viewing the pages in github. The Jekyll processor handles converting md links to html links when the documentation is pushed to the main website (see e.g. the link in https://open.xdmod.org/11.0/cloud.html functions as expected). Also if you view the md pages using github the links also work (see e.g. https://github.com/ubccr/xdmod/blob/main/docs/cloud.md)

Please can you add the steps to reproduce to the comments, so we can confirm the bug.

Thanks,

Joe

PR #2028, opened 2025-04-09
Remove Data Analytics Framework token authentication via GET/POST params.

aaronweeden:
## Description

This PR removes the ability to authenticate to the Data Analytics Framework REST API via API token in GET or POST parameters and makes it so the token must be provided via `Authorization` header. The GET/POST param functionality was in place for CentOS 7 to work around the `Authorization` header being eaten by Apache, and it is no longer needed.

These changes mean the current method of counting Data Analytics Framework requests for our reporting KPIs will no longer work (they involved reading the logs with `ident = 'rest.logger.db' or ident = 'controller.log'` and searching for the `Bearer` parameter). To correct this, this PR adds logging with a new `ident = 'daf'`.

This PR also updates the error messages and documentation to use consistent lowercasing of "token" and to move the duplicate error message strings into variables.

This PR also refactors some duplicated code into functions.

There is also a PR for https://github.com/ubccr/xdmod-xsede/pull/570.

## Motivation and Context
Code cleanup and better practice not to have the API token in the request parameters.

## Tests performed

This PR updates the token auth integration tests.

In a Docker container with the changes from this PR, I also made requests via the version of `xdmod-data` from https://github.com/ubccr/xdmod-data/pull/68 and confirmed the logs appeared as expected, e.g.,:

```
+-------+---------------------+-------+----------+-----------------------------------------------------------------------------------------------------------+
| id    | logtime             | ident | priority | message                                                                                                   |
+-------+---------------------+-------+----------+-----------------------------------------------------------------------------------------------------------+
| 14122 | 2025-04-09 19:00:20 | daf   |        6 | {"message":"User 3 requested /controllers/metric_explorer.php with API token using xdmod-data Python v1.1.0.dev09"} |
| 14123 | 2025-04-09 19:00:21 | daf   |        6 | {"message":"User 3 requested /controllers/user_interface.php with API token using xdmod-data Python v1.1.0.dev09"}  |
| 14125 | 2025-04-09 19:00:21 | daf   |        6 | {"message":"User 3 requested /v1/warehouse/export/realms with API token using xdmod-data Python v1.1.0.dev09"}      |
| 14127 | 2025-04-09 19:00:22 | daf   |        6 | {"message":"User 3 requested /v1/warehouse/raw-data with API token using xdmod-data Python v1.1.0.dev09"}           |
+-------+---------------------+-------+----------+-----------------------------------------------------------------------------------------------------------+
```

## Checklist:

- [x] The pull request description is suitable for a Changelog entry
[...]

PR #2054, opened 2025-06-25
Restore Data Analytics Framework token authentication via GET/POST params.

aaronweeden:
## Description

This PR restores the ability to authenticate to the Data Analytics Framework REST API via API token in a `Bearer` parameter of GET or POST requests instead of an `Authorization` header. This functionality had been removed in #2028, but after that PR was merged, it was discovered that the eating of the `Authorization` header described in that PR also happens in Rocky 8, and a proper fix requires an update to the Apache configuration. Because updating the Apache configuration requires the admin to make manual changes, this is deemed a substantial enough change to warrant deferring it to a major release. So, for now, we are rolling back to allowing the `Bearer` parameter to be used when making GET/POST requests.

This PR also refactors the code a bit to make it easier to read and maintain. Specifically:
* Remove the constant `HEADER_KEY` because the string `'Bearer'` is unlikely to change, and it is more readable not to have to look up the value of the constant.
* Move the `authenticateToken` method from `BaseControllerProvider` into the `Tokens` class, renamed as `authenticate` — this relegates all the handling of the token authentication to the `Tokens` service to encapsulate it all in one place.
* Rename the `authenticateToken` method in the `Tokens` class to `authenticateController` since it is meant to be a temporary function until the controller endpoints are refactored into the new REST stack in a future version of XDMoD.
* Rename the `authenticate` method in the `Tokens` class to `authenticateToken` and make it private since now there are no external classes that need to use it.
* Move common exception throwing in the `Tokens` class into a function to simplify the repeated code.

## Tests performed

In addition to making sure the CI tests passed, I also ran various tests on my port on `xdmod-dev` with the changes from this PR, running `curl` with and without a token in a `Bearer` parameter for the various endpoints that use token authentication.

This PR changes the integration test for the API tokens to put the token in the query parameters instead of the header. A future PR that fixes the Apache configuration should also update this to test both the case where the token is in the header and the case where the token is instead in the query parameters.

## Checklist:

### DIFF
--- a/classes/Models/Services/Tokens.php
+++ b/classes/Models/Services/Tokens.php
@@ -4,6 +4,7 @@ namespace Models\Services;
 
 use CCR\Log;
 use Exception;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 use XDUser;
@@ -15,11 +16,6 @@ use XDUser;
 class Tokens
 {
     /**
-     * This is the key that will be used when adding an API Token to a request's headers.
-     */
-    const HEADER_KEY = 'Bearer';
-
-    /**
      * This is the delimiter that's used when returning a newly created API token to the user.
      */
     const DELIMITER = '.';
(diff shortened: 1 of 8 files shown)


classes/Models/Services/Tokens.php
```diff
@@ -4,6 +4,7 @@
 
 use CCR\Log;
 use Exception;
+use Symfony\Component\HttpFoundation\Request;
```
connersaeli: Is this addition necessary? It looks like the Symfony `Request` objects are used but not instantiated here.

connersaeli (approved):

PR #2059, opened 2025-07-03
Add support for date-based URL selection and updated the usageTab playwright test

waheed-saroyia:
**Description**

This pull request introduces functionality to control the date range in the Usage tab directly via URL parameters. Users can now specify a start_date, end_date, or a named duration in the URL to pre-load a report with a specific timeframe.

The implementation involves updating the XDMoD.Module.Usage.js component to parse these new URL parameters upon page load. The component now checks for these parameters after the initial render and uses the existing setValues() method on the DurationToolbar to update the UI, ensuring the date pickers and data reflect the state defined in the URL.

Additionally, this PR includes new Playwright tests in usageTab.spec.ts to validate this new functionality.

**Motivation and Context**

The primary motivation for this change is to enhance user experience by allowing them to bookmark or share direct links to reports that are pre-filtered for a specific date range. This avoids the need for users to manually re-select the dates every time they visit a shared link, improving workflow and collaboration. This addresses the request to make the application's state more accessible via the URL.

**Tests performed**

The following tests were performed to validate the changes:

Manual Testing:

- Verified that loading a report URL with &start_date=YYYY-MM-DD&end_date=YYYY-MM-DD correctly filters the data and updates the date picker UI.

- Verified that loading with only a start_date or end_date correctly sets the respective date and uses a sensible default for the other.

- Verified that loading with a &duration=... parameter (e.g., &duration=Yesterday) correctly selects the preset and updates the dates.

- Confirmed that the page loads with the default date range when no parameters are provided, ensuring no regressions.

Automated Testing (Playwright):

- Added a new test case to usageTab.spec.ts that navigates to a URL with valid start_time and end_time parameters and asserts that the date input fields are updated correctly.

- Added a new test case that provides invalid (non-date string) values to the URL parameters and asserts that the application handles it gracefully and falls back to the default date range.

- Added a new test case that provides a valid duration parameter and asserts that the duration selector button's text is updated to match.


html/gui/js/modules/Usage.js
```diff
@@ -971,6 +971,31 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                 const realm = params.get('realm');
                 const groupBy = params.get('group_by');
                 const statistic = params.get('statistic');
+                const start = params.get('start_time');
+                const end = params.get('end_time');
+                const duration = params.get('duration');
+
+                //parsing the 'start_time' and 'end_time' URL parameters into Date objects
+                const startDate = Date.parseDate(start, 'Y-m-d');
```
jpwhite4, line 978: Please explain the reason for not specifying the third parameter to the parseDate function. 

https://cdn.sencha.com/ext/gpl/3.4.1.1/docs/#!/api/Date-static-method-parseDate

html/gui/js/modules/Usage.js
```diff
@@ -971,6 +971,25 @@ Ext.extend(XDMoD.Module.Usage, XDMoD.PortalModule, {
                 const realm = params.get('realm');
                 const groupBy = params.get('group_by');
                 const statistic = params.get('statistic');
+                const start = params.get('start_time');
+                const end = params.get('end_time');
+                const duration = params.get('duration');
+
+                // Third parameter "strict" defaults to false, allowing date rollover
```
jpwhite4: So why are we not using the option to validate date strings while parsing? Seems like the sort of thing we want.  Does it not work properly or is there another issue with it? Please explain .

jpwhite4 (changes requested):

html/gui/js/modules/Usage.js
jpwhite4:
```suggestion
```

PR #2127, opened 2025-12-03
Simplify chart exporting

aestoltm:
## Description

Chart exporting before would essentially run the same steps that we did in `html/gui/js/PlotlyChartWrapper.js`. The function `XDMoD.utils.createChart()` should have been called before but ran into namespace issues back during the Highcharts conversion.

This is an attempt to properly export the chart. This approach will help avoid bugs by not needing to replicate chart fixups within the export file but instead running the same chart function used in the portal.

This also fixes a bug with the time zone code (present in `html/gui/js/PlotlyChartWrapper.js`) but not in `html/plotly_template.html` which was the motivation for these changes and to move away from this approach of duplicating code (which is not even correctly duplicated)

## Motivation and Context

Original approach was difficult to maintain and easy to introduce export bugs. Now everything runs through the same function (unless Job Viewer plots). The reason job viewer plots are different is because during the Highcharts conversion those were done first and we still needed to support the original Highcharts exporting for the rest of the portal.

Now that everything is in Plotly. There should be another PR to move the Job Viewer chart generation code into `html/gui/js/PlotlyChartWrapper.js` so that there is truly one function call. However, that should be a separate PR because it is much more involved.

## Tests performed

Tested on my dev port

## Checklist:

### DIFF
--- a/html/gui/js/PlotlyChartWrapper.js
+++ b/html/gui/js/PlotlyChartWrapper.js
@@ -21,7 +21,7 @@ XDMoD.utils.createChart = function (chartOptions, extraHandlers) {
         doubleClick: 'reset',
         doubleClickDelay: 500,
         showAxisRangeEntryBoxes: false,
-        staticPlot: chartOptions.isExport ? true : false
+        staticPlot: chartOptions.isExport
     };
     XDMoD.utils.deepExtend(baseChartOptions, chartOptions);
     const isEmpty = (!baseChartOptions.data) || (baseChartOptions.data && baseChartOptions.data.length === 0);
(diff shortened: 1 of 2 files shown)


html/gui/js/PlotlyChartWrapper.js
```diff
@@ -61,6 +62,7 @@ XDMoD.utils.createChart = function (chartOptions, extraHandlers) {
 
         if (baseChartOptions.data[0].type === 'pie') {
             baseChartOptions.layout.pieChart = true;
+            baseChartOptions.layout.margin.t += 30;
```
connersaeli, line 65: What is the significance of adding this 30px margin to the top of pie charts? I assume this change applies to all generated charts and not just exported charts.

aestoltm: It is to help push up the title a little bit because the auto computed margin was to close (especially with data labels). 

There are a few issues with the auto margin that can affect the pie chart title and subtitle (especially with data labels). I want to make a separate PR for 12.0 to address several margin issues I have noticed with pie charts and other plots.

connersaeli (approved):

PR #2128, opened 2025-12-03
Wordwrap long legends

aestoltm:
## Description

Depends on PR #2127 . Will also require PRs on the various XDMoD modules (https://github.com/ubccr/xdmod-xsede/pull/1099 and https://github.com/ubccr/xdmod-appkernels/pull/145).

Add word wrapping to long legend entries. Plotly JS does not support automatic word wrap on legends, therefore, we have to pragmatically determine the word wrap condition based on the chart div width (also including when we resize the browser window).

Before:
before

After:
after

## Motivation and Context

Long legends were unviewable before, especially detrimental for exported charts

## Tests performed

Tested on dev port

## Checklist:


classes/DataWarehouse/Access/Usage.php
```diff
@@ -848,6 +848,7 @@ function ($drillTarget) {
                         ) {
                             $rank = $meDataSeries['legendrank'] / 3;
                             $meDataSeries['name'] = "${rank}. " . $meDataSeries['name'];
+                            $meDataSeries['oname'] = "${rank}. " . $meDataSeries['oname'];
```
jpwhite4, line 851: I don't quite understand the need to add yet another parameter to the chart config that is a copy of an existing parameter. The word wrapping is all done in the javascript, so why do you need a new parameter?

Also (assuming this new parameter is necessary): Is this going to cause breakage with existing saved charts? or with charts saved in the report generator?

PR #2129, opened 2025-12-03
Fix `UserJobEfficiencyComponenet` and other dashboard errors

aestoltm:
## Description

@aaronweeden pointed out an error with the `UserJobEfficiencyComponenet` dashboard component. The issue was we assumed that plots always have an `axis` property when some custom plots do not. While exploring the issue I also found various issues with plots that do not have a subtitle.

## Motivation and Context

 `UserJobEfficiencyComponenet` dashboard component did not work at all. Other plots had resizing errors (if no subtitle)

## Tests performed

Tested on my dev port

## Checklist:


html/gui/js/libraries/PlotlyUtilities.js
```diff
@@ -315,13 +315,13 @@ function isTopLegend(layout) {
  * @return {Object} update - Contains Plotly JS relayout updates and subtitle line count
  */
 function adjustSubtitle(layout, subtitleIndex, legendTopCenter, firstRender) {
-    let prevLineCount = layout.annotations[1].text.match(/<br \/>/g);
+    let prevLineCount = layout.annotations[`${subtitleIndex}`].text.match(/<br \/>/g);
```
jpwhite4, line 318: how come you are converting the subtitleIndex variable to a string to do the array lookup? What is the difference between yyour code and `layout.annotations[subtitleIndex]`

html/gui/js/libraries/PlotlyUtilities.js
```diff
@@ -415,33 +402,49 @@ function relayoutChart(chartDiv, adjHeight, firstRender = false, isExport = fals
...
+            if (subtitleLineCount > 0 && firstRender) {
+                marginTop = subtitleUpdates.chartUpdates['margin.t'];
             }
-            if (topCenter) {
-                if (subtitleUpdates.subtitleLineCount > 0) {
-                    marginTop += subtitleHeight + 5;
-                    update['legend.y'] -= 0.025;
-                    update['margin.t'] += chartDiv._fullLayout.legend._height + subtitleHeight;
+
+            // Observed inconsistency with margin when subtitle is one line long. Unsure of the cause.
```
jpwhite4, line 415: This comment is not particularly useful - what is the inconsistency? how does this relate the code that the comment is next to?

html/gui/js/libraries/PlotlyUtilities.js
```diff
@@ -415,33 +402,49 @@ function relayoutChart(chartDiv, adjHeight, firstRender = false, isExport = fals
...
+            if (lineBreakCount > 0) {
+                if (firstRender) {
```
jpwhite4, line 417: please can you explain the logic here. It seems like on the first render you ignore the subtitle settings. How come?

jpwhite4 (changes requested):

PR #2130, opened 2025-12-03
`RestIngestor` refactor

aestoltm:
## Description

The `RestIngestor` will now extend the `StructuredFileIngestor` and save the JSON results to a tmp directory (e.g. `/www/scratch/rest-ingestor/`) before being ingested through a `DirectoryScanner` data endpoint

## Motivation and Context

Large refactor to `RestIngestor`.

## Tests performed

Tested with CiDER API (Operations API)

## Checklist:


classes/ETL/Ingestor/RestIngestor.php
```diff
@@ -982,4 +982,53 @@ private function applyVerifyDirective($value, \stdClass $directive)
         return true;
 
     }  // verifyTransformDirective()
+
+     /* ------------------------------------------------------------------------------------------
+      * Parses PHP obj/array and retrieves desired field. Supports a wildcard, "?", for the last item in an array.
```
jpwhite4: What is the use case for adding support for the 'last item in array' wildcard?

aestoltm: Originally was using this to get the latest conversion factors from the cider api (last element). However, the cider pipeline is going to create a history of the conversion factors so I do not actually use it anymore. I can't think of any other cases so we probably can remove it.

jpwhite4: I think it is better to not add it for now. If we need a capability in future then should be easy to add it in based on this work.

PR #2189, opened 2026-04-14
Simplify metric explorer getFormat

ryanrath:
## Description
Refactored `getFormat` to the much simpler new function `validateFormat` and removed `getFormat`. 

## Motivation and Context
These changes are to address the points raised in: https://github.com/ubccr/xdmod-xsede/issues/1154

## Tests performed
All automated tests.

## Checklist:

### DIFF
--- a/classes/DataWarehouse/Access/MetricExplorer.php
+++ b/classes/DataWarehouse/Access/MetricExplorer.php
@@ -22,10 +22,7 @@ class MetricExplorer extends Common
 {
     public function get_data($user)
     {
-        $requestedFormat = 'csv';
-        if (isset($this->request['format'])) {
-            $requestedFormat = strtolower($this->request['format']);
-        }
+        $requestedFormat = $this->request['format'] ?? null;
 
         $format = \DataWarehouse\ExportBuilder::validateFormat(
             $requestedFormat,
--- a/classes/DataWarehouse/Access/MetricExplorer.php
+++ b/classes/DataWarehouse/Access/MetricExplorer.php
@@ -22,13 +22,13 @@ class MetricExplorer extends Common
 {
     public function get_data($user)
     {
-        if(isset($this->request['config'])) {
-            $config = json_decode($this->request['config'], true);
-            $this->request = array_merge($config, $this->request);
+        $requestedFormat = 'csv';
+        if (isset($this->request['format'])) {
+            $requestedFormat = strtolower($this->request['format']);
         }
 
-        $format = \DataWarehouse\ExportBuilder::getFormat(
-            $this->request,
+        $format = \DataWarehouse\ExportBuilder::validateFormat(
+            $requestedFormat,
             'png',
             array(
                 'svg',
--- a/classes/DataWarehouse/ExportBuilder.php
+++ b/classes/DataWarehouse/ExportBuilder.php
@@ -620,4 +620,28 @@ class ExportBuilder
 
         return $name;
     }
+
+    /** Validates that the format requested by the user is located in the set of formats that are supported and either
+     * all formats are allowed ( signified by there being no $allowedFormats ) or the requested format was found in the
+     * set of allowed formats. If valid the requested format is returned. If no requested format is provided then the
+     * default value will be returned.
+     *
+     * @param ?string $requestedFormat
+     * @param string $default
+     * @param array $allowedFormats
+     * @return string
+     */
+    public static function validateFormat(?string $requestedFormat, string $default = 'jsonstore', array $allowedFormats = []): string
+    {
+        if (!isset($requestedFormat)) {
+            return $default;
+        }
+        $requestedFormat = strtolower($requestedFormat);
+        $formatSupported = isset(self::$supported_formats[$requestedFormat]);
+        $noFormatSubset = count($allowedFormats) === 0;
+        $requestedFormatInSubset = in_array($requestedFormat, $allowedFormats);
+
+
+        return $formatSupported && ($noFormatSubset || $requestedFormatInSubset) ? $requestedFormat : $default;
+    }
 }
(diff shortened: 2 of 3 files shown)


html/controllers/metric_explorer/get_rawdata.php
```diff
@@ -14,18 +14,12 @@
 
 $start = microtime(true);
 try {
-    if (isset($_REQUEST['config'])) {
-        $config = json_decode($_REQUEST['config'], true);
-        $_REQUEST = array_merge($config, $_REQUEST);
+    $requestedFormat = null;
+    if (isset($request['format'])) {
+        $requestedFormat = strtolower($request['format']);
```
aestoltm: Is `$request[]` supposed to be `$REQUEST` here? I know the symfony code doesn't deal with the super global directly anymore but I think it is supposed to be that here.

classes/DataWarehouse/Access/MetricExplorer.php
```diff
@@ -22,13 +22,13 @@ class MetricExplorer extends Common
 {
     public function get_data($user)
     {
-        if(isset($this->request['config'])) {
-            $config = json_decode($this->request['config'], true);
-            $this->request = array_merge($config, $this->request);
+        $requestedFormat = 'csv';
+        if (isset($this->request['format'])) {
+            $requestedFormat = strtolower($this->request['format']);
```
jpwhite4: What is the purpose of the strtolower call here? There is also a strtolower call in the validateFormat function.

classes/DataWarehouse/ExportBuilder.php
```diff
@@ -620,4 +585,28 @@ private static function formatElement($name)
...
+     * all formats are allowed ( signified by there being no $allowedFormats ) or the requested format was found in the
+     * set of allowed formats. If valid the requested format is returned. If no requested format is provided then the
+     * default value will be returned.
+     *
+     * @param ?string $requestedFormat
+     * @param string $default
+     * @param array $allowedFormats
+     * @return string
+     */
+    public static function validateFormat(?string $requestedFormat, string $default = 'jsonstore', array $allowedFormats = []): string
```
jpwhite4: Why is requestedFormat allowed to be not set? What circumstances would cause it to be not set and why do we want to support this?

html/controllers/metric_explorer/get_rawdata.php
jpwhite4: Same question about the strtolower call here: why is it being done when it is also called in the validateFormat function

classes/DataWarehouse/ExportBuilder.php
jpwhite4: What is the design reason for having default values for the `$default` and the `$allowedFormats` parameters? Where is this function called when these defaults are used?

jpwhite4 (changes requested):

html/controllers/metric_explorer/get_rawdata.php
ryanrath: it is and I'm not sure why it was changed, I'll change it back.

classes/DataWarehouse/ExportBuilder.php
ryanrath: Really it was just because the "ancestor" function `getFormat` had a `$default` and a `$formats_subset` ( which I renamed to `$allowedFormats` so that it made more sense. ).  It's never called without a `$default` or a `$allowedFormats` argument. I'll go ahead and remove those arguments.

ryanrath: I believe it was something I ran into while testing, but honestly it should never be null. I've tested removing the nullable and all tests passed so I'll be removing it

classes/DataWarehouse/Access/MetricExplorer.php
ryanrath: Just a hold over from the original code, this and the other strtolower will be removed.

ryanrath:
@jpwhite4 @aestoltm I've pushed changes based on your review comments, could you take another look when you have a sec.

classes/DataWarehouse/Access/MetricExplorer.php
```diff
@@ -22,13 +22,10 @@ class MetricExplorer extends Common
 {
     public function get_data($user)
     {
-        if(isset($this->request['config'])) {
-            $config = json_decode($this->request['config'], true);
-            $this->request = array_merge($config, $this->request);
-        }
+        $requestedFormat = $this->request['format'] ?? null;
```
jpwhite4, line 25: How do you know that the json_decode is not necessary any more?

classes/DataWarehouse/ExportBuilder.php
```diff
@@ -620,4 +585,28 @@ private static function formatElement($name)
...
+     * default value will be returned.
+     *
+     * @param string $requestedFormat
+     * @param string $default
+     * @param array $allowedFormats
+     * @return string
+     */
+    public static function validateFormat(string $requestedFormat, string $default, array $allowedFormats): string
+    {
+        if (!isset($requestedFormat)) {
```
jpwhite4, line 601: Is it possible for isset() to return false here?

html/controllers/metric_explorer/get_rawdata.php
```diff
@@ -14,18 +14,10 @@
...
-    $format = \DataWarehouse\ExportBuilder::getFormat(
-        $_REQUEST,
-        'jsonstore',
-        array(
-            'jsonstore'
-        )
-    );
+    $requestedFormat = $_REQUEST['format'] ?? null;
+
+    $format = \DataWarehouse\ExportBuilder::validateFormat($requestedFormat, 'jsonstore', ['jsonstore']);
```
jpwhite4, line 20: The definition for the new validateFormat function requires a `string`, but this code could pass in a `null` which will give a TypeError exception.

jpwhite4 (changes requested):

PR #2194, opened 2026-04-15
Update to escaping function

ryanrath:
## Description
Updating password_reset.php to use the correct escaping function. 

## Motivation and Context
Correct escaping == good.

## Tests performed
Manual verification of newly added test was performed by: 
- Reverting password_reset.php to original version and running the new test, confirm that it fails.
- Updating password_reset.php to this PR's version and running the new test, confirm that it passes.
All other automated tests pass.

## Checklist:

### DIFF
--- a/tests/integration/lib/Controllers/UserAdminTest.php
+++ b/tests/integration/lib/Controllers/UserAdminTest.php
@@ -880,13 +880,10 @@ class UserAdminTest extends BaseUserAdminTest
         // Next, snag the original first name so we can revert the changes after the test is done.
         $originalFirstName = $this->getPropertyFromUserProfile($this->helper, 'first_name');
         try {
-            $adminHelper->authenticateDashboard('mgr');
+            // Update cd's first_name
             $newFirstName = '<script>alert("boo!")</script>';
             $updateResponse = $this->helper->patch('rest/v1/users/current', null, ['first_name' => $newFirstName]);
 
-            if ($updateResponse[1]['http_code'] !== 200) {
-                echo var_export($updateResponse, true). "\n";
-            }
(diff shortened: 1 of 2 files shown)


tests/integration/lib/Controllers/UserAdminTest.php
```diff
@@ -866,6 +866,99 @@ protected function getUserVisits(array $options)
...
+        // We need an unauthenticated helper to simulate retrieving of the password reset page.
+        $publicHelper = new XdmodTestHelper();
+
+        // First, login as center director
+        $this->helper->authenticate('cd');
+
+        // Next, snag the original first name so we can revert the changes after the test is done.
+        $originalFirstName = $this->getPropertyFromUserProfile($this->helper, 'first_name');
+        try {
+            $adminHelper->authenticateDashboard('mgr');
```
jpwhite4: I'd move this line down to line 900. to make it more obvious that the adminHelper needs to be authenticated to do the password reset (but does not need to be there at all for the name change).

tests/integration/lib/Controllers/UserAdminTest.php
```diff
@@ -866,6 +866,99 @@ protected function getUserVisits(array $options)
...
+            $newFirstName = '<script>alert("boo!")</script>';
+            $updateResponse = $this->helper->patch('rest/v1/users/current', null, ['first_name' => $newFirstName]);
+
+            if ($updateResponse[1]['http_code'] !== 200) {
+                echo var_export($updateResponse, true). "\n";
+            }
```
jpwhite4: This looks like debug code that should not be here.

jpwhite4 (approved):

PR #2200, opened 2026-04-20
XDUser for Symfony Migration

connersaeli:
## Description

1. `getUserByID()`
* Remove redundant `initial` parameter for `array_reduce()`
2. `arrayToString()`
* I think this method should just be removed. The only location in the entire codebase it is used is the `XDUser::saveUser()` method for exception formatting. There is also no semantic connection to the underlying `XDUser` class. The comment states that it outputs a "meaningful string representation of [the input] array", but it is not clear what a "meaningful representation" means. I've instead moved this to where it is actually used in `saveUser()`. If it gets used elsewhere in the future, then the method can be refactored at that time.
3. `getRoles()`
* Refactor multiple `return` statements into a single `return`
* Replace `initial` parameter with initialized `$roles` array
* Refactor individual `if`'s into an `if else` block
4. `getPersonID()`
* Remove outdated comment
* Refactor redundant `\xd_security\` reference for `SessionSingleton` as it is already imported
* Refactor multiple `return` statements into a single line
5. `hash()`
* Refactor multiple `return`'s into one 

## Motivation and Context

https://github.com/ubccr/xdmod-xsede/issues/1183
## Tests performed

Tested with console and web portal on `symfony-dev`.

## Checklist:


classes/XDUser.php
```diff
@@ -625,7 +625,7 @@ public static function getUserByID($uid, &$targetInstance = NULL)
             $acl = new Acl($item);
             $carry [$acl->getName()] = $acl;
             return $carry;
-        }, array());
+        });
```
ryanrath: Per the `array_reduce` documentation, the last param `initial` is returned if the array being reduced is empty, which in the original case was an empty array, but is now null. This should probably never occur, but I don't know if we can definitively say that it can never occur. Combine this with the more restrictive typing of PHP 8.2+ and we'll get a runtime TypeError when `setAcls` is called w/ null. Here's the results of a quick test I just wrote for that case: 
```shell
Failed asserting that exception of type "TypeError" matches expected exception "Exception". Message was: "XDUser::setAcls(): Argument #1 ($acls) must be of type array, null given, called in /root/xdmod/tests/component/lib/XDUserTest.php on line 280" at
/root/xdmod/classes/XDUser.php:2137
/root/xdmod/tests/component/lib/XDUserTest.php:280
```

I'd personally feel better if we kept this `array()` in, at least for the time being.

ryanrath (changes requested):
Just the one change is all, the rest of it looks good.

classes/XDUser.php
connersaeli: Good catch. Updated here https://github.com/ubccr/xdmod/pull/2200/commits/a7aa461790f8d26a0c15b01c420c3c20252f24da

ryanrath (approved):

PR #2204, opened 2026-04-24
Add backticks to filter lists temp table

eiffel777:
This PR adds backticks to the temporary table name. The temporary table used when appending to the filters lists uses `uniqid` to help create a unique table name but `uniqid` sometimes returns a number with a decimal in it which is not valid for a table name without backticks. 

Also, use `$filterListBuilder->setLogger($logger)` to set the logger for the filter lists in `xdmod-build-filter-lists` as setting the logger in the constructor does not work.

## Tests performed
Tested in docker and xdmod-dev

## Checklist:

### DIFF
--- a/bin/xdmod-build-filter-lists
+++ b/bin/xdmod-build-filter-lists
@@ -89,8 +89,7 @@ build($realms, $logger, $appendToList);
 
 function build($realms, $logger, $appendToList){
     try {
-        $filterListBuilder = new FilterListBuilder();
-        $filterListBuilder->setLogger($logger);
+        $filterListBuilder = new FilterListBuilder($logger);        
         if (count($realms) > 0){
             foreach($realms as $realm){
                 if(!empty($realm)){
--- a/bin/xdmod-build-filter-lists
+++ b/bin/xdmod-build-filter-lists
@@ -89,7 +89,7 @@ build($realms, $logger, $appendToList);
 
 function build($realms, $logger, $appendToList){
     try {
-        $filterListBuilder = new FilterListBuilder($logger);        
+        $filterListBuilder = new FilterListBuilder($logger);
         if (count($realms) > 0){
             foreach($realms as $realm){
                 if(!empty($realm)){
--- a/bin/xdmod-build-filter-lists
+++ b/bin/xdmod-build-filter-lists
@@ -89,7 +89,8 @@ build($realms, $logger, $appendToList);
 
 function build($realms, $logger, $appendToList){
     try {
-        $filterListBuilder = new FilterListBuilder($logger);
+        $filterListBuilder = new FilterListBuilder();
+        $filterListBuilder->setLogger($logger);
         if (count($realms) > 0){
             foreach($realms as $realm){
                 if(!empty($realm)){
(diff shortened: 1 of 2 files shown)


bin/xdmod-build-filter-lists
```diff
@@ -89,7 +89,8 @@ build($realms, $logger, $appendToList);
 
 function build($realms, $logger, $appendToList){
     try {
-        $filterListBuilder = new FilterListBuilder($logger);
+        $filterListBuilder = new FilterListBuilder();
```
jpwhite4: What is the API supposed to be for FilterListBuilder?  if it is not supposed to accept a looger as the first argument then you'll must change all the places where it is called with a logger as the first arg not just this line of code.
```
grep "new FilterListBuilder" -R .
./bin/xdmod-build-filter-lists:        $filterListBuilder = new FilterListBuilder($logger);
./classes/OpenXdmod/DataWarehouseInitializer.php:        $filterListBuilder = new FilterListBuilder();
./classes/OpenXdmod/DataWarehouseInitializer.php:        $filterListBuilder = new FilterListBuilder();
./classes/OpenXdmod/DataWarehouseInitializer.php:        $filterListBuilder = new FilterListBuilder();
./classes/OpenXdmod/DataWarehouseInitializer.php:        $filterListBuilder = new FilterListBuilder();
./classes/DB/Aggregator.php:        $filterListBuilder = new FilterListBuilder($this->logger);
```

if it is supposed to have the logger as the first optional arg then the code should be changed to support that

classes/DB/FilterListBuilder.php
```diff
@@ -144,7 +144,7 @@ public function buildRealmLists($realmName, $appendToList = false)
             $lastModified = $journalHelper->getLastModified();
             $this->setLastModifiedStartDate($lastModified);
             $tempTableSql = "CREATE TEMPORARY TABLE $this->filterTemporaryTable AS SELECT * FROM $schema.$tableName where last_modified >= '$this->lastModifiedStartDate'";
-            $this->logger->debug("Creating temporary table: $tempTableSql");
+            $this->logger->debug("Creating temporary table: $tempTableSql with sql: $tempTableSql");
```
jpwhite4: Not sure what the purpose of this change is. Did you mean to type something else? Such as 
```
Creating temporary table: {$this->filterTemporaryTable} with sql: {$tempTableSql}"
```

jpwhite4 (changes requested):

bin/xdmod-build-filter-lists
jpwhite4: Given that it inherits from Loggable and the first argument to the logable class looks like the following: 
```
  public function __construct(LoggerInterface $logger = null)
    {
        $this->setLogger($logger);
    }
```

it probably make sense to fix the issue in the FilterListBuilder constructor instead

classes/DB/FilterListBuilder.php
eiffel777: That's just some debug code that wasn't needed that got left in when I commit it. It's gone now.

bin/xdmod-build-filter-lists
eiffel777: I updated the constructor to match the Loggable class.

jpwhite4 (approved):

PR #2222, opened 2026-05-14
Update maintenance page.

jpwhite4:
Adds alt tag to image and some html5 metadata that was previously missing. Also no need to run the php interpreter for this one.


html/maintenance/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+
+   <head>
+      <meta charset="UTF-8">
+      <meta http-equiv="refresh" content="10">
+      <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
aaronweeden, line 7:
```suggestion
      <meta name="viewport" content="width=device-width, initial-scale=1.0">
      <meta name="robots" content="noindex">
```
So the page isn't indexed while it's up.

html/maintenance/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+
+   <head>
+      <meta charset="UTF-8">
+      <meta http-equiv="refresh" content="10">
```
aaronweeden, line 6: We may want to avoid this. https://www.boia.org/blog/accessibility-tips-dont-use-meta-refresh-with-time-limits

aaronweeden (changes requested):

html/maintenance/index.html
jpwhite4: I'm open to suggestions as to what to do instead of this. Any suggestion?

aaronweeden (approved):

html/maintenance/index.html
aaronweeden: Don't have a good idea here yet. Can be fixed later.

PR #2226, opened 2026-05-18
Symfony Migration for Tokens

connersaeli:
## Description

These changes redo many of the changes made in https://github.com/ubccr/xdmod/pull/2054/ and the JupyterHub updates that were not merged into `symfony_migration`. Some relevant changes:

- Removed duplicate code from `authenticateToken()` in `Tokens` and `BaseController` that both extracted the raw token from the header.
- Removed a function that was commented out
- Removed `HEADER_KEY` and `BEARER` from `Tokens` (removed in https://github.com/ubccr/xdmod/pull/2054/)
- Simplified logic checks for tokens. For example `$request->headers->get()` returns null if it is not present. There's no need here to check if it's null then `get()` only if it is present.
- Removed `getTokenFromHeader()` method because it is only used in one place now. This method is an artifact of having the different Silex/html controllers and needing this code in two different places. This method can just be replaced with a ternary operator.

## Motivation and Context

https://github.com/ubccr/xdmod-xsede/issues/1171

## Tests performed

Tested on my port of `symfony-dev`

## Checklist:


src/Security/Helpers/Tokens.php
```diff
@@ -45,113 +34,30 @@ public function __construct(LoggerInterface $logger)
...
-        }
+        $authHeader = $request->headers->get('Authorization', '');
+        $rawToken = str_starts_with($authHeader, 'Bearer ') ? substr($authHeader, strlen('Bearer ')) : $rawToken;
+
         // If the token is not in the header, then fall back to extracting from
         // the GET/POST params.
-        if (empty($token)) {
-            $token = $request->get('Bearer');
-        }
+        $rawToken = $request->get('Bearer');
```
ryanrath: This line seems to always overwrite a potential value found in the header, is that expected?

ryanrath (changes requested):
Just the one question really.

src/Controller/InternalDashboard/InternalDashboardController.php
```diff
@@ -49,7 +49,7 @@ public function dashboardLogout(Security $security): Response
     #[Route('/internal_dashboard')]
     public function index(Request $request): Response
     {
-        $user = $this->getXDUser($request->getSession());
+        $user = $this->authorize($request);
```
jpwhite4: needs 'mgr' restriction.

src/Controller/InternalDashboard/SABUserController.php
```diff
@@ -48,7 +48,7 @@ private function enumTgUsers(Request $request): Response
         $query = $this->getStringParam($request, 'query');
         $userManagement = $this->getStringParam($request, 'userManagement');
 
-        $user = $this->getXDUser($request->getSession());
+        $user = $this->authorize($request);
```
jpwhite4: See https://github.com/ubccr/xdmod/pull/2232/changes/a592040cc78dc9d5a6e6c173fbde49915fc9ef64#diff-8ef047114134ad8e987be25b88f3c2cff7f27ef57c59536d2a2691948bac1ec1

src/Controller/BaseController.php
```diff
@@ -128,31 +129,37 @@ protected function getUserFromRequest(Request $request)
...
-                if (!$session->has('public_session_token')) {
-                    $session->set('public_session_token', 'public-' . microtime(true) . '-' . uniqid());
-                }
-                $xdUser = XDUser::getPublicUser();
-            }
+        $session = $request->getSession();
+        // Check for user info in existing session
+        if ($session->has('xdUser')) {
+            $xdUser = XDUser::getUserByID($session->get('xdUser'));
+        } elseif ($session->has('xdmod_token')) {
```
jpwhite4: I _think_ we want to not support being able to authenticate with the xdmod_token or xdUser cookies in 11.5

PR #2235, opened 2026-05-27
Allow local login via specific referer hash

aestoltm:
## Description

Allow local login via `#tutorial` referer hash

## Motivation and Context

Easy local login capabilities for upcoming PEARC26 tutorial

## Tests performed

Tested on xdmod-dev

## Checklist:


html/gui/js/CCR.js
jpwhite4, line 1417: I think it may be a better user-experience if the text changes to "Sign in with your tutorial account" when in tutorial mode. That way it will be explicit that this path is intended to support the tutorial users.

aestoltm:
Closed in favor of #2238

PR #2237, opened 2026-05-28
Fix migrations for 10.5.1 -> 11.0.0 -> 11.0.1.

aaronweeden:
## Description

This PR does two main things:
1. https://github.com/ubccr/xdmod/pull/2237/commits/43822256143c21025caf5ebba3ff1891d83dd1a5: Add CircleCI jobs for testing fresh install and upgrade with only the Jobs and Resource Specifications realms enabled using the new images created in https://tools-int-01.ccr.xdmod.org/git/UBCCR/ccr-private-xdmod/pulls/411. A bit of test refactoring is done in the process.
1. https://github.com/ubccr/xdmod/pull/2237/commits/8662fc2b8d67dff73f477735c8f0b0edadf86deb: Fix a bug (exposed by the new CircleCI jobs) in which the upgrades from 10.5.1 to 11.0.0 and from 11.0.0 to 11.0.1 fail if the Cloud realm is not enabled.

## Motivation and Context

Previously in #1986 I had mistakenly renamed some ETL migration pipelines such that they automatically ran during the upgrade rather than only conditionally running if the `modw_cloud.event` table existed; in that PR I had mistakenly assumed the names had typos and were intended to run automatically. As a consequence of this, if someone tries to upgrade along the path 10.5.1 -> 11.0.0 -> 11.0.1 without ever having ingested cloud data, it fails with the error message `Schema 'modw_cloud' does not exist`. This PR fixes this by renaming the pipelines back to names that do not match the filter used by `Etlv2Migration` (and which are not similar to the filter to reduce confusion).

## Tests performed

I updated the CircleCI tests and confirmed that they succeeded and the correct number of tests ran.

## Checklist:


tests/ci/bootstrap.sh
```diff
@@ -191,6 +191,8 @@ then
     jq . /etc/xdmod/resource_specs2.json > /etc/xdmod/resource_specs.json
     rm -f /etc/xdmod/resource_specs2.json
 
+    xdmod-ingestor
```
jpwhite4: Elsewhere in this file we run `xdmod-ingestor` as the xdmod user via `sudo -u xdmod xdmod-ingestor`. This will run as root - is this intended if so why?

Also is it really necessary to run xdmod-ingestor so many times? Why this extra run of it now?

jpwhite4 (changes requested):

tests/ci/bootstrap.sh
aaronweeden: Running as root was not intentional; I have fixed this.

The changes to `resources.json` and `resource_specs.json` above this line are testing [these instructions](https://open.xdmod.org/11.0/upgrade.html#resourcesjson-and-resource_specsjson) which specify to run `xdmod-ingestor` after making changes to the files. Not running `xdmod-ingestor` here makes the regression tests fail because the resource allocation types aren't ingested yet. It isn't necessary to run `xdmod-ingestor` again if the Storage realm is being tested because the `--datatype storage` option makes it run the `staging-ingest-common` pipeline that includes ingesting the `resources.json` file. Since the call I'm adding to `xdmod-ingestor` only needs to happen if Storage isn't being tested, I've moved it to the `else` below.

tests/component/lib/Query/AggregateQueryTest.php
```diff
@@ -488,6 +489,9 @@ public function testAddWhereJoin()
 
     public function testGetFormula()
     {
+        if (!in_array(strtolower('cloud'), Utilities::getRealmsToTest())) {
+            $this->markTestSkipped('Cloud realm is not being tested.');
+        }
```
eiffel777: Is seems it might be better to pick a query from the jobs realm that has a formula rather than skip this if it is just the jobs realm being tested. I don't think this is a test that is supposed to specifically test the cloud realm, rather a cloud realm statistic was used just because it had a formula for the statistic.

eiffel777 (changes requested):

tests/component/lib/Query/AggregateQueryTest.php
aaronweeden: Good idea, updated.

eiffel777 (dismissed):

PR #2254, opened 2026-06-30
(re)Moving bin/console into xdmod-admin

ryanrath:
## Description
This commit removes Symfony's `bin/console` script and moves that functionality into our `xdmod-admin` script.

## Motivation and Context
We don't want to "pollute" a systems `$PATH` with a non-intuitive executable that's meant for use with XDMoD. So instead this PR adds a new option to `xdmod-admin` `-c, --console` that provides the same functionality that was in previously in `console`. 

## Tests performed
All automated tests were run / passed.

## Checklist:


src/Helper/SymfonyCommandHelper.php
```diff
@@ -28,12 +28,13 @@ class SymfonyCommandHelper
...
      * @throws \RuntimeException if a non-zero exit code is returned by the Symfony Command.
      */
-    public static function executeCommand(string $command, array $options = [], string $env = 'prod', bool $debug = false): void
+    public static function executeCommand(string $command, array $options = [], string $env = 'prod', bool $debug = false): array
     {
         list($statusCode, $output) = self::doExecuteCommand($command, $options, $env, $debug);
         if ($statusCode !== 0) {
             throw new \RuntimeException("Error Running Symfony Command $command\n$output");
         }
+        return [$statusCode, $output];
```
jpwhite4: The only way you can get to this code is if `$statusCode === 0`. So there is no point in returning it as it give no useful information

bin/xdmod-admin
```diff
@@ -229,15 +235,23 @@ function main()
...
                 default:
-                    $logger->crit("Cannot perform '$action' on '$entity'");
+                    $logger->critical("Cannot perform '$action' on '$entity'");
                     exit(1);
                     break;
             }
             break;
+        case 'console':
+            try {
+                list($statusCode, $output) = \CCR\Helper\SymfonyCommandHelper::executeCommand($params);
```
jpwhite4: Isn't `$params` just the next token? how will we support symfony commands that have multiple arguments? Or are they all single argument commands?

jpwhite4 (changes requested):

bin/xdmod-admin
ryanrath: *Isn't $params just the next token?* 
It is.

*how will we support symfony commands that have multiple arguments?* 
Currently the way to do that is enclose the command in quotes so that php sees it as a single argument. Which I will admit may be less than ideal from a user who is not us point of view. But I also think that 99.9% of the time an open xdmod user should never need to run a raw symfony command and if there are use cases that are more common, those should be added to xdmod-admin as new commands.

*Or are they all single argument commands?*
They are not.

src/Helper/SymfonyCommandHelper.php
ryanrath: agreed, I'll just return the output.

ryanrath: @jpwhite4 This change has been made.

bin/xdmod-admin
jpwhite4: Have you tested this? by code inspection the executeCommand function needs a string with the symfony command and an array of options for the rest of the options. But you are not calling it in this way and I don't see where there exists code that splits on the space in order for your suggestion to work.

bin/xdmod-admin
```diff
@@ -39,6 +39,12 @@ function main()
 {
     global $argv, $logger, $force;
 
+    # Check if we should run in xdmod-admin or symfony mode
+    if ($argv[1] === '--console') {
+        $command = implode(' ', array_slice($argv, 2));
```
jpwhite4: This is going to throw away information about the command line arguments.

For example: the input command line of
```
--file "filename with spaces" --print
```
would be mangled into the incorrect
```
--file filename with spaces --print
```

I suggest you just pass in the (already parsed and processed) slice of argv `array_slice($argv, 2)` then to construct the input via the Symfony `ArgvInput` class.

jpwhite4 (changes requested):

bin/xdmod-admin
```diff
@@ -39,6 +39,11 @@ function main()
 {
     global $argv, $logger, $force;
 
+    # Check if we should run in xdmod-admin or symfony mode
+    if ($argv[1] === '--console') {
```
jpwhite4:
```suggestion
    if ($argc > 1 && $argv[1] === '--console') {
```

(and add `$argc` to the list of globals)

tests/unit/lib/Helper/SymfonyCommandHelperTest.php
```diff
@@ -0,0 +1,57 @@
...
+     * Provides the test cases for `testCommandOutput`.
+     *
+     * @return array[]
+     */
+    public function provideCommands(): array
+    {
+        return [
+            ['cache:clear', ['was successfully cleared']],
+            ['dotenv:dump', ['Successfully dumped .env files in .env.local.php']],
+            ['router:match /warehouse/dimensions', ['ccr_warehouse_getdimensions', ' CCR\Controller\WarehouseController::getDimensions()']]
```
jpwhite4: please can you add a test for the ArrayInput case.

ryanrath: @jpwhite4 Absolutely! I've added a couple of tests for ArgvInput and then $argc to globals and the `if` statement in xdmod-admin

jpwhite4 (approved):

PR #2266, opened 2026-07-20
Added all columns into the group by

kylershih0614:
## Description

I changed the script in classes/Realm/Groupby.php to put all of the columns that were in the select statement into the group by. Also wrapped the division of numbers in the configuration/datawarehouse.d/ref files in nullif statements.

Edited errors under strict mode startup like start-date and end date, as well as using CURRENT_TIMESTAMP  as a placeholder. For end_dates and end times, 2037-12-31 was used to make sure a proper far future end date was set.

## Motivation and Context

This change fixes some test cases from failing under sql strict mode. This is not a complete fix of the strict mode errors.

These changes allow for the xdmod setup script to run without issue under strict mode.

## Tests performed

Manual testing with the xdmod website, as well as passing the entire test suite without strict mode.

## Checklist:


configuration/etl/etl_tables.d/xdb/users.json
```diff
@@ -91,7 +91,7 @@
                 "name": "token_expiration",
                 "type": "timestamp",
                 "nullable": false,
-                "default": 0
+                "default": "CURRENT_TIMESTAMP"
```
eiffel777: How/Where in the code is data inserted into this column? Is there always a value provided or will the default value be set? Since this is an expiration date, if the value of the column is the timestamp of when a row is inserted my guess is that `CURRENT_TIMESTAMP` would not be the correct value as a default since the expiration date would be the same time the row is inserted.

configuration/etl/etl_tables.d/acls/user_tokens.json
```diff
@@ -23,12 +23,14 @@
...
-                "nullable": false
+                "nullable": false,
+                "default": "CURRENT_TIMESTAMP"
             },
             {
                 "name": "expires_on",
                 "type": "timestamp",
-                "nullable": false
+                "nullable": false,
+                "default": "CURRENT_TIMESTAMP"
```
eiffel777: How/Where in the code is data inserted into this column? Is there always a value provided or will the default value be set? Since this is an expiration date, if the value of the column is the timestamp of when a row is inserted my guess is that `CURRENT_TIMESTAMP` would not be the correct value as a default since the expiration date would be the same time the row is inserted.

eiffel777 (approved):

PR #2268, opened 2026-07-20
Fix bug with `UserJobEfficiencyComponent` charts

aestoltm:
## Description

Pulling necessary changes from #2129 so that user job efficiency charts load.

## Motivation and Context

There was a missing check on axis existence, which is needed for pie charts since they do not have axis properties.

## Tests performed

Tested on xdmod-dev

## Checklist:


connersaeli:
I'm seeing a couple things that look different from before. It looks like the pie charts are full instead of the semicircles before. Because they're full circles now, the text that should be below the semicircle is now overlapping with the bottom part of the circle. If the semicircle from before can't be preserved, then I think the text should at least be placed elsewhere to not overlap with the chart.
image

jpwhite4 (approved):

jpwhite4:
Should this be a merge into xdmod11.0 instead of main?

PR #2277, opened 2026-07-27
Symfony: Consolidate route role related authorization mechanisms

ryanrath:
## Description
- Added a new Symfony Package `symfony/expression-language`. This is the mechanism that allows us to specify the expression that will drive the `MgrRequired` Attribute functionality ( `"mgr" in role_names` ).
- Moved `AdminController`, and `PersonController` to `src/Controller/InternalDashboard` as the routes provided by these controllers are only called from within the Internal Dashboard. 
- For all routes in the `InternalDashboard` namespace, the `#[MgrRequired]` Attribute was added ( with the exception of `SummaryController::summary3` as this route is meant to be called by an unauthenticated user. 
- Removed all instances of ` $this->denyAccessUnlessGranted('IS_AUTHENTICATED_FULLY');` and `$this->authorize($request, ['mgr']);` within Internal Dashboard controllers as this functionality is now provided by the `MgrRequired` Attribute.
- Where the user returned by `$this->authorize($request, ['mgr']);` was actually used by a function, it was replaced with `$user = $this->getXDUser($requestion->getSession());` as this is the same mechanism that `$this->authorize()` uses.
- Some minor updates were made to the `RouteBasedExceptionListener` to format return values in the way that our tests were expecting. Where appropriate I added comments detailing which routes required which changes. 

## Motivation and Context
We've already moved to using Symfony for our authentication and these changes bring us closer to using them for Authorization. 

## Tests performed
All automated tests: unit, component, regression, integration and playwright were performed. 

## Checklist:d


src/EventListeners/RouteBasedExceptionListener.php
```diff
@@ -17,24 +21,54 @@
...
+
     public function onKernelException(ExceptionEvent $event): void
     {
         $request = $event->getRequest();
-        $route   = $request->attributes->get('_route');
+        $route = $request->attributes->get('_route');
 
         $exception = $event->getThrowable();
 
+        $this->logger->error('OnKernelException', ['route' => $route, 'exception' => $exception]);
```
jpwhite4: Is this logger->error desirable here? Did the pre-symfony code log all exceptions? if so where and in what format?

ryanrath: Nope, it was something used during debugging. It's been removed in the latest commit.

src/Security/Listeners/RoleRequiredAttributeListener.php
```diff
@@ -0,0 +1,139 @@
+<?php declare(strict_types=1);
+
+namespace CCR\Security\Listeners;
+
+use CCR\Security\Attributes\RoleRequired;
+use Psr\Log\LoggerInterface;
```
connersaeli:
```suggestion
```

src/EventListeners/RouteBasedExceptionListener.php
```diff
@@ -1,11 +1,15 @@
 <?php
+
 namespace CCR\EventListeners;
 
+use Psr\Log\LoggerInterface;
```
connersaeli:
```suggestion
```

connersaeli (approved):

